PR #9, opened 2015-01-08
Update for rust master (9f1ead8fa)

tomjakubowski:


src/lib.rs
```diff
@@ -244,7 +244,7 @@ impl fmt::Show for LogLevel {
 impl Logger for DefaultLogger {
     fn log(&mut self, record: &LogRecord) {
         match writeln!(&mut self.handle,
-                       "{}:{}: {}",
+                       "{:?}:{:?}: {:?}",
```
alexcrichton: I think that these should all largely be using `{}` because this is all human-readable output, could you switch `LogLevel` from `Show` to `String` and then add a `derive(Show)` to it?

tomjakubowski:
Updated

PR #16, opened 2015-02-02
Fix FromStr for new rustc

maurer:
I switched the ```FromStr``` instances to use ```Result<a,()>``` instead of Some as there was no obvious error messages for this kind of parsing (it's just turning strings into enum elements).

The Travis build will fail until a new nightly is created, since it is to deal with breaking changes in rustc.


TheNeikos:
The build seems to fail due to Travis not using the latest rust version? I can't explain it any other way.

Perhaps can @maurer resubmit it to make it build again? (If Travis has update that is.)

maurer:
https://travis-ci.org/rust-lang/log/builds/49326475#L136
Seems to be failing due to a dependency failing to build (it appears due ot the same issue, different package). I can take a look at the other package if someone else hasn't later today.

maurer:
Neeevermind. I failed to fix the subpackage, only the main "log" package, which is why Travis is failing. (I didn't even notice the subpackage since I wasn't using it)

maurer:
Code passes now.

PR #28, opened 2015-03-10
Respect padding for LogLevel names

CraZySacX:
LogLevel names are frequently output in log files and the ability to pad them to a width helps.


alexcrichton:
Thanks!

PR #29, opened 2015-03-11
Initialize env_logger using any environment variable key

nukep:
I feel that for larger scale, non-rust-lang projects, having an environment variable named `RUST_LOG` may be undesirable. This PR provides a way to use a different environment variable if it's desired.
The changes are not breaking.

Of course, the user has always been able to filter log messages on a per-module basis, and thus also per-application (`app1=info,app2,...`).
This PR is more for projects that want to stay consistent with their environment variable naming conventions (e.g. `APP1_FOO`, `APP1_BAR`, `APP1_LOG`...).

### DIFF
--- a/env/src/lib.rs
+++ b/env/src/lib.rs
@@ -73,10 +73,10 @@
 //! ## Enabling logging
 //!
 //! Log levels are controlled on a per-module basis, and by default all logging
-//! is disabled except for `error!`. Logging is controlled via the `RUST_LOG`
-//! environment variable. The value of this environment variable is a
-//! comma-separated list of logging directives. A logging directive is of the
-//! form:
+//! is disabled except for `error!`. Logging is controlled either via the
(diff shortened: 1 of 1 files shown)


alexcrichton:
This may not be quite the way that we would want to go about this. If you want to programmatically control how logging is initialized, we probably want to expose the configuration exactly as is. For example the convenience function `init` would look like:

```rust
fn init() {
    init_with_config(Config::from_env("RUST_LOG"));
}

fn init_with_config(cfg: Config) { ... }
```

The idea here is that configuration via another environment variable is only one axis of runtime configuration, you may also not want to use environment variables at all.

nukep:
That's fair!
However, I'll leave the feature alone for until I can get around to implementing it (or unless someone else does).

Being able to initialize the logger without environment variables would make `env_logger` somewhat of a misnomer, so I'll close this and possibly open another PR in the future that also renames the crate.

PR #31, opened 2015-03-18
Remove dependence on std::rt

alexcrichton:
This moves the usage of `rt::at_exit` to `libc::atexit` instead. It's not
looking like it's likely for `rt::at_exit` to be stable at 1.0, and
`libc::atexit` is at least supported on major platforms.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -148,7 +148,9 @@
        html_favicon_url = "http://www.rust-lang.org/favicon.ico",
        html_root_url = "http://doc.rust-lang.org/log/")]
 #![warn(missing_docs)]
-#![feature(core, std_misc)]
+#![feature(core)]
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -605,7 +612,8 @@ pub struct SetLoggerError(());
 
 impl fmt::Display for SetLoggerError {
     fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
-        write!(fmt, "attempted to set a logger after the logging system was already initialized")
+        write!(fmt, "attempted to set a logger after the logging system ]\
```
sfackler: Typo

alexcrichton: Oops, fixed!

PR #53, opened 2015-07-29
Change dependency on libc to one on shutdown_hooks

CruzBishop:
This ~~may cause~~ causes an inherited minimum Rust version of 1.1.0 due to the inclusion of std::os::raw::c_int as a return value of the (non-public) atexit() function. ~~I need Travis to let me know.~~ Perhaps this means this shouldn't be included in the 0.3 series but in 0.4?

If I end up changing the method signature in the future, I'll update the log crate within an hour. I probably won't need to, though :)

Edit: Updated travis.yml to test on 1.1.0 as a minimum and also the "current stable" version in the future

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,8 @@
 language: rust
 sudo: false
 rust:
-  - 1.0.0
+  - 1.1.0
+  - stable
   - beta
   - nightly
 script:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,4 +17,4 @@ name = "filters"
 harness = false
 
 [dependencies]
-libc = "0.1"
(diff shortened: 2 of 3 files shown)


CruzBishop:
@huonw Sorry, is there a way to specify a minimum Rust version in Cargo at the moment? I'm assuming not since Rust is not, for lack of a better word, shipped via cargo at the moment :\

alexcrichton:
Thanks! This is such a small dependency though I don't think it's worth bringing in an extra crate for now. If we need fancier bindings to `atexit` then I think it's fine to bring in a crate but that can always happen later!

CruzBishop:
@alexcrichton Thanks for the honest answer :)

Issue #18, opened 2015-02-07
Feature request: Logging only for testing


derekdreery:
It would be good to have a macro which does a println/log during testing, but is removed when not testing.

Another alternative would be a log statement which is removed when ndebug is set.

derekdreery:
Or am I misunderstanding?

jimmycuadra:
I'm wondering about this as well. I'm building a library that uses Hyper, and I want to run my tests with Hyper's logs turned on, but it's not obvious how to initialize env_logger within the test suite.

jonas-schievink:
I want this too, this would make testing much easier. It would be great if the default log level would be set to `DEBUG` when `cfg(test)` is active.

KodrAus:
I think this is a bit out-of-scope for `log`, but definitely something we could look at supporting more smoothly in `env_logger`, or another alternative crate specifically for logging in tests.

We've had a few years without activity here so I'll go ahead and close this one now as out-of-scope, but please leave a comment if there's something new to add!

PR #56, opened 2015-09-13
Add instructions and example for using env_logger in tests

carols10cents:
This took me some trial and error to get working. 

This PR is also me asking if I'm doing this right at all... i *think* this is the best way to get log messages in tests, but I could be missing something. These make me think I'm not missing something, that this is the best way at the moment:

* https://github.com/rust-lang/log/issues/18
* https://www.reddit.com/r/rust/comments/31y66n/how_to_log_debug_in_library_tests/ctue5n0?context=10000
* http://stackoverflow.com/questions/30177845/how-to-initialize-the-logger-for-integration-tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,3 +79,78 @@ fn main() {
     // ...
 }
 ```
+
+## In tests
+
+Tests can use the `env_logger` crate to see log messages generated during that test:
+
+```toml
+[dependencies]
+log = "0.3"
+
+[dev-dependencies]
+env_logger = "0.3"
+```
+
+```rust
+#[macro_use]
+extern crate log;
+
+fn add_one(num: i32) -> i32 {
+    info!("add_one called with {}", num);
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -79,3 +79,78 @@ fn main() {
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    extern crate env_logger;
+
+    #[test]
+    fn it_adds_one() {
+        env_logger::init().unwrap();
```
alexcrichton: Perhaps this could suggest something like:

```rust
let _ = env_logger::init();
```

That way multiple tests could call this and every test that has this will be guaranteed to have logging enabled?

carols10cents:
Oops! Sorry for the delay in fixing this :)

alexcrichton:
Thanks!

PR #58, opened 2015-10-02
Statically disable debug logging in optimized builds

mbrubeck:
This PR modifies `log!` and associated macros to disable logging at compile time for the Debug and Trace log levels when the `debug_assertions` option is disabled (which it is by default when optimizations are enabled).

The goal of this change is to make it easy to leave debug logging in place throughout a project and its dependencies, but to pay for it only in debug builds, and have it completely optimized out in release builds.

The `log_level` configuration option provides an existing way to do this. However, in practice there is no supported way to toggle `log_level` for a package *and* all of its dependencies when building with Cargo.  However, Cargo does provide per-profile options for controlling rustc's built-in `debug_assertions` option. This option is a pretty good fit for debug logging, since it is intended to statically disable code that is useful for debugging but may have a performance cost.  In fact, rustc's internal liblog [already uses it this way for its `debug!` logging macro](https://github.com/rust-lang/rust/blob/168a23ebe1729386138fa71643382fdd64fac205/src/liblog/macros.rs#L160).

This PR is alternative to rust-lang/cargo#1948 which was closed by @alexcrichton in favor of “options that are local to the `log` crate itself.”  Compared to the Cargo PR, this one is not as configurable, but does not require any external changes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,12 +2,15 @@ language: rust
 sudo: false
 rust:
   - 1.0.0
+  - stable
   - beta
   - nightly
 script:
   - cargo build --verbose
   - cargo test --verbose
   - cargo test --verbose --manifest-path env/Cargo.toml
+  - cargo run --verbose --manifest-path tests/max_level_features/Cargo.toml
+  - cargo run --verbose --manifest-path tests/max_level_features/Cargo.toml --release
   - cargo doc --manifest-path env/Cargo.toml
 after_success: |
   [ $TRAVIS_BRANCH = master ] &&
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -12,8 +12,8 @@
 /// This macro will generically log with the specified `LogLevel` and `format!`
 /// based argument list.
 ///
-/// The `log_level` cfg can be used to statically disable logging at various
-/// levels.
+/// The `max_level_*` features can be used to statically disable logging at
+/// various levels.
 #[macro_export]
 macro_rules! log {
     (target: $target:expr, $lvl:expr, $($arg:tt)+) => ({
@@ -23,12 +23,7 @@ macro_rules! log {
             __module_path: module_path!(),
         };
         let lvl = $lvl;
-        if !cfg!(log_level = "off") &&
-                (lvl <= $crate::LogLevel::Error || !cfg!(log_level = "error")) &&
-                (lvl <= $crate::LogLevel::Warn || !cfg!(log_level = "warn")) &&
-                (lvl <= $crate::LogLevel::Debug || !cfg!(log_level = "debug")) &&
-                (lvl <= $crate::LogLevel::Info || !cfg!(log_level = "info")) &&
-                lvl <= $crate::max_log_level() {
+        if lvl <= $crate::__static_max_level() && lvl <= $crate::max_log_level() {
             $crate::__log(lvl, $target, &_LOC, format_args!($($arg)+))
         }
     });
@@ -37,8 +32,7 @@ macro_rules! log {
 
 /// Logs a message at the error level.
 ///
-/// Logging at this level is disabled if the `log_level = "off"` cfg is
-/// present.
+/// Logging at this level is disabled if the `max_level_off` feature is present.
 #[macro_export]
 macro_rules! error {
     (target: $target:expr, $($arg:tt)*) => (
@@ -51,8 +45,12 @@ macro_rules! error {
 
 /// Logs a message at the warn level.
 ///
-/// Logging at this level is disabled if any of the following cfgs are present:
-/// `log_level = "off"` or `log_level = "error"`.
+/// Logging at this level is disabled if any of the following features are
+/// present: `max_level_off` or `max_level_error`.
+///
+/// When building in release mode (i.e., without the `debug_assertions` option),
+/// logging at this level is also disabled if any of the following features are
+/// present: `release_max_level_off` or `max_level_error`.
 #[macro_export]
 macro_rules! warn {
     (target: $target:expr, $($arg:tt)*) => (
@@ -65,9 +63,13 @@ macro_rules! warn {
 
 /// Logs a message at the info level.
 ///
-/// Logging at this level is disabled if any of the following cfgs are present:
-/// `log_level = "off"`, `log_level = "error"`, or
-/// `log_level = "warn"`.
+/// Logging at this level is disabled if any of the following features are
+/// present: `max_level_off`, `max_level_error`, or `max_level_warn`.
+///
+/// When building in release mode (i.e., without the `debug_assertions` option),
+/// logging at this level is also disabled if any of the following features are
+/// present: `release_max_level_off`, `release_max_level_error`, or
+/// `release_max_level_warn`.
 #[macro_export]
 macro_rules! info {
     (target: $target:expr, $($arg:tt)*) => (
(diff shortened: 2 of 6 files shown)


alexcrichton:
I wonder if it may be worth putting this sort of functionality behind a feature flag for now to preserve the old behavior. I know that I personally really enjoy having debug logs in Cargo even in release builds! Although the other way around may also make sense where a feature flag enables debug logging in release builds as well.

Either way, though, could you add a test for this functionality? Also cc @sfackler, curious on your thoughts as well.

mbrubeck:
> I know that I personally really enjoy having debug logs in Cargo even in release builds!

Yes — if this change is accepted, developers will have reason to be more picky about which log levels they use: `info` or lower for things that should remain in release builds, and `debug` or `trace` only for things that they would like compiled out in production.

I'll work on adding a feature flag and test.

sfackler:
I'm on board with a feature flag, but would really like some way to have the more robust `log_level` cfg configurability exposed in a reasonable way through cargo.

mbrubeck:
Me too! Maybe you can help me convince the Cargo developers to either add `log_level` explicitly, or add a way to set arbitrary `--cfg` options that apply to the entire dependency tree. :)

mbrubeck:
@sfackler This now includes a test for the new behavior, and a feature to restore the old behavior.  Feel free to bikeshed the feature name; I wasn't sure what to call it.

Note: I added the Rust "stable" channel to `.travis.yml` and removed Rust 1.0.0, because Rust 1.0.0 doesn't include Cargo feature support.

sfackler:
I think we'll actually want to invert the behavior of the feature and keep everything enabled by default to make sure everything's backwards compatible.

mbrubeck:
@sfackler:

> I think we'll actually want to invert the behavior of the feature and keep everything enabled by default to make sure everything's backwards compatible.

I flipped the feature so the new behavior is now opt-in (though personally I still feel that statically disabling debug code in release builds is a better default for most projects).

sfackler:
Thinking about it more, I kind of want to go further and have sets of Cargo features to control the maximum release and debug log levels that would supplant the current `log_level` cfg things that aren't really usable right now. Something like:

```toml
[features]
max_level_off = []
max_level_error = ["max_level_off"]
max_level_warn = ["max_level_error"]
max_level_info = ["max_level_warn"]
max_level_debug = ["max_level_info"]
max_level_trace = ["max_level_debug"]

debug_max_level_off = []
debug_max_level_error = ["debug_max_level_off"]
debug_max_level_warn = ["debug_max_level_error"]
debug_max_level_info = ["debug_max_level_warn"]
debug_max_level_debug = ["debug_max_level_info"] # The naming here's a bit bizarre :/
debug_max_level_trace = ["debug_max_level_debug"]
```

And then have something like

```rust
#[inline(always)]
#[doc(hidden)]
pub fn __static_max_level() -> LogLevelFilter {
    if cfg!(debug_assertions) {
        if cfg!(feature = "debug_max_level_trace") {
            LogLevelFilter::Trace
        } else if cfg!(feature = "debug_max_level_debug") {
            LogLevelFilter::Debug
        } else if cfg!(feature = "debug_max_level_info") {
            LogLevelFilter::Info
        } else if cfg!(feature = "debug_max_level_warn") {
            LogLevelFilter::Warn
        } else if cfg!(feature = "debug_max_level_error") {
            LogLevelFilter::Error
        } else if cfg!(feature = "debug_max_level_off") {
            LogLevelFilter::Off
        } else {
            LogLevelFilter::Trace
        }
    } else {
        if cfg!(feature = "max_level_trace") {
            LogLevelFilter::Trace
        } else if cfg!(feature = "max_level_debug") {
            LogLevelFilter::Debug
        } else if cfg!(feature = "max_level_info") {
            LogLevelFilter::Info
        } else if cfg!(feature = "max_level_warn") {
            LogLevelFilter::Warn
        } else if cfg!(feature = "max_level_error") {
            LogLevelFilter::Error
        } else if cfg!(feature = "max_level_off") {
            LogLevelFilter::Off
        } else {
            LogLevelFilter::Trace
        }
    }
}
```

mbrubeck:
Updated to replace the `cfg(log_level)` with a set of Cargo features, based on @sfackler's suggestion above.

.travis.yml
```diff
@@ -1,13 +1,15 @@
 language: rust
 sudo: false
 rust:
-  - 1.0.0
+  - stable
```
alexcrichton: Can you leave `1.0.0` here? The intention with that is to ensure that this continues to always compile on 1.0.0 Rust

src/macros.rs
```diff
@@ -95,9 +101,13 @@ macro_rules! debug {
 
 /// Logs a message at the trace level.
 ///
-/// Logging at this level is disabled if any of the following cfgs are present:
-/// `log_level = "off"`, `log_level = "error"`, `log_level = "warn"`,
-/// `log_level = "info"`, or `log_level = "debug"`.
+/// Logging at this level is disabled if any of the following features are present:
+/// `max_level_off`, `max_level_error`, `max_level_warn`, `max_level_info`, or `max_level_debug`.
+///
+/// When building in release mode (i.e., without the `debug_assertions` option), logging at this
```
alexcrichton: Could you be sure to wrap these docs to 80 chars? (the style of this crate)

alexcrichton:
This looks good to me, thanks @mbrubeck! 

r? @sfackler

.travis.yml
mbrubeck: Rust 1.0.0 does not support Cargo features.

alexcrichton: Are you sure? Features have been around since Cargo's inception basically...

mbrubeck:
Updated to wrap comments to 80 columns and restore Rust 1.0.0 to the list of Travis jobs. (An earlier version of this PR failed to parse Cargo.toml with Rust 1.0, but it looks like it's passing now.)

src/macros.rs
```diff
@@ -23,12 +23,7 @@ macro_rules! log {
             __module_path: module_path!(),
         };
         let lvl = $lvl;
-        if !cfg!(log_level = "off") &&
-                (lvl <= $crate::LogLevel::Error || !cfg!(log_level = "error")) &&
-                (lvl <= $crate::LogLevel::Warn || !cfg!(log_level = "warn")) &&
-                (lvl <= $crate::LogLevel::Debug || !cfg!(log_level = "debug")) &&
-                (lvl <= $crate::LogLevel::Info || !cfg!(log_level = "info")) &&
-                lvl <= $crate::max_log_level() {
+        if lvl <= $crate::__static_max_level() && lvl <= $crate::max_log_level() {
```
sfackler: Should we worry about breaking back-compat for people that are actually using the log_level cfg? I don't know if any of those people exist, but might be worth leaving this around to support them? Thoughts @alexcrichton?

sfackler:
Looks good to me modulo the back-compat questions!

mbrubeck:
I think if we ever want to make breaking changes to this crate, it should happen now while it's still in the nursery, and sooner is better than later.

So unless we want to support two (static) ways of setting the log level forever, I think it's better to remove the old one now. If we bump the package version, it shouldn't cause any silent breakage.

sfackler:
I'm not opposed to making breaking changes to this crate, but we have to be somewhat careful about how we do it. The location of the crate in Github doesn't matter all that much - liblog is *incredibly* heavily depended upon, and bad things will happen if a single executable has multiple copies lying around (some log messages will disappear into the ether!). If we want to break back compat, I'd like to do it to make significantly larger changes than just cutting some obscure cfg switches.

mbrubeck:
So I guess the question is whether this breaks anyone's code in practice. I believe the only built-in way to make Cargo set the `log_level` option is to output `cargo:rustc-cfg=log_level="debug"` from a build script. Is it possible to find out if any packages on crates.io are doing this in their build scripts?

alexcrichton:
Yeah I also suspect that this actually isn't being used that much in practice (e.g. it's obscure and not super documented), so I'm fine breaking it for now (and we can always add it back in a future point release). Along the lines of the whole "technically but not practically breaking"

Thanks again @mbrubeck!

sfackler:
:+1:

PR #59, opened 2015-10-02
[docs] Use a better example for log_enabled!

mbrubeck:
The previous example code for the `log_enabled!` macro doesn't properly demonstrate why to use it, since the use of `log_enabled!` doesn't improve anything.  (The `debug!` macro already prevents the expensive call when debug logging is disabled.)

This changes the example to one that has better performance compared to `debug!` alone.

r? @sfackler

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -123,10 +123,12 @@ macro_rules! trace {
 ///
 /// # fn foo() {
 /// if log_enabled!(Debug) {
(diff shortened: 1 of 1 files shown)


sfackler:
Strictly speaking, the old setup will still improve performance when debug logging is dynamically disabled, but this change seems like a nice improvement regardless. Thanks!

PR #62, opened 2015-10-27
env_logger: Add customized init functions

diwic:
This adds the possibility to customize how to init env_logger by
parsing a custom environment variable, or a logging spec directly.

If you were ever annoyed that `RUST_LOG=debug cargo run` outputted tons of debug info from cargo when all you wanted was to see debug output from the program you wanted to run, you're going to love this feature.

### DIFF
--- a/env/src/lib.rs
+++ b/env/src/lib.rs
@@ -75,7 +75,8 @@
 //!
 //! Log levels are controlled on a per-module basis, and by default all logging
 //! is disabled except for `error!`. Logging is controlled via the `RUST_LOG`
-//! environment variable. The value of this environment variable is a
+//! environment variable (this can be customized using `init_with_env_var` or
+//! `init_with_spec`). The value of this environment variable is a
 //! comma-separated list of logging directives. A logging directive is of the
 //! form:
 //!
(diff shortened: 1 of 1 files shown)


alexcrichton:
In the past I've thought that we either want a generic `init` or a custom builder type to build up the configuration for this library (rather than a few assorted entrypoint functions), e.g. along the lines of https://github.com/rust-lang-nursery/log/pull/48 (which seems to have fallen by the wayside)

You can filter out logging via `RUST_LOG=foo=debug` where `foo` is the name of the program you're debugging.

diwic:
I don't think there would be any harm in doing both - have a few convenience entry points for the most common usages, and a configuration/builder/etc for more uncommon use cases.

That said, it looks like #48 would be good enough for my use case too, if it actually gets in.

alexcrichton:
Using a builder is generally much more composable as it's easier to add new functions and it's "one less concept to understand" because there's one main entry point (`init`) and then maximal customization through a builder. The builder can have convenience functions for parsing specs or looking at different environment variables.

dbrgn:
Yes, configuration options would be very nice.

dtolnay:
This is now supported using the builder from #48. To expand on what Alex mentioned about composability: in #146 we are considering distinguishing between `init` which panics and `try_init` which returns a Result. This PR would have required us to consider also `try_init_with_env_var` and `try_init_with_spec` which is an example of not being nicely composable.

PR #65, opened 2015-11-14
Added "force-stdout" compilation feature

dbrgn:
Hi, I'd like to have a way to force all logs from env_logger to go to stdout instead of stderr.

As there's currently no way to pass in configuration to the env_logger (see #62), I settled for a compilation feature. That might even be the better option, because the user can choose at compile time whether he/she wants to log to stderr or not.

In my case I need it to build a Docker container. The Docker "logs" command only considers stdout, as far as I know.

Honestly I didn't manage to test this. When I pointed an existing project at the code directory (even without my changes), logging stops. But that's probably just a problem on my side.

If you like this approach / change, I can also provide documentation.

### DIFF
--- a/env/Cargo.toml
+++ b/env/Cargo.toml
@@ -17,3 +17,7 @@ path = ".."
 
 [dependencies]
 regex = "0.1"
+
+[features]
+default = []
+force-stdout = []
--- a/env/src/lib.rs
+++ b/env/src/lib.rs
@@ -314,7 +314,11 @@ impl Log for Logger {
             }
         }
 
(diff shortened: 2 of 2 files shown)


dbrgn:
To use this, the user would use the following line in `Cargo.toml`:

    env_logger = { version = "^0.3.2", features = ["force-stdout"] }

dbrgn:
Ah, that information about Docker and stdout is actually wrong. The compilation option might still be useful. (But maybe you want to solve it through a builder instead.)

alexcrichton:
Yeah I think I would personally prefer this as a builder option, this is in theory only activated from the application anyway.

PR #68, opened 2016-01-10
Split crate into log and log_core

### LINKED ISSUE #13: Use core not std
I believe everything in the logging interface just requires core, not std. (The `Box<Log>` could become a `Unique<Log>` as it is never deallocated.) This is important for allowing the interface to reach its full potential, e.g. serial port debugging before allocation is set up. https://github.com/rust-lang/bitflags was similarly changed to use core for the same reason.

If this sounds good, I would be happy to do the implementation work.

Amanieu:
This splits the `log` crate into `log` and `log_core`. `log_core` has no dependency on `libc` or `libstd`, which allows it to be used in kernel or embedded environments.

Both crates export a similar API, with a few exceptions:
- Only in `log`: panic handler support, `set_logger` function
- Only in `log_core`: `set_logger_raw`, `shutdown_logger`

`log` re-exports most of the definitions from `log_core`, which allows libraries using `log_core` to work with executables using `log`.

Fixes #13


Amanieu:
It seems the build is failing on 1.0.0 due to the lack of `Box::into_raw` and `Box::from_raw`. Are we still interested in supporting such an old version of Rust? If so I can change those calls into transmutes.

core/src/lib.rs
```diff
@@ -0,0 +1,892 @@
...
+//! ```
+
+#![doc(html_logo_url = "https://www.rust-lang.org/logos/rust-logo-128x128-blk-v2.png",
+       html_favicon_url = "https://www.rust-lang.org/favicon.ico",
+       html_root_url = "https://doc.rust-lang.org/log/")]
+#![warn(missing_docs)]
+#![cfg_attr(feature = "no_std", no_std)]
+
+#[cfg(all(feature = "no_std", not(test)))]
+mod std {
```
sfackler: You can simplify this to `extern crate core as std`.

core/src/lib.rs
```diff
@@ -0,0 +1,892 @@
...
+}
+
+fn ok_or<T, E>(t: Option<T>, e: E) -> Result<T, E> {
+    match t {
+        Some(t) => Ok(t),
+        None => Err(e),
+    }
+}
+
+// Reimplemented here because std::ascii is not available in libcore
```
sfackler: Might be worth filing a rust-lang/rust issue about this.

src/lib.rs
```diff
@@ -152,418 +152,19 @@
...
-    }
-}
+extern crate log_core;
 
-impl MaxLogLevelFilter {
-    /// Gets the current maximum log level filter.
-    pub fn get(&self) -> LogLevelFilter {
-        max_log_level()
-    }
+// Re-export most of log_core, but not set_logger_raw and shutdown_logger
```
sfackler: Can we `#[doc(inline)]` the public API parts of this?

src/lib.rs
```diff
@@ -152,418 +152,19 @@
...
+                   __log, __static_max_level};
 
-    /// Sets the maximum log level.
-    pub fn set(&self, level: LogLevelFilter) {
-        MAX_LOG_LEVEL_FILTER.store(level as usize, Ordering::SeqCst)
-    }
-}
+#[macro_use]
+#[path = "../core/src/macros.rs"]
+mod macros;
```
sfackler: I think it's possible to reexport macros cross crate but I could be wrong there.

sfackler:
Awesome!

`into_raw` pushes the minimum supported Rust version up to 1.4.0, which I'm personally fine with. Thoughts, @alexcrichton?

core/src/lib.rs
```diff
@@ -0,0 +1,892 @@
...
+//! # impl log_core::Log for SimpleLogger {
+//! #   fn enabled(&self, _: &LogMetadata) -> bool { false }
+//! #   fn log(&self, _: &log_core::LogRecord) {}
+//! # }
+//! # fn main() {}
+//! pub fn init() -> Result<(), SetLoggerError> {
+//!     unsafe {
+//!         log_core::set_logger_raw(|max_log_level| {
+//!             max_log_level.set(LogLevelFilter::Info);
+//!             Box::into_raw(Box::new(SimpleLogger))
```
sfackler: Might be a bit weird to be using liballoc in an example for a no_std crate?

src/lib.rs
Amanieu: It seems this is already done by default. The documentation generated for the log crate contains documentation for all the re-exported items.

src/lib.rs
Amanieu: `#[macro_reexport]` exists but it is unstable. I think this is because of issues with how `$crate` is handled.

sfackler: Ah right

src/lib.rs
sfackler: Perfect. Are all of the `__foo` methods hidden in these docs as well?

Amanieu: Yep

Amanieu:
I've addressed all the given comments.

alexcrichton:
I think I'd personally prefer to only have one crate, `log`, which everyone uses (std and no_std). Would it be possible to have this with basically the same api, but a subset is removed for the no_std version?

sfackler:
Logger registration is different in log_core which would complicate things. I guess those functions could be kept around in the std version though.

Amanieu:
That's what I tried originally, but the problem is that there is no way to disable the `libc` dependency in cargo if the `no_std` feature is enabled. Instead what I had to do was add a `use_std` feature that is enabled by default, make the `libc` dependency optional and have it get selected by the `use_std` feature. Since `use_std` is enabled by default, you would have to pass `--no-default-features` to cargo to build `log` with no_std support.

I figured this was getting too complicated and decided that it would be cleaner to split the no_std core into a separate crate.

sfackler:
We're only using libc for atexit iirc, and I'd personally be fine just copying the extern declaration in directly if that's a blocker.

sfackler:
Oh, actually, a no_std feature will result in bad behavior. If any dependency activates it, then a std-using root crate will suddenly lose its logger registration function. The "right thing" to do would be the default feature.

Amanieu:
I think having a separate `log_core` crate is cleaner than messing around with default features. However I can also reimplement my old `use_std` solution if that is preferred.

alexcrichton:
Instead of having a no_std feature it could instead be a "std" feature that's enabled by default, that way enabling no_std support would just be `default-features = false`.

I also think that while the registration is different, the ergonomic registration could just defer to the unergonomic registration, and then a no_std version only has the unergonomic variants.

I'm wary to split such a core crate a `log` into two crates. It's somewhat confusing that if you're no_std you have to use log_core, when everywhere else you just use log. It's much easier conceptually to just use log everywhere.

Amanieu:
I opened another PR (#70) that doesn't split the crate

PR #69, opened 2016-01-12
Fixed handling of filter regexps.

psii:
Regexps in RUST_LOG as explained in the module documentation should
only allow log messages that match the given expression and not the
other way around.

### DIFF
--- a/env/Cargo.toml
+++ b/env/Cargo.toml
@@ -17,3 +17,7 @@ path = ".."
 
 [dependencies]
 regex = "0.1"
+
+[[test]]
+name = "regexp_filter"
+harness = false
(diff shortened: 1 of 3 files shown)


alexcrichton:
Can you be sure to add a test for this as well? It seems odd that this wasn't tested before...

psii:
I attempted to write a test case before submitting this pull request, but I
didn't see a straightforward way to do this. Some restructuring of the code
is necessary. I can look for a solution but it will take some time.

2016-01-12 17:49 GMT+01:00 Alex Crichton <notifications@github.com>:

alexcrichton:
One way to do this is to add an integration test under `tests/foo.rs`, disable the test harness, and then have the test re-execute itself with an environment variable set. That way it can test the stdout/stderr.

psii:
Added a test case which is implemented like you suggested.

alexcrichton:
Thanks!

PR #70, opened 2016-01-12
Add support for no_std

### LINKED ISSUE #13: Use core not std

Amanieu:
Second attempt at support `no_std`, this time without splitting the crate into two. Instead, a `use_std` feature is used, which is enabled by default.

Fixes #13

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,10 @@ rust:
   - nightly
 script:
   - cargo build --verbose
+  - ([ $TRAVIS_RUST_VERSION != nightly ] || cargo build --verbose --no-default-features)
   - ([ $TRAVIS_RUST_VERSION != nightly ] || cargo build --verbose --features nightly)
-  - cargo test --verbose
+  - ([ $TRAVIS_RUST_VERSION == 1.0.0 ] || cargo test --verbose)
+  - ([ $TRAVIS_RUST_VERSION != nightly ] || cargo test --verbose --no-default-features)
   - cargo test --verbose --manifest-path env/Cargo.toml
   - cargo run --verbose --manifest-path tests/max_level_features/Cargo.toml
   - cargo run --verbose --manifest-path tests/max_level_features/Cargo.toml --release
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -136,10 +136,50 @@
 //! #   fn log(&self, _: &log::LogRecord) {}
 //! # }
 //! # fn main() {}
+//! # #[cfg(feature = "use_std")]
 //! pub fn init() -> Result<(), SetLoggerError> {
 //!     log::set_logger(|max_log_level| {
 //!         max_log_level.set(LogLevelFilter::Info);
 //!         Box::new(SimpleLogger)
+//!     })
+//! }
+//! ```
+//!
+//! # Use with `no_std`
+//!
+//! To use the `log` crate without depending on `libstd`, you need to specify
+//! `default-features = false` when specifying the dependency in `Cargo.toml`.
+//! This makes no difference to libraries using `log` since the logging API
+//! remains the same. However executables will need to use the `set_logger_raw`
+//! function to initialize a logger and the `shutdown_logger_raw` function to
+//! shut down the global logger before exiting:
+//!
+//! ```rust
+//! # extern crate log;
+//! # use log::{LogLevel, LogLevelFilter, SetLoggerError, ShutdownLoggerError,
+//! #           LogMetadata};
+//! # struct SimpleLogger;
+//! # impl log::Log for SimpleLogger {
+//! #   fn enabled(&self, _: &LogMetadata) -> bool { false }
+//! #   fn log(&self, _: &log::LogRecord) {}
+//! # }
+//! # impl SimpleLogger {
+//! #   fn flush(&self) {}
+//! # }
+//! # fn main() {}
+//! pub fn init() -> Result<(), SetLoggerError> {
+//!     unsafe {
+//!         log::set_logger_raw(|max_log_level| {
+//!             static LOGGER: SimpleLogger = SimpleLogger;
+//!             max_log_level.set(LogLevelFilter::Info);
+//!             &SimpleLogger
+//!         })
+//!     }
+//! }
+//! pub fn shutdown() -> Result<(), ShutdownLoggerError> {
+//!     log::shutdown_logger_raw().map(|logger| {
+//!         let logger = unsafe { &*(logger as *const SimpleLogger) };
+//!         logger.flush();
 //!     })
 //! }
 //! ```
@@ -150,10 +190,16 @@
 #![warn(missing_docs)]
 #![cfg_attr(feature = "nightly", feature(panic_handler))]
 
+#![cfg_attr(not(feature = "use_std"), no_std)]
+
+#[cfg(not(feature = "use_std"))]
+extern crate core as std;
+
+#[cfg(feature = "use_std")]
 extern crate libc;
 
-use std::ascii::AsciiExt;
 use std::cmp;
+#[cfg(feature = "use_std")]
 use std::error;
 use std::fmt;
 use std::mem;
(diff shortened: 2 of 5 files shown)


src/lib.rs
```diff
@@ -150,10 +190,17 @@
 #![warn(missing_docs)]
 #![cfg_attr(feature = "nightly", feature(panic_handler))]
 
+#![cfg_attr(all(not(feature = "use_std"), not(test)), no_std)]
```
alexcrichton: I believe that `no_std` should be usable even during `--test`, so the `not(test)` clause shouldn't be necessary here

src/lib.rs
```diff
@@ -150,10 +190,17 @@
...
+
+#[cfg(all(not(feature = "use_std"), not(test)))]
```
alexcrichton: As above, I think it's ok to avoid the `not(test)` clause

src/lib.rs
```diff
@@ -150,10 +190,17 @@
...
+#[macro_use]
```
alexcrichton: This macro use should be ok to omit because the compiler will inject `#[macro_use] extern crate core`

src/lib.rs
```diff
@@ -143,6 +144,45 @@
 //!     })
 //! }
 //! ```
+//!
+//! # Use with `no_std`
+//!
+//! To use the `log` crate without depending on `libstd`, you need to specify
+//! `default_features = false` when specifying the dependency in `Cargo.toml`.
```
alexcrichton: s/default_features/default-features/

src/lib.rs
```diff
@@ -163,32 +210,32 @@ use std::sync::atomic::{AtomicUsize, ATOMIC_USIZE_INIT, Ordering};
...
 // the logger. At that point, the at_exit routine just waits for the refcount
 // to reach 0 before deallocating the logger. Note that the refcount does not
-// necessarily monotonically decrease at this point, as new log calls still
-// increment and decrement it, but the interval in between is small enough that
-// the wait is really just for the active log calls to finish.
-static LOGGER: AtomicUsize = ATOMIC_USIZE_INIT;
+// necessarily monotonically decrease at this point, as new  log calls still
+// increment and decrement it, but the interval in between is  small enough
+// that the wait is really just for the active log calls to finish.
+static mut LOGGER: Option<*const Log> = None;
```
alexcrichton: Was this protocol change to accomodate a null pointer for a logger?

Amanieu: The original code used a `Box<Box<Log>>` in an atomic. The outer box is necessary because the inner one is a fat pointer, which doesn't fit in an atomic. The new code just guards a non-atomic fat pointer using an atomic integer.

src/lib.rs
Amanieu: The tests actually depend on `libstd` since they make use of `.to_string()`.

Amanieu:
Updated with given feedback.

.travis.yml
```diff
@@ -1,14 +1,15 @@
 language: rust
 sudo: false
 rust:
```
sfackler: Can we continue testing against our minimum supported Rust version, which'll be 1.4.0 I believe.

Amanieu: Done

src/lib.rs
```diff
@@ -578,32 +643,88 @@ pub fn max_log_level() -> LogLevelFilter {
...
+    extern fn shutdown() {
+        shutdown_logger().map(|logger| unsafe {
+            Box::from_raw(logger as *mut Log);
+        }).ok();
+    }
+}
 
-    unsafe {
-        assert_eq!(libc::atexit(shutdown), 0);
+/// Sets the global logger from a raw pointer.
```
sfackler: We should probably note here and in `shutdown_logger` that they should only be used in a `no_std` setup.

src/lib.rs
```diff
@@ -578,32 +643,88 @@ pub fn max_log_level() -> LogLevelFilter {
...
+            Ok(())
+        }
+        Err(_) => Err(SetLoggerError(())),
+    };
 
-    let logger = Box::new(make_logger(MaxLogLevelFilter(())));
-    let logger = unsafe { mem::transmute::<Box<Box<Log>>, usize>(logger) };
-    LOGGER.store(logger, Ordering::SeqCst);
+    extern fn shutdown() {
+        shutdown_logger().map(|logger| unsafe {
```
sfackler: This is going to blow up if someone's called `shutdown_logger`.

src/lib.rs
```diff
@@ -578,32 +643,88 @@ pub fn max_log_level() -> LogLevelFilter {
...
+///
+/// This function may only be called once in the lifetime of a program, and may
+/// not be called before `set_logger_raw`. Once the global logger has been shut
+/// down, it can no longer be re-initialized by `set_logger_raw`. Any log
+/// events that occur after the call to `shutdown_logger` completes will be
+/// ignored.
+///
+/// The pointer that was originally passed to `set_logger_raw` is returned on
+/// success. At that point it is guaranteed that no other threads are
+/// concurrently the logger object.
```
sfackler: Gerund missing.

src/lib.rs
Amanieu: In that case the second call to `shutdown_logger` will simply fail, and this code will not be executed. In the worse case this leads to the `Box<Log>` being leaked, which isn't considered unsafe.

src/lib.rs
Amanieu: Note added, `shutdown_logger` already has a note saying that it shouldn't be used with `set_logger`.

src/lib.rs
alexcrichton: In that case you can have an explicit `extern crate std` in the `mod test` and use the libstd prelude

.travis.yml
alexcrichton: Hm which part requires 1.4.0? Ideally this would continue to compile on 1.0.0 as long as possible, but if it's too hard it may not be worth it

src/lib.rs
```diff
@@ -578,32 +643,93 @@ pub fn max_log_level() -> LogLevelFilter {
...
+/// ignored.
+///
+/// The pointer that was originally passed to `set_logger_raw` is returned on
+/// success. At that point it is guaranteed that no other threads are
+/// concurrently accessing the logger object.
+///
+/// This function should not be called when the global logger was registered
+/// using `set_logger`, since in that case the logger will automatically be shut
+/// down when the program exits
+pub fn shutdown_logger() -> Result<*const Log, ShutdownLoggerError> {
```
alexcrichton: This and `set_logger_raw` may want to be paired up in terms of naming or organization to set them apart from `set_logger` a bit more (and to make it clear that `shutdown_logger` isn't necessary most of the time)

.travis.yml
Amanieu: I think the code depends on these feature:
- `Box::into_raw`
- `Box::from_raw`
- `(*const Trait) as *const Struct` (used in one of the doc tests)

The first two can easily be changed into `mem::transmute`, but I'm not sure about the last one.

src/lib.rs
Amanieu: So what exactly are you suggesting? `shutdown_logger_raw`? Or something like `log::raw::{set_logger, shutdown_logger}`?

src/lib.rs
Amanieu: I tried it and it doesn't work because of the `extern crate core as std;` line just below. I think it's fine to just leave the this line as it is.

alexcrichton: I mean putting the crate in the module it's needed, not up here. This is a core crate of the Rust ecosystem and is likely to be used as an example for downstream code (e.g. idioms and such), so I'd prefer to have it as idiomatic as possible.

Amanieu: Sorry, I was being stupid with namespaces. I've moved the `extern crate
std` into the `tests` module.

.travis.yml
alexcrichton: Ah ok, would it be possible to leverage `mem::transmute` for now then? It's fine to just build the library on 1.0.0 instead of running the full test suite which may require more features.

src/lib.rs
alexcrichton: I think a module perhaps, but it seems kinda arbitrary. Including raw in the name seems prudent though.

alexcrichton:
@sfackler I'm happy with this [modulo a bit of bikeshedding](https://github.com/rust-lang-nursery/log/pull/70#discussion_r49694721), and I'll leave it up to you to merge

alexcrichton:
Also thanks @Amanieu, I'm excited to see more libs start picking up the `no_std` standard!

Amanieu:
Updated again, it should work on 1.0.0 now.

src/lib.rs
```diff
@@ -618,23 +744,43 @@ impl fmt::Display for SetLoggerError {
...
+
+impl fmt::Display for ShutdownLoggerError {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        write!(fmt, "attempted to shut down a logger without an active logger")
+    }
+}
+
+// The Error trait is not available in libcore
+#[cfg(feature = "use_std")]
+impl error::Error for ShutdownLoggerError {
```
sfackler: @alexcrichton there's nothing blocking us from adding `Error` to libcore right?

Doesn't affect this PR of course.

sfackler:
Looks good to me, thanks!

sfackler:
I'm going to take a pass over the docs to make sure everything looks right and then push a release.

sfackler:
Made a couple minor tweaks: #71

src/lib.rs
alexcrichton: Yeah I think it should be easy enough to add

Amanieu: This comment explains why `Error` is in libstd: https://github.com/rust-lang/rust/blob/master/src/libstd/error.rs#L39

alexcrichton: Oh right, forgot about that.

PR #86, opened 2016-06-23
Made Logger public.

### LINKED ISSUE #85: Add ability to log to both env_logger and a custom logger
At the moment, `env_logger::init` initializes the env logger, and registers it using the one-shot `log::set_logger`. This means there's no way to combine env logging with a custom logger.

Discussion on IRC with @alexcrichton here: http://logs.glob.uno/?c=mozilla%23servo&s=21+Jun+2016&e=21+Jun+2016#c459280

asajeffrey:
At the moment, the env logger cannot be combined with other loggers, since a) `Logger` is private, and b) logger creation and registration are combined in one function. This PR fixes those.

Fixes #85.

### DIFF
--- a/env/src/lib.rs
+++ b/env/src/lib.rs
@@ -141,7 +141,8 @@ use std::mem;
 
 use log::{Log, LogLevel, LogLevelFilter, LogRecord, SetLoggerError, LogMetadata};
 
-struct Logger {
+/// The logger.
+pub struct Logger {
     directives: Vec<LogDirective>,
     filter: Option<Regex>,
     format: Box<Fn(&LogRecord) -> String + Sync + Send>,
@@ -245,7 +246,15 @@ impl LogBuilder {
     /// global logger may only be initialized once. Future initialization
     /// attempts will return an error.
     pub fn init(&mut self) -> Result<(), SetLoggerError> {
+        log::set_logger(|max_level| {
+            let logger = self.build();
+            max_level.set(logger.filter());
+            Box::new(logger)
+        })
+    }
 
+    /// Build an env logger.
+    pub fn build(&mut self) -> Logger {
         if self.directives.is_empty() {
             // Adds the default filter if none exist
             self.directives.push(LogDirective {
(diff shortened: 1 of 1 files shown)


asajeffrey:
Should we rename `Logger` to `EnvLogger`?

alexcrichton:
I sort of liked one of the previous iterations where just `Box<Log>` was returned, perhaps that could be done instead?

asajeffrey:
The problem is that requires doing an extra dynamic method dispatch on every call to `log`, you end up with O(N) dynamic calls (where N is the number of loggers) rather than just 1.

alexcrichton:
Hm ok, seems reasonable to me. Thoughts @sfackler?

I think this may want some more docs, but I wouldn't mind doing so myself as well.

sfackler:
This seems reasonable to me. One little comment.

env/src/lib.rs
```diff
@@ -283,6 +280,21 @@ impl LogBuilder {
...
+        let mut builder = LogBuilder::new();
+
+        if let Ok(s) = env::var("RUST_LOG") {
+            builder.parse(&s);
+        }
+
+        builder.build()
+    }
+
+    pub fn filter(&self) -> LogLevelFilter {
```
sfackler: Not sure if this should be called `filter` or something like `max_level` to match up with the thing passed into `init`?

alexcrichton: `filter` seems ok to me as the returned type is `LogLevelFilter`, but I've always been thrown a bit at least when working out the interaction between `LogLevel`, `LogLevelFilter`, and the various functions that consume them.

sfackler: Yeah it's a bit awkward. The split's there to avoid having to deal with
what happens if you log at the "off" level.

asajeffrey: I think of it as an instance of the usual "counting fence posts" vs "counting gaps" feature.

asajeffrey:
It might be nice if `level` was in the `Log` interface, since that would simplify registering a logger, and make it easier to write log combinators. But that's a separate issue. Is there anything blocking this PR? It's needed for https://github.com/servo/servo/issues/11841.

asajeffrey:
This is blocking work on Servo logging / error reporting. Can we merge?

alexcrichton:
@sfackler how do you feel about merging, want to sort out naming issues beforehand?

sfackler:
No, it seems fine.

PR #89, opened 2016-07-28
Allow `log_panics` on stable, as Rust 1.10 stabilized `panic_handler`

kevinmehall:
Leaving it behind a Cargo feature in preemptive compliance with [RFC 1619](https://github.com/rust-lang/rfcs/pull/1619).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ release_max_level_info  = []
 release_max_level_debug = []
 release_max_level_trace = []
 
-nightly = []
+nightly = ["log_panics"]
+log_panics = []
 use_std = []
 default = ["use_std"]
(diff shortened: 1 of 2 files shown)


sfackler:
See https://crates.io/crates/log-panics for an out of crate version that also (optionally) supports backtraces.

@alexcrichton thoughts on this? I don't think log_panics will stay in this crate in the long term. Do you think we should land this change to get off the nightly name, or just deprecate this function entirely?

alexcrichton:
@sfackler yeah I'd be fine pushing towards the log-panics crate, the feature of backtraces there alone I think makes it perhaps more useful than the implementation in this crate.

kevinmehall:
Sounds good to me.

PR #93, opened 2016-08-27
column in LogLocation

jchlapinski:
Since column!() macro is readily available I have added column() getter to LogLocation. It is probably not the most useful information for logging, but it is cheap and nice to have IMHO.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -579,6 +579,8 @@ pub struct LogLocation {
     pub __file: &'static str,
     #[doc(hidden)]
     pub __line: u32,
+    #[doc(hidden)]
+    pub __column: u32,
 }
 
 impl LogLocation {
@@ -595,6 +597,11 @@ impl LogLocation {
     /// The line containing the message.
     pub fn line(&self) -> u32 {
         self.__line
+    }
+
+    /// The column at which the message was logged.
+    pub fn column(&self) -> u32 {
+        self.__column
     }
 }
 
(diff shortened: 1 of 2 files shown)


sfackler:
Could you expand a bit on the use cases you see for the column number in logger output?

jchlapinski:
# rustc is using source file reference of the form `(src/module/file.rs:12:12)`
# In IDEA + intellij-rust for some reason on the output console strings like `src/module/file.rs:12` (no colum) are not recognized and not converted into clickable links - granted that this is an issue with either IDEA or intellij-rust but anyway I ended up just adding dumb column number (like 1) to my logs which makes IDEA happy but is a substandard solution IMHO.
# In Eclipse + RustDR + Grep console if there is no column number after CTRL+clicking the link from the console editor sets the cursor on the end of the line. This makes long debug messages to scroll editor window horizontally, which I find annoying at times. Also for whatever reason the whole line is highlighted. With the column specified cursor the cursor sits at the beginning of the logging macro, which I find more convenient.

On a final note I do not consider the column number to be an essential information for logging, but on the other hand it is probably just as useful as module_path. Knowing the module by itself is probably to little to reliably find where the logging occurred in the source code, and if we do have file path and line number we probably do not need module path. In conclusion I would argue that column number is simply nice to have, and it comes with very little price as it stands. Hopefully we will also have something like __FUNC__ at some point, since that would be much more informative for logging and error reporting.

Regards

PR #98, opened 2017-01-02
Updated Regex dependency

Aaronepower:

PR #100, opened 2017-01-11
Upgrade regex to 0.2

upsuper:


sfackler:
Looks like a dup of #98?

upsuper:
Hmmm, seem to be yes.

This time the motivation is that Servo in general don't allow duplicate dependency except a few small crates, and there is one crate we want to use already upgraded to regex 0.2 so we need to upgrade that in all other dependencies. And you can see from servo/servo#14957 that this is the only remaining blocker now.

sfackler:
Ok, that seems reasonable. Let's close this and land the earlier PR then and cut an 0.4 env-logger release.

PR #104, opened 2017-03-08
Add ability to be used in rustc

alexcrichton:
I'm prototyping support for pulling in crates.io crates into rustc and I believe
this'll be the support necessary for living in the sysroot.


alexcrichton:
FWIW I intend to PR this commit: https://github.com/alexcrichton/rust/commit/d27bae535d6d3026716bedb06138d45ab5effafa

as soon as it passes tests

PR #105, opened 2017-03-29
Make log target configurable

dbrgn:
Right now `env_logger` always logs to stderr. This might be undesireable in some cases.

This PR adds a way to configure the log target. It can be set to stdout if desired.

Note that writing to stdout is implemented with `println!` instead of `writeln!`, so that the output is properly captured in tests. See https://users.rust-lang.org/t/cargo-doesnt-capture-stderr-of-running-tests/9222/.

I did not find out how to properly add a test, but tested it manually. If you want a test to be added, pointers would be welcome.

### DIFF
--- a/env/src/lib.rs
+++ b/env/src/lib.rs
@@ -152,11 +152,19 @@ mod filter;
 #[path = "string.rs"]
 mod filter;
 
+/// Log target, either stdout or stderr.
+#[derive(Debug)]
+pub enum LogTarget {
+    Stdout,
+    Stderr,
+}
+
 /// The logger.
 pub struct Logger {
     directives: Vec<LogDirective>,
     filter: Option<filter::Filter>,
     format: Box<Fn(&LogRecord) -> String + Sync + Send>,
+    target: LogTarget,
 }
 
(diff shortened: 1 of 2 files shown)


dbrgn:
@alexcrichton @sfackler any chance that someone could take a look at this? It would considerably improve test output (since right now all test output is intermingled with log messages that aren't properly captured). The change should be quick to review, only around 20 lines of code.

env/src/lib.rs
```diff
@@ -337,7 +356,12 @@ impl Log for Logger {
             }
         }
 
-        let _ = writeln!(&mut io::stderr(), "{}", (self.format)(record));
+        match self.target {
+            LogTarget::Stdout => println!("{}", (self.format)(record)),
```
sfackler: Should this be a `writeln!` to `io::stdout` instead? There are some differences around capture in the test harness. Not sure which approach would be more expected.

dbrgn: `writeln!` with `io::stdout` is not being captured by the test harness according to my tests.

sfackler:
Thanks!

PR #108, opened 2017-04-21
add emoji to log level name for iOS

king6cong:
because we can't tweak Xcode 8 to use color in console
https://github.com/robbiehanson/XcodeColors/issues/88

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -243,6 +243,8 @@ static MAX_LOG_LEVEL_FILTER: AtomicUsize = ATOMIC_USIZE_INIT;
 
 static LOG_LEVEL_NAMES: [&'static str; 6] = ["OFF", "ERROR", "WARN", "INFO",
                                              "DEBUG", "TRACE"];
+static LOG_LEVEL_EMOJIES: [&'static str; 6] = ["", "❤️", "💛", "💙",
+                                             "💚", "💜"];
 
 /// An enum representing the available verbosity levels of the logging framework
 ///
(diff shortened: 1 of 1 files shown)


king6cong:
image

sfackler:
This seems like something an individual logging implementation should do, not the base log crate. This will also break parsing since `LOG_LEVEL_NAMES` is used for that as well.

king6cong:
@sfackler Yeah, agree. 

please see also https://github.com/sfackler/log4rs/issues/44 😄

king6cong:
Updated, only add method to get emoji
Not sure if this should be in this base log crate, 
but I think it may be useful for multiple log implementations
Usage example maybe like this:
https://github.com/sfackler/log4rs/pull/45

sfackler:
How standard are these choices of emoji?

king6cong:
It is compatible with https://github.com/SwiftyBeaver/SwiftyBeaver

dtolnay:
This doesn't need to be provided in `log`. I agree with **@​sfackler** that individual logging implementations can select emojis if they want.

As far as I can tell there isn't a standard but :x::warning::information_source: is probably the most widely recognizable set and works way better in monochrome.

ThuggishNuggets:
Just a heads up, if you use this in your logging in Xcode you'll get some pretty buggy behaviors trying to use debugger commands as well as some pretty serious performance hits on the logging since Xcode's console text layout engine doesn't seem to like rendering emoji with fixed width fonts.

One bug that's really annoying with this is the autocomplete menus keep flickering and disappearing when using emojis in logging statements.

We ended up removing them from our implementation it was so bad, but YMMV.

PR #153, opened 2017-05-20
Replaced my_logger with env_logger in docs

### LINKED ISSUE #120: Use env_logger instead of my_logger in doc example
That way the example should be more easily copy-pasteable!

omh1280:
Fixes #120 

Ottavio

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -81,10 +81,10 @@
 //! extern crate env_logger;
 //!
 //! fn main() {
+//!     // Select env_logger, one possible logger implementation
+//!     // (see https://doc.rust-lang.org/log/env_logger/index.html)
 //!     env_logger::init().unwrap();
 //!     
-//!     // Set the RUST_LOG level prior to running the binary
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -78,13 +78,16 @@
...
 //! extern crate log;
-//! extern crate my_logger;
+//! extern crate env_logger;
 //!
 //! fn main() {
-//!     my_logger::init();
-//!
+//!     env_logger::init().unwrap();
+//!     
+//!     // Set the RUST_LOG level prior to running the binary
```
dtolnay: This sounds like it says the next line `info!("starting up")` sets the RUST_LOG level. Also RUST_LOG contains more than just the level.

I would prefer to put this above the previous line and say something like:

```rust
// Select env_logger, one possible logger implementation
// (see https://doc.rust-lang.org/log/env_logger/index.html)
env_logger::init().unwrap();
```

omh1280: Ok!

PR #156, opened 2017-05-20
Add an error type to FromStr impls

### LINKED ISSUE #119: FromStr impls for LogLevel, LogLevelFilter use () as error type
These implementations should use a dedicated opaque error type rather than `()`, and this error should implement the `Error` trait.

nivkner:
Fixes #119

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -340,14 +340,14 @@ fn eq_ignore_ascii_case(a: &str, b: &str) -> bool {
 }
 
 impl FromStr for LogLevel {
-    type Err = ();
-    fn from_str(level: &str) -> Result<LogLevel, ()> {
+    type Err =  LogLevelParseError;
+    fn from_str(level: &str) -> Result<LogLevel, LogLevelParseError> {
         ok_or(LOG_LEVEL_NAMES.iter()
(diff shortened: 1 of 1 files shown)


alexcrichton:
Awesome, thanks @nivkner! I think in light of https://github.com/rust-lang-nursery/log/issues/142 though we may wish to remove the `Log` prefix here perhaps?

@sfackler do you have thoughts on how to manage breaking changes like this? Do you think we should hold off on merging or maybe merge to a separate branch?

dtolnay:
I would merge to master along with a note in Cargo.toml.

```toml
publish = false # this branch contains breaking changes
```

nivkner:
Removed the `Log` prefix as requested.

alexcrichton:
Thanks @nivkner! We'll take it from here and figure out the best way to land this with other breaking changes to the `log` crate.

sfackler:
Seems fine to merge to master - we can always cut an 0.3.x branch if we need more releases from that version.

dtolnay:
@nivkner any reason these messages are slightly different? I would prefer to pick one or the other.

dtolnay:
Ah, this matches all of the existing error messages. I filed #166 to follow up.

PR #161, opened 2017-05-21
Add serde impls to LogLevel and LogLevelFilter

### LINKED ISSUE #143: Add serde impls for LogLevel/LogLevelFilter

aergonaut:
Fixes #143 

Adds `serde` as an optional dependency and puts the `Serialize` and `Deserialize` traits behind `cfg_attr`. I wasn't sure what version I should specify. 1.0.7 was the latest when I checked, but maybe it's too specific? Would `>= 1.0` be better?

Saw the post on the internals forum, wanted to help out!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,9 +8,11 @@ before_script:
   - pip install 'travis-cargo<0.2' --user && export PATH=$HOME/.local/bin:$PATH
 script:
   - cargo build --verbose
+  - cargo build --verbose --features serde
   - ([ $TRAVIS_RUST_VERSION != nightly ] || cargo build --verbose --no-default-features)
   - ([ $TRAVIS_RUST_VERSION != nightly ] || cargo build --verbose --features nightly)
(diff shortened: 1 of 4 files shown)


alexcrichton:
Thanks! Could you also add some CI configuration to test this on Travis/AppVeyor? Also, I think the dependency here is safe to just be `serde = "1.0"` as `>=` is actually a little too eager as it may or may not be compatible with 2.0 (we're not sure if it will be).

dtolnay:
@aergonaut are you still working on this or would you need to hand it off to someone else?

aergonaut:
I would still like to work on this! I will try to come back to it by the end of the week.

dtolnay:
No problem, let us know if you get stuck.

aergonaut:
@dtolnay @alexcrichton I think this is ready now! I added `cargo test --features serde` to the Travis and Appveyor configs. I also fixed up the feature flag usage so it actually works.

PR #163, opened 2017-05-23
Added keywords to log and env_logger Cargo.toml

budziq:
Hi,
I've added `keywords = ["logger", "log", "logging"]` to both log and env_logger.
These are the best I could came up with.

resolves https://github.com/rust-lang-nursery/log/issues/140


budziq:
done

Issue #123, opened 2017-05-19
Document `max_level_*`, `release_max_level_*` features in crate docs


alexcrichton:
These features of the crate control the compile-time max log level and should be documented!

dtolnay:
Fixed in #165.

PR #165, opened 2017-05-23
Documentation: added max_log_level per target

AndyGauge:
For #123


dtolnay:
Thanks! Could you check on the build failures?

AndyGauge:
If this doesn't resolve it I'll have to address it tomorrow.

PR #172, opened 2017-05-24
make error description consistent with display

### LINKED ISSUE #166: Inconsistent error Display and description() messages
All of our Error types have slightly different Display and description() representations.

- SetLoggerError
  - display: attempted to set a logger after the logging system was already initialized
  - description: set_logger() called multiple times
- ShutdownLoggerError
  - display: attempted to shut down the logger without an active logger
  - description: shutdown_logger() called without an active logger
- LevelParseError
  - display: attempted to convert a string that doesn't match an existing log level
  - description: called from_str() on a string without a matching log level

Is this part of an existing convention to make Display more narrative and description() more technical? If not, how about we pick a consistent message for each error?

nivkner:
Fixes #166

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -270,6 +270,11 @@ static MAX_LOG_LEVEL_FILTER: AtomicUsize = ATOMIC_USIZE_INIT;
 static LOG_LEVEL_NAMES: [&'static str; 6] = ["OFF", "ERROR", "WARN", "INFO",
                                              "DEBUG", "TRACE"];
 
+static SET_LOGGER_ERROR: &'static str = "attempted to set a logger after the logging system \
+                     was already initialized";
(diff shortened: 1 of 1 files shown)


budziq:
Hi @nivkner 
Your PR fails due to tests::test_error_trait not being updated with the new `description`

nivkner:
Fixed,  thanks.

nivkner:
nightly only problem  with `use` statements. Is that a change I should know about?

dtolnay:
Oh :fox_face: this is failing in the no_std build because Display is implemented but the `#[cfg(feature = "use_std")]` line means Error is not implemented, so Display cannot get the description from Error.

How about private statics instead? Used by both Display and Error.

```rust
static SET_LOGGER_ERROR: &'static str = "...";
```

Issue #141, opened 2017-05-19
Change html_root_url to https://docs.rs/$crate/$version


alexcrichton:
in src/lib.rs

morrme:
I can help with this.

dtolnay:
Great! How is this going @morrme? Let us know if you get stuck.

morrme:
@dtolnay Thanks! I was waiting for the go-ahead; just sent over a PR:

dtolnay:
#168 did this for `log`. Reopening to do the same for `env_logger` as well.

morrme:
would you like me to take care of that one too?

dtolnay:
Yes please do.

Issue #139, opened 2017-05-19
Update Cargo.toml documentation link to docs.rs


alexcrichton:
Should point do docs.rs/log

dtolnay:
This still needs to be done for env_logger.

morrme:
@dtolnay  I'm on it.

dtolnay:
Fixed in #160 and #175.

PR #173, opened 2017-05-24
update html_root_url for env_logger

morrme:
Closes: #141

@dtolnay there is a `documentation` line in `env/Cargo.toml` here : https://github.com/morrme/log/blob/issue141/env/Cargo.toml#L7  

Should that be updated as well?


dtolnay:
Yes. I reopened #139 to track that.

PR #177, opened 2017-05-27
Add more documentation to Metadata struct

MarkMcCaskey:
The example is pretty redundant (copied almost verbatim from another example).  Let me know what  I should add or remove to make this better!

Thanks!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -568,6 +568,46 @@ impl<'a> Record<'a> {
 }
 
 /// Metadata about a log message.
+///
+/// # Use
+///
+/// `Metadata` structs are created when users of the library use
+/// logging macros.
+///
+/// They are consumed by implementations of the `Log` trait in the
+/// `enabled` method. 
+///
+/// `Record`s use `Metadata` to determine the log message's severity
+/// and target.
+///
+///
+/// Users should use the `log_enabled!` macro in their code to avoid
+/// constructing expensive log messages.
+///
+/// ## Examples
+///
+/// ```rust
+/// # #[macro_use] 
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -568,6 +568,46 @@ impl<'a> Record<'a> {
...
+/// `Metadata` structs are created when users of the library use
+/// logging macros.
+///
+/// They are consumed by implementations of the `Log` trait in the
+/// `enabled` method. 
+///
+/// `Record`s use `Metadata` to determine the log message's severity
+/// and target.
+///
+///
```
dtolnay: Please remove the extra blank line. Rustdoc treats this just like one blank line anyway.

src/lib.rs
```diff
@@ -568,6 +568,46 @@ impl<'a> Record<'a> {
...
+/// Users should use the `log_enabled!` macro in their code to avoid
+/// constructing expensive log messages.
+///
+/// ## Examples
```
dtolnay: Please use a full-sized heading like we do [elsewhere](https://github.com/rust-lang-nursery/log/blob/50ce9a648f2f0c8089cf52e32d78c2741d3ac3ce/src/macros.rs#L54).

src/lib.rs
```diff
@@ -568,6 +568,46 @@ impl<'a> Record<'a> {
...
+///
+/// ```rust
+/// # #[macro_use] 
+/// # extern crate log;
+///
```
dtolnay: This will include a blank line at the top of the example code. Please put `#` on this line to hide it.

selection_053

Issue #129, opened 2017-05-19
LogRecord doc improvements


alexcrichton:
- How are these values created?
  - Are these for logger implementors or log user?
  - What's the difference between the 'level'/'target' accessors here
    and on the attached LogMetadata?
  - Add an example

jehiggs:
Hey! I'd like to have a go at this, so long as it's not taken already?

dtolnay:
Thanks, go for it @jehiggs.

jehiggs:
Alright, pull request is [#178](https://github.com/rust-lang-nursery/log/pull/178)

dtolnay:
Fixed in #178.

PR #178, opened 2017-05-27
Record Doc Improvements

jehiggs:
This addresses the doc improvements suggested by [#129](https://github.com/rust-lang-nursery/log/issues/129) for the Record structure.

Let me know if there's anything missing or anything that can be improved!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -530,6 +530,8 @@ impl LevelFilter {
 
 /// The "payload" of a log message.
 ///
+/// # Use
+///
 /// `Record` structures are passed as parameters to the [`log`] method of the
 /// [`Log`] trait. Logger implementors manipulate these structures in order to
 /// display log messages. `Record`s are automatically created by the [`log!`]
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -528,11 +528,51 @@ impl LevelFilter {
...
+///
+///        println!("{}:{} -- {}",
+///                 record.level(),
+///                 record.location().module_path(),
+///                 record.args());
+///    }
+/// }
+/// ```
+///
+/// [`log`]: trait.Log.html#method.log
```
dtolnay: Markdown links are case insensitive so you can't have both [\`log\`] and [\`Log\`]. How about the following?

```markdown
bla bla [`log`][method.log] method

[method.log]: trait.Log.html#method.log
```

dtolnay:
Thank you!

src/lib.rs
jehiggs: I didn't realize that - thanks!

Issue #127, opened 2017-05-19
LogLocation doc improvements


alexcrichton:
- How are these values created?
  - Are these for logger implementors or log user?
  - Add an example

jehiggs:
Mind if I have a go at this one too?

dtolnay:
Go ahead @jehiggs.

dtolnay:
Fixed in #179.

PR #179, opened 2017-05-27
Update docs for Location struct and add example

jehiggs:
This provides the doc updates as suggested in [#127](https://github.com/rust-lang-nursery/log/issues/127).

I realize I've accidentally introduced some whitespace changes; I can revert these if required.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -615,7 +615,7 @@ impl<'a> Record<'a> {
 /// logging macros.
 ///
 /// They are consumed by implementations of the `Log` trait in the
-/// `enabled` method. 
+/// `enabled` method.
 ///
 /// `Record`s use `Metadata` to determine the log message's severity
 /// and target.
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -694,11 +694,50 @@ impl Log for NopLogger {
...
+/// ```
+///
 /// # Warning
 ///
 /// The fields of this struct are public so that they may be initialized by the
-/// `log!` macro. They are subject to change at any time and should never be
+/// [`log!`] macro. They are subject to change at any time and should never be
 /// accessed directly.
+///
+/// [`log!`]: macro.log!.html
```
dtolnay: `macro.log!.html` redirects to `macro.log.html` so you would save the user a redirect by pointing to `macro.log.html` directly.

dtolnay:
Great work!

Next time please include "Fixes #127" in the comment when you open the PR so GitHub auto-closes the right issue when the PR is merged.

PR #185, opened 2017-05-30
max_level doc improvements

rap2hpoutre:
see: https://github.com/rust-lang-nursery/log/issues/134

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -803,9 +803,16 @@ impl MaxLevelFilter {
 
 /// Returns the current maximum log level.
 ///
-/// The `log!`, `error!`, `warn!`, `info!`, `debug!`, and `trace!` macros check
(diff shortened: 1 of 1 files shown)


rap2hpoutre:
Or maybe https://github.com/rust-lang-nursery/log/issues/125 I'm not sure anymore. Anyway, it improves max_level docs!

rap2hpoutre:
@dtolnay Thanks! 

You wrote:

> The MaxLevel could also be linked but we can follow up separately.

Ok, I can do this! The [documentation](https://doc.rust-lang.org/log/log/fn.max_level.html) says:

> The maximum log level is set by the `MaxLevel` token passed to loggers. 

Should it refer to [`set_logger`](https://doc.rust-lang.org/log/log/fn.set_logger.html)? or [`MaxLevelFilter` `set` method](https://doc.rust-lang.org/log/log/struct.MaxLevelFilter.html#method.set)?

PR #188, opened 2017-06-02
Added links to logging frameworks

budziq:
solves https://github.com/rust-lang-nursery/log/issues/115

- I've left env_logger on the list as the most popular implementation (as it is to be separated from main log repo it should not be penalized)
- I've chosen links to docs.rs over crates.io. But I do not have strong preference
- the "simple" section was a little crowded so I've pruned the least popular crates but sadly it has left us with [`simple_logger`] and [`pretty_env_logger`] which do not provide nice docs

I am eager hear any suggestions.

[`simple_logger`]: https://docs.rs/simple_logger/*/simple_logger/
[`pretty_env_logger`]: https://docs.rs/pretty_env_logger/*/pretty_env_logger/

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -92,7 +92,27 @@
 //! }
 //! ```
 //!
-//! # Logger implementations
+//! # Available logging implementations
+//!
+//! In order to produce log output executables have to use
+//! a logger implementation compatible with the facade.
+//! There are many available implementations to chose from,
+//! here are some of the most popular ones:
+//!
+//! * Simple minimal loggers:
+//!     * [env_logger]
+//!     * [simple_logger]
+//!     * [simplelog]
+//!     * [stderrlog]
+//!     * [flexi_logger]
+//! * Complex configurable frameworks:
+//!     * [log4rs]
+//!     * [fern]
+//! * Adaptors for other facilities:
+//!     * [syslog]
+//!     * [slog-stdlog]
+//!
+//! # Implementing a Logger
 //!
 //! Loggers implement the [`Log`] trait. Here's a very basic example that simply
 //! logs all messages at the [`Error`][level_link], [`Warn`][level_link] or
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -208,6 +229,16 @@
 //! [`MaxLevelFilter`]: struct.MaxLevelFilter.html
 //! [`set_logger_raw`]: fn.set_logger_raw.html
 //! [`shutdown_logger_raw`]: fn.shutdown_logger_raw.html
+//! [env_logger]: https://docs.rs/env_logger/*/env_logger/
+//! [simple_logger]: https://docs.rs/simple_logger/*/simple_logger/
```
dtolnay: https://github.com/borntyping/rust-simple_logger would be a better landing page.

src/lib.rs
```diff
@@ -92,6 +92,27 @@
 //! }
 //! ```
 //!
+//! # Available logging frameworks
+//!
+//! In order to produce log output executables have to utilize
```
dtolnay: "Utilize" is not the right word. See https://english.stackexchange.com/questions/143941/when-to-use-use-and-when-to-use-utilize-in-a-sentence/.

src/lib.rs
```diff
@@ -208,6 +229,16 @@
 //! [`MaxLevelFilter`]: struct.MaxLevelFilter.html
 //! [`set_logger_raw`]: fn.set_logger_raw.html
 //! [`shutdown_logger_raw`]: fn.shutdown_logger_raw.html
+//! [env_logger]: https://docs.rs/env_logger/*/env_logger/
+//! [simple_logger]: https://docs.rs/simple_logger/*/simple_logger/
+//! [simplelog]: https://docs.rs/simplelog/*/simplelog/
```
dtolnay: https://github.com/drakulix/simplelog.rs would be a better landing page.

src/lib.rs
```diff
@@ -92,6 +92,27 @@
...
+//! In order to produce log output executables have to utilize
+//! a framework compatible with the facade.
+//! There are many available logging implementations to chose from.
+//! Here are some of the most popular ones:
+//!
+//! * Simple minimal loggers:
+//!     * [env_logger]
+//!     * [simple_logger]
+//!     * [simplelog]
+//!     * [pretty_env_logger]
```
dtolnay: Please remove this from the list for now and file an issue asking the author to add documentation. As is, it doesn't seem intended for people to use.

src/lib.rs
```diff
@@ -92,6 +92,27 @@
 //! }
 //! ```
 //!
+//! # Available logging frameworks
+//!
+//! In order to produce log output executables have to utilize
+//! a framework compatible with the facade.
```
dtolnay: Almost none of these are a "framework." Maybe only `log4rs`. Something like `env_logger` is definitely not a "framework." I think above it refers to these using the word "implementation."

> The `log` crate provides a single logging API that abstracts over the actual logging implementation.

src/lib.rs
budziq: Filed an issue against the repo https://github.com/seanmonstar/pretty-env-logger/issues/2 already got positive response from the author :)

src/lib.rs
budziq: Huh who would know? Thanks!

src/lib.rs
budziq: I was quite at loss in regard to wording as https://docs.rs/log/0.3.8/log/ landing page most refers to "implementations" as "frameworks" I agree that my wording was amiss. 

@dtolnay Maybe rest of the docs should be updated to tone down the "framework" wording?

budziq:
Done. Also updated the wording on following heading from "Logger Implementation" to "Implementing a Logger" to avoid weird stuttering between sections

src/lib.rs
dtolnay: Thanks, I didn't notice how pervasive it is in the existing documentation. I filed #191 to follow up.

PR #189, opened 2017-06-02
add link to from_str method in ParseLevelError

nivkner:
Has no existing issue.
Just thought it might need a link like the other items.


nivkner:
So... @dtolnay, could you also check this PR?
I assume this didn't get noticed while #190 got merged is because it doesn't relate to any issue.

PR #195, opened 2017-06-04
Toned down the references to the "frameworks"

budziq:
fixes https://github.com/rust-lang-nursery/log/issues/191

As usual my English might not be up to snuff, so I would be glad for any suggestions.

Also added link to [pretty_env_logger](https://docs.rs/pretty_env_logger/*/pretty_env_logger/) as it is one of the most popular logger sinks and my changes to its docs have been published now (related to https://github.com/rust-lang-nursery/log/issues/115).


brson:
Thanks @budziq

PR #196, opened 2017-06-05
remove shutdown and add `flush` to `Log` trait

### LINKED ISSUE #117: Remove logger shutdown
This adds some complexity to the logger plus extra cost on all logger calls. In our discussion of this issue the review team also concluded that it would be prudent to maybe add a top-level flushing method to explicitly ensure that a common part of shutdown, flushing, can still be manually performed. Flushing would be a noop by default.

danielscottt:
Fixes #117

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -150,6 +150,7 @@
 //!             println!("{} - {}", record.level(), record.args());
 //!         }
 //!     }
+//!     fn flush(&self) {}
 //! }
 //!
 //! # fn main() {}
@@ -173,6 +174,7 @@
 //! # impl log::Log for SimpleLogger {
 //! #   fn enabled(&self, _: &Metadata) -> bool { false }
 //! #   fn log(&self, _: &log::Record) {}
+//! #   fn flush(&self) {}
 //! # }
 //! # fn main() {}
 //! # #[cfg(feature = "use_std")]
@@ -189,19 +191,15 @@
 //! `default-features = false` when specifying the dependency in `Cargo.toml`.
 //! This makes no difference to libraries using `log` since the logging API
 //! remains the same. However executables will need to use the [`set_logger_raw`]
-//! function to initialize a logger and the [`shutdown_logger_raw`] function to
-//! shut down the global logger before exiting:
+//! function to initialize a logger.
 //!
 //! ```rust
 //! # extern crate log;
-//! # use log::{Level, LevelFilter, SetLoggerError, ShutdownLoggerError,
-//! #           Metadata};
+//! # use log::{Level, LevelFilter, Log, SetLoggerError, Metadata};
 //! # struct SimpleLogger;
 //! # impl log::Log for SimpleLogger {
 //! #   fn enabled(&self, _: &Metadata) -> bool { false }
 //! #   fn log(&self, _: &log::Record) {}
-//! # }
-//! # impl SimpleLogger {
 //! #   fn flush(&self) {}
 //! # }
 //! # fn main() {}
@@ -213,12 +211,6 @@
 //!             &SimpleLogger
 //!         })
 //!     }
-//! }
-//! pub fn shutdown() -> Result<(), ShutdownLoggerError> {
-//!     log::shutdown_logger_raw().map(|logger| {
-//!         let logger = unsafe { &*(logger as *const SimpleLogger) };
-//!         logger.flush();
-//!     })
 //! }
 //! ```
 //!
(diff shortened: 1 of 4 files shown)


mrhota:
@danielscottt Wasn't _main_ idea of #117 to remove shutdown? I thought exposing flush functionality was more of a since-we're-in-the-neighborhood sort of thing.

danielscottt:
@mrhota It is quite possible that I misunderstood the issue altogether.

src/lib.rs
```diff
@@ -1153,12 +1155,6 @@ mod panic {
 
 struct LoggerGuard(&'static Log);
 
-impl Drop for LoggerGuard {
```
sfackler: This synchronization needs to stick around, since `Log` isn't an `unsafe` trait.

danielscottt: What would the implementation of `Drop` synchronize on for `LoggerGuard`?  Is the idea to synchronize all `__log` calls around flush?   Is that how the need for `shutdown` is removed?

sfackler: The refcounting logic ensures there aren't any log calls actively using the logger while we tear it down, which would be a use-after-free. We can't depend on flush doing this synchronization for us - the default implementation does nothing at all for example!

We're removing shutdown because it's not sufficiently useful compared to its cost in all of this refcounting. The addition of `flush` is only semi-related - it's just a way to allow people to make sure that any buffers or whatever are flushed before their application exits so they don't drop messages on the floor.

danielscottt: Thanks for the guidance, here!  As I feared, I fundamentally misunderstood the ask.  I'll clean this up and repush shortly.

manuthambi: @sfackler Perhaps I am totally confused, but why would you need that synchronization? Without shutdown, the logger is never deallocated (or dropped), so we are never going to tear it down?

danielscottt: This is why I left it here after I refactored my first pass, which misunderstood the task. But on account of my initial confusion, I can certainly imagine that I'm missing something.

manuthambi: Also, I think the load/stores from `STATE` only need `Ordering::Acquire`/`Release` not `SeqCst`. Using `SeqCst` synchronizes with any unrelated `SeqCst` read/writes somewhere else in the program.

sfackler: Oh, did I totally miss the fact that this PR also removed shutdown? Oops >_>

src/lib.rs
```diff
@@ -716,6 +692,9 @@ pub trait Log: Sync + Send {
     /// Implementations of `log` should perform all necessary filtering
     /// internally.
     fn log(&self, record: &Record);
+
+    /// Flushes any buffered records.
+    fn flush(&self) {}
```
sfackler: We should maybe not have a default implementation here? `Write::flush` does't have a default as a comparison.

brson:
Thanks for reviewing @sfackler.

src/lib.rs
danielscottt: The [issue]( https://github.com/rust-lang-nursery/log/issues/117) mentioned `flush` ought to have a default implementation, but I'm happy to remove it here. This obviously has ramifications for `env_logger`; would its implementation be expected to use the same reference counting pattern previously used here when coordinating on `shutdown`?

danielscottt: I'll go ahead and prep a patch that does this in the meantime.

sfackler: I think the issue mentioned a default impl to retain back compat, which we don't really need to worry about since we're removing shutdown at the same time.

I don't think we'd want to add the refcounting into the flush impl for env_logger - it'd defeat the purpose of getting rid of shutdown in the first place! The use case for flush is when you're for example, in the shutdown process of your application. If you use something like env_logger then just killing the process is fine - any log calls that finished before the exit will end up on the console. But, if you're using a logger that performs the IO asynchronously, you can end up in situations where log lines that should have been output aren't. Imagine something like

```rust
info!("shutting down");
log::flush();
process::exit(0);
```

All we need to make sure is that the "shutting down" message makes it all the way through the logging infrastructure. If there are other threads doing things in the background, the state of their messages is not super important since they're racing with this thread.

brson:
Looks like this is ready to go. Thanks @danielscottt!

@sfackler can you take another look?

src/lib.rs
```diff
@@ -200,11 +198,8 @@
 //!         })
 //!     }
 //! }
-//! pub fn shutdown() -> Result<(), ShutdownLoggerError> {
-//!     log::shutdown_logger_raw().map(|logger| {
-//!         let logger = unsafe { &*(logger as *const SimpleLogger) };
-//!         logger.flush();
-//!     })
+//! pub fn shutdown() {
```
sfackler: Let's just remove this bit of the example.

src/lib.rs
```diff
@@ -1178,12 +1092,6 @@ mod panic {
 
 struct LoggerGuard(&'static Log);
```
sfackler: This type can be removed entirely now that we don't need a destructor, but that can be handled later.

src/lib.rs
```diff
@@ -715,6 +696,9 @@ pub trait Log: Sync + Send {
     /// Implementations of `log` should perform all necessary filtering
     /// internally.
     fn log(&self, record: &Record);
+
+    /// Flushes any buffered records.
+    fn flush(&self) {}
```
sfackler: Looks like we still need to remove the default impl here.

aturon:
Ping @sfackler

sfackler:
Thanks!

sfackler:
Sorry for the very slow review cycle D:

PR #198, opened 2017-06-08
expanded log! docs for `max_level_*`

budziq:
due to https://github.com/rust-lang-nursery/log/issues/197 this might be insufficient or even overly verbose if the actual descriptions for "Crate feature flags" would be documented thoroughly in one place. 

fixes https://github.com/rust-lang-nursery/log/issues/125


brson:
Thanks @budziq !

PR #199, opened 2017-06-15
set_logger wants to panic

brson:
This is the same as https://github.com/rust-lang-nursery/log/pull/187. Seeing if the CI is any less confused by this PR. cc @alisha17


brson:
The answer is 'no'.

PR #200, opened 2017-06-20
Create builders for `Metadata`, `Record`, and `Level`.

### LINKED ISSUE #116: Support construction of `Record`s and `Metadata`
Testing and "shim loggers" typically want to do this so we should support these constructions! We probably want to have the construction go through builders to ensure that we can flexibly add more fields in the future. 

I'm not tagging this issue as "easy" just yet as it's got some API design involved, but if we get a concrete design listed here I think it'd be great to tag this issue as "easy" afterwards for an implementation to come by.

omh1280:
Fixes #116.

Create builders to support the construction of `Metadata`, `Record`, and `Level`. I decided to take `mut self` as the parameter for each builder method because it allows:
```rust
let record = RecordBuilder::new().args(...).metadata(...).build()
```
Whereas taking `&mut self` leads to the temporary object created by `RecordBuilder::new()` being dropped before `build()`.

Not done yet--still need to add examples/docs. Please review everything else!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -585,7 +585,7 @@ impl LevelFilter {
 ///
 ///        println!("{}:{} -- {}",
 ///                 record.level(),
-///                 record.location().module_path(),
+///                 record.module_path(),
 ///                 record.args());
 ///    }
 ///    fn flush(&self) {}
@@ -597,14 +597,20 @@ impl LevelFilter {
 /// [`log!`]: macro.log.html
 /// [`level()`]: struct.Record.html#method.level
 /// [`target()`]: struct.Record.html#method.target
-#[derive(Debug)]
+#[derive(Clone, Debug)]
 pub struct Record<'a> {
     metadata: Metadata<'a>,
-    location: &'a Location,
     args: fmt::Arguments<'a>,
+    module_path: &'static str,
+    file: &'static str,
+    line: u32,
 }
 
 impl<'a> Record<'a> {
+    /// Returns a new builder
+    pub fn builder() -> RecordBuilder<'a> {
+        RecordBuilder::new()
+    }
     /// The message body.
     pub fn args(&self) -> &fmt::Arguments<'a> {
         &self.args
@@ -615,11 +621,6 @@ impl<'a> Record<'a> {
         &self.metadata
     }
 
-    /// The location of the log directive.
-    pub fn location(&self) -> &Location {
-        self.location
-    }
-
     /// The verbosity level of the message.
     pub fn level(&self) -> Level {
         self.metadata.level()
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -641,6 +641,69 @@ impl<'a> Record<'a> {
...
 }
 
+/// Builder for [`Record`](struct.Record.html).
+///
+/// Typically should only be used by log library creators or for testing and "shim loggers".
+/// The `RecordBuilder` can set the different parameters of `Record` object, and returns
+/// the created object when `build` is called.
+///
+/// # Example
+/// ```rust
```
brson: Nit: can you add a blank line here?

src/lib.rs
```diff
@@ -641,6 +641,69 @@ impl<'a> Record<'a> {
...
+///
+/// let error_metadata = MetadataBuilder::new()
+///                         .target("myApp")
+///                         .level(Level::Error)
+///                         .build();
+///
+/// let my_record = RecordBuilder::new(&error_location)
+///                         .args(format_args!("Error!"))
+///                         .metadata(error_metadata)
+///                         .build();
```
brson: Example is missing closing backticks.

brson:
@omh1280 thank you.

This all looks right to me with the exception that the builder pattern we tend to prefer is with `&mut self` instead of `self`. Unless there's a exceptional reason to use `self` here can you switch them to `&mut self`?

cc @alexcrichton can you peek at this? Do you agree that the builders should be by ref?

src/lib.rs
```diff
@@ -699,6 +762,54 @@ impl<'a> Metadata<'a> {
...
+/// the created object when `build` is called.
+///
+/// # Example
+/// ```rust
+/// let target = "myApp";
+/// use log::{Level, MetadataBuilder};
+/// let mut level_test = MetadataBuilder::new()
+///                         .level(Level::Debug)
+///                         .target(target)
+///                         .build();
```
brson: Ditto about the blank line and missing backticks.

src/lib.rs
```diff
@@ -801,6 +912,63 @@ impl Location {
     }
 }
 
+/// A builder for [`Location`](struct.Location.html).
+///
+/// Often used with [`RecordBuilder`](struct.RecordBuilder.html)'s [`new`], which
+/// takes a `&Location`.
+///
+/// # Example
+/// ```rust
```
brson: Please add blank line here too.

alexcrichton:
> cc @alexcrichton can you peek at this? Do you agree that the builders should be by ref?

Ah yeah I agree that these should be `&mut self` across the board

omh1280:
Gotcha. I'll update next time I get a chance!

omh1280:
Fixed! 

- I had to add `Copy` and `Clone` to prevent `cannot move out of borrowed content` in the `build` functions. Is this OK? Is there a better way to do this?
- As far as I can tell, `RecordBuilder::new` _needs_ to take a reference to `Location` because it cannot create a default which has the appropriate lifetime. 
- Finally, is there any reason [`Record`](https://doc.rust-lang.org/log/src/log/lib.rs.html#611) contains a reference to `Location` instead of owning it?

Thanks for the review 😄

brson:
Thanks @omh1280!

ping @alexcrichton

brson:
Oh, I guess I can continue the review myself...

brson:
> I had to add Copy and Clone to prevent cannot move out of borrowed content in the build functions. Is this OK? Is there a better way to do this?

In this case, this seems fine. I don't see a reason for Record and Metadata not to be Copy/Clone. @alexcrichton what you think? An alternative would be for build to take `self`, but it looks to me like we prefer for builders to be reusable when possible.

> As far as I can tell, RecordBuilder::new needs to take a reference to Location because it cannot create a default which has the appropriate lifetime.

Yes.

> Finally, is there any reason Record contains a reference to Location instead of owning it?

I suspect this a subtle efficiency matter. This type is passed around during logging and copying one pointer is faster than copying the entire struct.

This patch looks fine to me now, but I'd be more confident if @alexcrichton took another look.

alexcrichton:
I think we may want to remove `Copy` as it may be a bit restrictive, but adding `Clone` to these types seems fine by me!

Some other points I'd have:

* In addition to `RecordBuilder::new` perhaps we could also have something like `Record::builder()`? I think @sfackler has pioneered this pattern in the past and it can sometimes help the ergonomics of importing less items.
* Could you add to the documentation the default value for each builder method?

omh1280:
Thanks Alex! Let me know if it still needs work (or if you want me to rebase to clean up commits)

alexcrichton:
Thanks @omh1280! @sfackler I'm curious, what do you think about the 3 builders here vs trying to collapse them all into one? Or perhaps having one builder with methods that delegate to the others?

src/lib.rs
```diff
@@ -815,6 +954,72 @@ impl Location {
...
+    ///
+    /// - `module_path`: [`module_path!()`]
+    /// - `file`: [`file!()`]
+    /// - `line`: 0
+    /// [`module_path!()`]: https://doc.rust-lang.org/std/macro.module_path.html
+    /// [`file!()`]: https://doc.rust-lang.org/std/macro.file.html
+    pub fn new() -> LocationBuilder {
+        LocationBuilder {
+            location: Location {
+                __module_path: module_path!(),
```
sfackler: These defaults are not going to be particularly useful - I'd probably lean towards making them empty strings.

sfackler:
I can't actually remember why `Location` exists as a separate construct from `Record` I'd probably say we just inline those fields?

It does seem nice to have forwarding methods from the `Record` builder to the `Metadata` builder though.

omh1280:
To confirm: 
- Move `__module_path`, `__file`, and `__line` into `Record` and scrap `Location`? (and change macros etc.)
- Add forwarding methods so `RecordBuilder` can build entirely without `MetadataBuilder`?

sfackler:
Yep!

If we're changing the macros, we may want to change the `__log` function to be publicly exposed and take a `LogRecord` directly, now that you can construct that with the public API.

omh1280:
Sorry for the long wait--I'm in the middle of doing this I promise :)

omh1280:
Updated!

This was a pretty big change, please let me know if I forgot something.

- Remove `Location` completely
- Inline `module_path`, `line`, and `file` into `Record`
- Forward `target` and `level` functions in `RecordBuilder` to `MetadataBuilder`

You now don't have to `use` any builder structs:
```rust
use super::{Record, Level};
let record = Record::builder()
    .module_path("foo")
    .file("bar")
    .line(30)
    .target("myApp") // forwarded
    .level(Level::Error) // forwarded
    .build();
```

I didn't make `__log` public: I can do that if that's agreed on /cc @alexcrichton

alexcrichton:
This looks great to me! I'll leave @sfackler to sign off though.

@omh1280 it looks like there may be some CI failures as well?

omh1280:
Didn't see that--fixed!

sfackler:
Looks good to me but it needs a rebase.

omh1280:
Squashed + rebased

PR #201, opened 2017-06-21
logging levels and corresponding macros

jesusbv:
As discussed in [issue 121](https://github.com/rust-lang-nursery/log/issues/121),

The loggin levels are mentioned in the use section

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -26,18 +26,19 @@
 //!
 //! # Use
 //!
-//! ## Logging levels
+//! The basic use of the log crate is through the five logging macros: [`error!`],   
+//! [`warn!`], [`info!`], [`debug!`] and [`trace!`] 
+//! where `error!` represents the highest-priority log level, and `trace!` the lowest.  
+//!
+//! Each of these macros accept format strings similarly to [`println!`].  
 //! 
-//! This library makes use of the following macros:
 //! 
-//! - [error](https://github.com/rust-lang-nursery/log/blob/master/src/macros.rs#L73)
-//! - [warn](https://github.com/rust-lang-nursery/log/blob/master/src/macros.rs#L104)
-//! - [info](https://github.com/rust-lang-nursery/log/blob/master/src/macros.rs#L138)
-//! - [debug](https://github.com/rust-lang-nursery/log/blob/master/src/macros.rs#L172)
-//! - [trace](https://github.com/rust-lang-nursery/log/blob/master/src/macros.rs#L209)
-//! 
-//! All of these macros accept formatting strings
-//! similarly to [println!](https://github.com/rust-lang/rust/blob/master/src/libstd/macros.rs#L127)
+//! [`error!`]: ./macro.error.html   
+//! [`warn!`]: ./macro.warn.html  
+//! [`info!`]: ./macro.info.html  
+//! [`debug!`]: ./macro.debug.html   
+//! [`trace!`]: ./macro.trace.html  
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -26,6 +26,19 @@
...
+//! This library makes use of the following macros:
+//! 
+//! - [error](https://github.com/rust-lang-nursery/log/blob/master/src/macros.rs#L73)
+//! - [warn](https://github.com/rust-lang-nursery/log/blob/master/src/macros.rs#L104)
+//! - [info](https://github.com/rust-lang-nursery/log/blob/master/src/macros.rs#L138)
+//! - [debug](https://github.com/rust-lang-nursery/log/blob/master/src/macros.rs#L172)
+//! - [trace](https://github.com/rust-lang-nursery/log/blob/master/src/macros.rs#L209)
+//! 
+//! All of these macros accept formatting strings
+//! similarly to [println!](https://github.com/rust-lang/rust/blob/master/src/libstd/macros.rs#L127)
```
brson: Thanks a bunch @jesusbv, and sorry for taking so long to get back to you.

I wonder if you might take a slightly different approach here, and instead of creating a new 'logging levels' section, write some prose that gets this same information across, setting the stage for the detail that comes in following section.

It might begin like

```
# Use

The basic use of the log crate is through the five logging macros: [`error!`], [`warn!`], { etc }, where `error!` represents the highest-priority log level, and `trace!` the lowest. Each of these macros accept format strings similarly to `println!`.
```

For the hyperlinks you'll want to use rustdoc links, not code links; for the internal links, relative links, and the std links absolute links.

So the `error!` link would probably be `./macro.error.html` and the `println!` link `https://doc.rust-lang.org/stable/std/macro.println.html`.

If you convert the list to prose you'll probably want to use out-of-line links like

```
The [`error!`] macro.

[`error!`]: ./macro.error.html
```

brson:
Thanks a bunch @jesusbv. Sorry for the long delay. Feel free to ping me on irc or email any time.

src/lib.rs
jesusbv: Thanks for the update @brson. Sorry not to reply sooner.

I have converted it to prose version and updated the links.

src/lib.rs
```diff
@@ -26,6 +26,20 @@
 //!
 //! # Use
 //!
+//! The basic use of the log crate is through the five logging macros: [`error!`],   
+//! [`warn!`], [`info!`], [`debug!`] and [`trace!`] 
+//! where `error!` represents the highest-priority log level, and `trace!` the lowest.
```
brson: Can you put a comma before 'where'?

src/lib.rs
```diff
@@ -148,8 +161,8 @@
 //! improve performance of log messages at levels that are disabled. In the
 //! case of our example logger, we'll want to set the maximum log level to
 //! [`Info`][level_link], since we ignore any [`Debug`][level_link] or
-//! [`Trace`][level_link] level log messages. A logging implementation
-//! should provide a function that wraps a call to [`set_logger`], handling
+//! [`Trace`][level_link] level log messages. A logging framework should
+//! provide a function that wraps a call to [`set_logger`], handling
```
brson: These changes of "implementation" to "framework" seem indicative of a problem. Previously there was a patch merged that converted "framework" to "implementation", and I assume you did not mean to change these all to "framework". Can you take a look at this diff and see if all these changes are intended? It may be prudent to re-start from the master branch and just make the intended change again.

PR #203, opened 2017-06-22
Enable use of Logger as a filter

jethrogb:
The spec parsing and matching of env_logger is very useful, but it will only let you log to stdout/stderr. With this change you can embed env_logger's logic in your own logger.

### DIFF
--- a/env/src/lib.rs
+++ b/env/src/lib.rs
@@ -157,6 +157,7 @@ mod filter;
 pub enum Target {
     Stdout,
     Stderr,
+    Silent,
 }
 
 /// The logger.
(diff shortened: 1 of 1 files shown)


jethrogb:
ping @sfackler @alexcrichton @huonw

jethrogb:
Refiled as https://github.com/sebasmagri/env_logger/pull/4

jethrogb:
Re-opening this while there's still uncertainty about splitting off of env_logger

KodrAus:
`env_logger` is all spun-out now and @jethrogb has implemented this over in https://github.com/sebasmagri/env_logger/pull/4.

Thanks again for getting this through @jethrogb!

PR #219, opened 2017-08-14
[env_logger] Adds a newline to format string before writing

tailhook:
This is important so in unix systems most of the time buffer is output
atomically (until certain length of message). Which means you don't get
two lines squashed together if they are written simultaneously.

I believe most of the time it's doesn't influence performance, because
underlying buffer of string is probably overallocated a bit anyway.

### DIFF
--- a/env/src/lib.rs
+++ b/env/src/lib.rs
@@ -371,7 +371,8 @@ impl Log for Logger {
         match self.target {
             Target::Stdout => println!("{}", (self.format)(record)),
             Target::Stderr => {
-                let _ = writeln!(&mut io::stderr(), "{}", (self.format)(record));
+                let value = (self.format)(record) + "\n";
+                let _ = io::stderr().write_all(value.as_bytes());
             },
         };
     }


sfackler:
Stdout is line buffered - when will this change behavior?

tailhook:
> Stdout is line buffered - when will this change behavior?

Is it? I observe broken logs every time I start two processes. Logs are written to stderr by default, though. Is there any difference between stdout and stderr?

tailhook:
Well, it looks like undocumented, but yes, stdout is line buffered whereas stderr isn't. So we could revert to the old behavior on stdout.

tailhook:
Done. Also fixed bug. (I used to think that it's enough to run tests in root directory rather than `env`)

KodrAus:
Thanks for this @tailhook! Do you mind rebasing this over at https://github.com/sebasmagri/env_logger, which is the new home for `env_logger`?

tailhook:
Well, it's not so useful after: https://github.com/sebasmagri/env_logger/pull/19

I'm not sure what to do:
1. Now this behavior depends on actual formatter. Default one is wrong at this respect, and also a bad example to do.
2. On the other hand, if default formatter fixed it makes the aformentioned PR useless?

KodrAus:
We've started discussing this over in https://github.com/sebasmagri/env_logger/issues/22 so I'll close this PR now. I'd definitely like to come up with a good solution to this before a `1.0` of `env_logger`!

Issue #218, opened 2017-08-14
Figure out macro invocation syntax for the new release


sfackler:
There are a couple of constraints here, IMO:

* It must be possible to turn `println!(...)` into `debug!(...)` with no changes to the inner bit.
* It must be possible to optionally specify a target string.
* It must be possible to optionally specify a set of parameters.
    * We may or may not actually support this in the initial release, but I want to make sure the macro syntax will work out for it.

The current syntax is as println with an optional `target: $expr,` at the start to specify the target. I don't really think that sticking things before the formatting part is going to work for parameters, so we'll need some other way of working with that.

One possibility is that string interpolation isn't supported when providing parameters, since all of the dynamic info should be provided in the parameters. Not sure how people would feel about that though.

cc @alexcrichton @dtolnay @dpc

alexcrichton:
Is there a downside to the current syntax today? If so mind listing that out?

sfackler:
The primary concern is how to fit parameters in. The matcher for the format arguments has to be (I think?) `$($t:tt)*` so it has to be the last thing in the macro, at least at its level of paren nesting. Right now the only other thing we have is the target so it's either `info!("foo: {}", foo)` or `info!(target: "some_target", "foo: {}", foo)`. Once we add parameters though, this seems pretty gross: `info!(target: "some_target", params: { foo: foo, bar: "hello", }, "foo: {}", foo)`.

alexcrichton:
Oh right yes sorry! I think I misunderstood.

dtolnay:
Not proposing this as a solution, but it's perfectly possible to parse format args that are not the end of a level of nesting.

```rust
macro_rules! info {
    // public
    (target: $t:expr, $fmt:expr) => {
        info!([$fmt]);
    };
    (target: $t:expr, $fmt:expr, $($rest:tt)*) => {
        info!([$fmt] $($rest)*);
    };

    // private
    ([$($fmt:tt)*]) => {
        println!($($fmt)*);
    };
    ([$($fmt:tt)*] params: { $($k:ident : $v:expr),* }) => {
        println!($($fmt)*);
        $(
            println!("params[{}] = {:?}", stringify!($k), $v);
        )*
    };
    ([$($fmt:tt)*] params: { $($k:ident : $v:expr,)* }) => {
        info!([$($fmt)*] params: { $($k : $v),* });
    };
    ([$($fmt:tt)*] $n:ident = $v:expr) => {
[...]

PR #220, opened 2017-08-22
Allow `#[macro_use(info)]` to import individual macros.

### LINKED ISSUE #54: #[macro_use(info)] raise error: macro undefined: 'log!'
Code:

``` rust
#[macro_use(info)]
extern crate log;
```

Error:

``` rust
<log macros>:4:1: 4:4 error: macro undefined: 'log!'
<log macros>:4 log ! ( $ crate:: LogLevel:: Info , $ ( $ arg ) * ) ; )
               ^~~
```

I will use `#[macro_use]` for now but I thought it was interesting [to limit](https://doc.rust-lang.org/stable/reference.html#macro-related-attributes) the macro to import in the namespace

``` bash
$rustc --version
rustc 1.1.0-dev
```

mjkillough:
Inline logic from `log!` into `error!`/`warn!`/`...`.

Alternatively, we could create a helper function which all macros call (passing in the output of `file!`/`line!`/etc.), but this helper function would need to be in the public API of `log`, which is a bit ugly.

After playing around with alternatives, I figured I may as well open a PR for this option (which I consider to be the least of all evils). If the duplication is undesirable, then feel free to close! :)

Fixes #54.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,9 @@ publish = false # this branch contains breaking changes
 name = "filters"
 harness = false
 
+[[test]]
+name = "macro_use"
+
 [features]
 max_level_off   = []
 max_level_error = []
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -41,6 +41,9 @@
 macro_rules! log {
     (target: $target:expr, $lvl:expr, $($arg:tt)+) => ({
         let lvl = $lvl;
+        // Warning: This code is duplicated in `log!`/`error!`/`warn!`/... -
+        // see discussion in #54. Once rust-lang/rust#25003 is fixed, this may
+        // no longer be necessary.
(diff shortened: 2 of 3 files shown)


mjkillough:
Looking at the proposal in #218, it may be that the macros will soon become more complex, making the code duplication a lot more painful...

KodrAus:
Thanks for doing this @mjkillough! My personal feeling on this is that we should hold off for now, assuming we want to solve this problem this way, so there's less friction to exploring structured logging.

@sfackler might have other ideas though.

sfackler:
Yeah, I'd be a bit worried about the implementations desynchronizing, especially as the macros become more complicated. Making a script to generate the module is one option.

mjkillough:
Yep, that's fair enough! I'll close this one.

PR #262, opened 2018-02-07
List android_log crate in README

dbrgn:
I'm not affiliated with the crate, but I think it's an interesting addition because it shows that logging can also be done cleanly on platforms like Android, not just on regular Linux.


sfackler:
Thanks!

Issue #148, opened 2017-05-19
Ensure 0.3 works with the next major release


alexcrichton:
We'll want to publish a version of 0.3 that works as a compatibility layer to the 0.4 version of the `log` crate

daboross:
If it's alright and no one else has started, I can start working on a shared backend crate for this.

I'm trying to think of the best way to have it be efficient, while still allowing for maximum change. Right now my best idea is to have a versioned API that would be somewhat like the following:

```rust
pub fn register_v1_log(log: Box<V1Log>) -> Result<(), ...>;
//! this might also want to take `*const V1Log` instead for most flexibility. 

pub fn log_v1(record: V1Record);

trait V1Log {
    fn log(&self, record: V1Record);
}
```

`V1Record` would be a mirror of what the `log` crate's current `LogRecord` struct is, and the `log` crate would simply have a newtype wrapper around it for the exported `LogRecord` type.

`V1Record` would be constructed with a `V1RecordBuilder`: `#[inline(always)]` builder methods should allow for the same performance we have now without a builder, and this would allow us to add more fields without bumping to `V2Record` immediately. 

`V1Log` would have a blanket impl like this in `log`:

```
#[doc(hidden)]
impl<T: Log> V1Log for T {
    fn log(&self, record: V1Record) {
        <Self as Log>>::log(&LogRecord { inner: record });
    }
}
```

For the next version of `log` which brings an incompatible `LogRecord` format, `log_impl` would introduce `V2Log` and `V2Record` to match, and would `impl From<V1Record> for V2Record` and vice versa.

Internally, `log_impl` would store an enum `VersionedLog` which could be a box of any of the `V*Log` traits, and would convert records accordingly when logged.

The `log_impl` would provide one other feature, a set of macros `if_trace_enabled!()` through `if_error_enabled!()`, which would be controlled through `max_level_*` and `release_max_level_*` crate features, similar to `log`'s current ones. These macros would then be used in `log` instead of the features, and the `log` features would simply forward to depending on the `log_impl` ones.

There would be overhead of converting records if the `log` macro version and logger version don't match, but I believe a scheme like this would fully support both public breaking API changes in `log`, and breaking changes in how `LogRecord` is internally structured.

If we don't implement any breaking changes in `LogRecord`s format, we could even just have `log_impl` with `V1*` always, and there would be no overhead.

Thoughts on this?

brson:
[...]

PR #264, opened 2018-02-14
explicitly test against 1.16.0

cardoe:
Explicitly test against 1.16.0 like
fb65deee08b0f1d86102cb3ab49aba96b480d768 to ensure we know if support is
dropped. ref #148

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: rust
 sudo: false
 rust:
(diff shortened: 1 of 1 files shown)


cardoe:
Ah alright. I was trying to put a bow on #148 after it got linked to me on a crate I support with an old version of Rust. Ah you get the picture from all these issues.

PR #268, opened 2018-04-08
Level and LevelFilter can now be deserialized from byte slices

Lukazoid:
This is required by some deserializers which deserialize identifiers using their binary representation. An example of one such serializer is [Ron](https://github.com/ron-rs/ron/blob/a5a4b2ee21f5be0b07c55844e2689ae8a521da3b/src/de/mod.rs#L441-L446).

This implementation should be equivalent to what serde derive would [implement for enum variants](https://github.com/serde-rs/serde/blob/5efb22ebee128f1ec4a755556230bdd66c3f2026/serde_derive/src/de.rs#L2084-L2097).


sfackler:
This seems reasonable to me. Agree @dtolnay?

Issue #197, opened 2017-06-08
crate feature flags - require better documentation


budziq:
While working on https://github.com/rust-lang-nursery/log/issues/125 I've discovered that [log](https://github.com/rust-lang-nursery/log/blob/master/src/lib.rs#L211)'s "crate feature flags" documentation might be insufficient:

- "Features" headline is ambiguous. I find [ndarray](https://docs.rs/ndarray/0.9.1/ndarray/#crate-feature-flags) "crate feature flags" much clearer
- actual feature names are not mentioned in crate level docs (just wildcarded `max_level_*` and `release_max_level_*`)
- `max_level_trace` and `release_max_level_trace` are never mentioned in docs at all and cannot be discovered without inspecting the source code

I would suggest using ndarray's rustdoc as a template.

related to https://github.com/brson/rust-api-guidelines/issues/43

PR #274, opened 2018-05-30
Added Crate Feature Flags Documentation

cowang4:
I had difficulty figuring out why I couldn't get `serde` support, until I figured out that I was missing a feature flag in my `Cargo.toml`. This PR addresses that issue for others by adding a short section to the crate's docs about it.

Please make suggestions and comments, so that we can make rust's documentation even better!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -233,7 +233,7 @@
 //! configured in your `Cargo.toml`.
 //!
 //! * `std` allows use of `std` crate instead of the default `core`. Enables using `std::error` and
-//! set_boxed_logger functionality.
+//! `set_boxed_logger` functionality.
 //! * `serde` enables support for serialization and deserialization of `Level` and `LevelFilter`.
 //!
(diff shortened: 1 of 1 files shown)


cowang4:
Oh yeah, this is inspired by #197.

src/lib.rs
```diff
@@ -227,6 +227,19 @@
 //! [dependencies]
 //! log = { version = "0.4", features = ["max_level_debug", "release_max_level_warn"] }
 //! ```
+//! # Crate Feature Flags
+//!
+//! The following crate feature flags are avaliable in addition to the filters. They are
+//! configured in your `Cargo.toml`.
+//!
+//! * `std` allows use of `std` crate instead of the default `core`. Enables using `std::error` and
+//! set_boxed_logger functionality.
```
KodrAus: nitpick: Let's put _set_boxed_logger_ in backticks here

PR #278, opened 2018-06-09
Further optimize code size

Amanieu:
I had a look at #276 and thought I could do better!

There basically two areas where an improvement can be found:

- LLVM generates pretty poor code for the default implementations of `PartialOrd::{lt,le,gt,ge}` which forward to `partial_cmp`. So I added versions which perform the desired comparison directly.
- The target and location are almost always compile-time constants and can be passed in as a single argument which points to a static constant. This unfortunately doesn't work for the log level since we first need to make a local copy of it so that it is only evaluated once.

A main function containing nothing but `warn!("hello world")` shrinks from 160 bytes to 98 bytes in x86_64 with this change.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1221,10 +1221,7 @@ pub fn logger() -> &'static Log {
 pub fn __private_api_log(
     args: fmt::Arguments,
     level: Level,
-    target: &str,
-    module_path: &str,
-    file: &str,
-    line: u32,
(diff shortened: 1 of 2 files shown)


sfackler:
Thanks!

alexcrichton:
@sfackler mind doing a release with this in it? It may actually be a bit of a perf boost as well for the comparisons!

sfackler:
Done!

PR #280, opened 2018-06-18
Elaborated documentation around log levels.

golddranks:
Two changes:

1) fixed the mistake in the example: `release_max_level_warn` disables `trace`, `debug`, and `info`, not `trace`, `info`, and `warn`.

2) The explanation of log levels was a bit misleading because it was said that `error!` represents the *highest-priority log level*. However, the *log level* that shows only `error` is the *lowest*. I removed the word level there, and added additional explanation about log levels as *the level of verbosity*.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -29,11 +29,8 @@
 //! The basic use of the log crate is through the five logging macros: [`error!`],
 //! [`warn!`], [`info!`], [`debug!`] and [`trace!`]
 //! where `error!` represents the highest-priority log messages
-//! and `trace!` the lowest. The log messages are filtered by setting the log
-//! level which means the level of verbosity: the lowest level doesn't log
-//! at all and the higher ones log increasingly much, starting from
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -28,8 +28,12 @@
 //!
 //! The basic use of the log crate is through the five logging macros: [`error!`],
 //! [`warn!`], [`info!`], [`debug!`] and [`trace!`]
-//! where `error!` represents the highest-priority log level, and `trace!` the lowest.
-//!
+//! where `error!` represents the highest-priority log messages
```
KodrAus: Maybe we could simplify this explanation a bit to something like:

_where `error!` represents the highest-priority log messages and `trace!` the lowest. The log messages are filtered by configuring the log level to exclude messages with a lower priority._

What do you think?

golddranks:
Nice, it captures the intention while being simpler and more concise! I'll update the PR in a sec.

PR #287, opened 2018-08-01
Ensure that initialization is complete after `set_logger_inner`.

### LINKED ISSUE #286: Avoiding potentially lost log messsages when INITIALIZING
`set_logger_inner` is careful to initialize its log state once, even when called from multiple threads:

https://github.com/rust-lang-nursery/log/blob/48bdd5679448e0ec43a20fb0e6a86900dcdefddd/src/lib.rs#L1154

and `logger` uses a `NopLogger` if the state isn't `INITIALIZED`:

https://github.com/rust-lang-nursery/log/blob/48bdd5679448e0ec43a20fb0e6a86900dcdefddd/src/lib.rs#L1211

However, if the thread doing the initialization doesn't run fast enough, `STATE` might still be `INITIALIZING` when `set_logger_inner` returns in another thread, so there is a window in that other thread in which log messages could be lost.

Would it make sense to either:
 - Make `set_logger_inner` wait if `STATE` is `INITIALIZING` until it changes to `INITIALIZED`, so that the state is always `INITIALIZED` by the time it returns, or
 - Add a way for user code to wait until `STATE` is `INITIALIZED`, so that user code can opt to wait if it wishes to avoid dropping messages

?

sunfishcode:
If `set_logger_inner` is called while `STATE` is `INITIALIZING`, make it wait until the `STATE` is `INITIALIZED` so that callers can assume that the logging state is fully initialized after a call to `set_logger_inner`.

Note that this is a very simple implementation which could deadlock if `make_logger()` calls back into `set_logger_inner`, but that shouldn't happen since log's documentation says that `set_logger` should only be called once. If this is a concern, a more complex implementation that avoids this problem would be possible.

Fixes #286.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -292,7 +292,7 @@ use std::error;
 use std::fmt;
 use std::mem;
 use std::str::FromStr;
-use std::sync::atomic::{AtomicUsize, Ordering, ATOMIC_USIZE_INIT};
+use std::sync::atomic::{AtomicUsize, Ordering, ATOMIC_USIZE_INIT, spin_loop_hint};
 
 #[macro_use]
 mod macros;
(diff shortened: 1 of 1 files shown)


sfackler:
spin_loop_hint isn't stable on 1.16.0, so we won't be able to use that unfortunately. We could detect the compiler version at build time and use it when supported, but I'm sure it'd be worth it rather than just spinning.

sfackler:
Thanks! I misinterpreted the issue you filed initially - the reasoning for this change makes sense to me now.

We don't need to worry about the reentrancy issue since the closures passed to set_logger_inner are all controlled by the log crate.

PR #290, opened 2018-08-29
Port set_max_level interface from 0.4.x

breeswish:
0.4.x provides `set_max_level`, so that when logging library cannot set max level (for example, slog), logging library user can manually set it.

However In 0.3.x, `max_level` can be set only in `set_logger` interface with a `MaxLogLevelFilter` token, which forbids logging library user from setting it. This PR simply exports the `set_max_level` in 0.3.x like in 0.4.x, so that users can call it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "log"
-version = "0.3.9"
+version = "0.3.10"
 authors = ["The Rust Project Developers"]
 license = "MIT/Apache-2.0"
 readme = "README.md"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -682,6 +682,14 @@ pub fn max_log_level() -> LogLevelFilter {
     LogLevelFilter::from_new(log::max_level())
 }
 
+/// Sets the global maximum log level.
+///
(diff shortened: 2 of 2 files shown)


sfackler:
If people want to use features from 0.4.x, it seems like they should use 0.4.x. What's preventing you from upgrading?

breeswish:
@sfackler Hi, currently I'm using log + slog + slog-stdlog.

Starting from log 0.4.x, the location's lifetime changed from `&'static str` to `&'a str`, which is [not supported by the current slog](https://github.com/slog-rs/stdlog/pull/5). Unfortunately, supporting this change in slog is [a breaking change](https://github.com/slog-rs/slog/pull/171#discussion_r196636867) as well, which will not be shipped very soon (there are [other features to be carried](https://github.com/slog-rs/slog/wiki/Things-to-make-different-next-time-around-(slog-v3%3F)) in slog to bump to a new major version). Simply speaking, upgrading to use log 0.4.x in my case is not possible currently.

KodrAus:
It's been a few years here and it doesn't look like we're supporting releases to `0.3.x` so I'll close this one for now.

Thanks for doing the backport @breeswish

PR #293, opened 2018-09-21
Fix README example

rivertam:
I think it makes more sense now.

(Sorry if this is inappropriate; this is intended to be lighthearted)

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ pub fn shave_the_yak(yak: &mut Yak) {
         match find_a_razor() {
             Ok(razor) => {
                 info!("Razor located: {}", razor);
-                yak.shave(razor);
+                razor.shave(yak);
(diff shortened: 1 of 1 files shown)


KodrAus:
🤔 The razor is really a component of the shave and on its own it can't shave any yaks. The yak really needs to shave with the razor. Maybe we could do something like:

```rust
yak.shave_with(razor);
```

It's hard to know what the best way to shave yaks is without an RFC.

dtolnay:
I think `yak.shave(razor)` is sufficiently clear for the purpose of example code and I prefer it over `razor.shave(&mut yak)`, but I would accept a PR to say `yak.shave_with(razor)` as Ashley suggested.

Closing for now because this PR has been inactive for a while. Thanks anyway!

PR #298, opened 2018-10-26
Fixed Readme containing confusing instructions for Rust 2018

Gilgamesh22:
the new Rust 2018 notation does not remove the need for extern crate. However the example shown does not represent that.


KodrAus:
It's been a little while now so I'll go ahead and close this one. The current example is correct for `edition = 2018` crates depending on `log`, but I'll try clarify the text a little more in the readme.

Thanks again @Gilgamesh22

PR #299, opened 2018-10-26
exclude CI files from crates.io

ignatenkobrain:


Cargo.toml
```diff
@@ -13,6 +13,7 @@ A lightweight logging facade for Rust
 """
 categories = ["development-tools::debugging"]
 keywords = ["logging"]
+exclude = ["/.travis.yml", "appveyor.yml"]
```
dtolnay: I would expect either both or neither of these to have a leading `/`.

ignatenkobrain:
I missed it for second. Could you fix it, please?

Cargo.toml
dtolnay:
```suggestion
exclude = ["/.travis.yml", "/appveyor.yml"]
```

PR #302, opened 2018-10-27
Fix log_enabled with new-style imports.

sfackler:
Fixes #301 

r? @dtolnay


sfackler:
Beaten to the punch by #301!

PR #313, opened 2019-01-03
Add minimum version policy to readme

KodrAus:
Closes rust-lang-nursery#252

Just calls out the explicit version of rustc that we test against in CI, along with a policy for how we'll change it. I copied it from lazy-static.

cc @sfackler are you on board with this? Or would you prefer something different?

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ implementation that is most suitable for its use case.
 
 `1.16.0+`
 
-This version is explicitly tested in CI and may only be bumped in new minor versions. Any changes to the supported minimum version will be called out in the release notes.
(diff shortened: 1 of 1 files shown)


sfackler:
We haven't technically lived by this historically - the newest (facade over 0.4) release of 0.3 bumped the minimum version from 1.15 to 1.16 for example. In that instance, the benefit of not losing log messages from crates using old versions easily outweighs moving the minimum supported version IMO.

KodrAus:
Ah right, that does make sense.

How does:

> we'll bump it as needed but call out changes in release notes.

sound then?

sfackler:
That seems right, but we should emphasize that the bar for bumping is quite high, and we'll want to keep it as a pretty old version regardless.

KodrAus:
Good call, I've updated the wording. If you're happy with this then I'll squash it down before merging.

PR #317, opened 2019-01-30
An RFC for inline expression and value logging

dekellum:
Add support to `log` for inline expression and value logging, as a superset of the `dbg!` macro recently added to rust `std`.

[Rendered](https://github.com/dekellum/log/blob/rfc-v-macros/rfcs/0001-inline-expression-value-logging.md)

Implementation PR: #316

### DIFF
--- a/rfcs/0001-inline-expression-value-logging.md
+++ b/rfcs/0001-inline-expression-value-logging.md
@@ -178,7 +178,7 @@ macro(s), at least in 2018 edition projects, adding this feature to
   _-v_ macros will likely need to be adapted, when structured logging
   is implemented ([log RFC 296]).
 
-The [implementation PR] (as of this writing) adds just 74 lines
+The [implementation PR] (as of this writing) adds just 65 lines
 of non-test code, all of which is `macro_rules!`.
 
 # Unresolved Questions
--- a/rfcs/0001-inline-expression-value-logging.md
+++ b/rfcs/0001-inline-expression-value-logging.md
@@ -96,14 +96,38 @@ documentation, in the form of rustdoc with doc-tests.
 
 ## Multiple expression support
 
-[RFC 2173] included multiple-expression support for `std::dbg!` but
-was closed, at least in part due to perceived complexity with multiple
-expressions and tuple value returns, in preference to [RFC 2361] which
-was ultimately merged and implemented. While `log` is arguably a "more
-advanced" tool, the complexity of multiple expression support does not
-seem to be warranted by the functional gain or convenience win, over
-using the existing formatted logging macros for more complex
-formatting requirements.
+[RFC 2173] included multiple expression printing and return of values
+via tuple for `std::dbg!`, but was closed in preference to [RFC 2361] as
+merged and implemented.  RFC 2361 on this particular [design
+aspect][2361-single]:
+
+> If the macro accepts more than one expression (returning a tuple),
+> there is a question of what to do with a single
+> expression. Returning a one-value tuple `($expr,)` is probably
+> unexpected, but _not_ doing so creates a discontinuity in the macro's
+> behavior as things are added. With only one expression accepted,
+> users can still pass a tuple expression or call the macro multiple
+> times.
+
+In relation to the proposed design of this RFC, accepting multiple
+expressions would also be at odds with allowing an optional custom
+format string as a preceding parameter.  To support both would require
+an additional marker, e.g. `debugv!(format: "{} = {:x}", i, j)`, as
+further complication and bulk.
+
+As suggested in RFC 2361, explicitly passing a multiple expression
+tuple works when desired, and avoids complications to both the
+syntax and macro implementation:
+
+```rust
+let j = 19;
+let (q, r) = debugv!((j/4, j%4));
+\\           ^-- debug log message: (j / 4, j % 4) = (4, 3)
+let (q, r) = debugv!("quarter {} = {:?}", (j/4, j%4));
+\\           ^-- debug log message: quarter (j / 4, j % 4) = (4, 3)
+assert_eq!(q, 4);
+assert_eq!(r, 3);
+```
 
 ## DSL extension of existing log macros
 
@@ -189,3 +213,4 @@ None.
 [RFC 2361]: https://github.com/rust-lang/rfcs/pull/2361
 [log RFC 296]: https://github.com/rust-lang-nursery/log/pull/296
 [implementation PR]: https://github.com/rust-lang-nursery/log/pull/316
+[2361-single]: https://github.com/rust-lang/rfcs/blob/master/text/2361-dbg-macro.md#accepting-a-single-expression-instead-of-many
--- a/rfcs/0001-inline-expression-value-logging.md
+++ b/rfcs/0001-inline-expression-value-logging.md
@@ -42,32 +42,52 @@ equivalent.
 
 ## `std::dbg!` doesn't work well in projects with `log` configured
 
-In the context of unit tests, when using `cargo test`, _stderr_ (or
-_stdout_) configured log output isn't captured, while `dbg!`, at least
-when used within the tests, is captured.  For passing tests, only log
-output will be shown. For failing tests, log and `dbg!` output are
-shown out of order, with `dbg!` captured and buffered for output at
-the end of the test.
+In the context of unit tests, the `cargo test` harness attempts to
+capture _stderr_/_stdout_ independently for each test, but the
+mechanism is incomplete and fragile: Output from threads other than
+the test thread _escapes_ the capture, as does the output of most all
+logger implementations, even when on the test thread. Interest in
+fixing this has most recently been subsumed by
+[_Tracking issue for eRFC 2318, Custom test frameworks_, rust-lang/rust#50297 (comment)][issue 50297]—but
+with no clear commitment to a complete implementation and fix.
 
-A workaround is the rather elusive `cargo test -- --nocapture` flags,
-while selecting a single test by name or using `-j 1` (single thread)
-to prevent interleaving of output. But even with these workarounds in
-place, or in the context of examples or other executables:
+This results in the practical issue that mixing `log` output with
+`println!`, `eprintln!`, or `std::dbg!` results in confusing
+buffering, with only the latter being captured. When tests pass, only `log`
+output will be shown. When tests fail, `log` and `std::dbg!` output are
+both shown, but out of order, with `std::dbg!` captured and buffered
+for output after the test panics.
+
+For new users of `log` this can be particularly confusing, since they
+are often working through their own project bugs while at the same
+time trying to understand this inconsistent `cargo test` behavior.
+
+In such mixed usage, no capture is must less confusing than
+partial/broken output capture, so the rather elusive and fun to type
+`cargo test -- --nocapture --test-threads=1` can suffice as a
+workaround. Anecdotally: The author finally discovered the last flag
+of this puzzle just rencently, and previously was incorrectly using the
+_documented_ `-j 1` flag.
+
+Even with the hard-earned knowledge of these workarounds, or if test
+output capture eventually becomes reliable, when mixing `std::dbg!` with
+`log`:
 
 * The output is jarringly in two different formats: one configurable
   by the logging implementation, the other hard-coded by `std::dbg!`
-  which acts effectively like its own micro-logging system.
+  which is effectively its own micro-logging system.
 
 * `std::dbg!` is hard-coded to use "pretty" multi-line format (via
   `{:#?}`), which is also jarring by normal logging conventions.
 
-* `std::dbg!` isn't beholden to log's level or other filtering and
-  thus can't be kept in a project.
+* `std::dbg!` isn't beholden to `log`'s level or other filtering and
+  thus can't be reasonably kept in a project, at least outside of test
+  code.
 
 * Other log metadata or output options like module name (`target`) or
   thread names are not available with `dbg!`
 
-Below, us an excerpt of a debugging session combining `log::trace!`,
+Below is an excerpt of a debugging session combining `log::trace!`,
 `log::debug!`, and `std::dbg!`:
 
 ```txt
@@ -78,8 +98,7 @@ TRACE tokio_threadpool::sender: execute; count=1
 TRACE tokio_threadpool::pool:     -> submit external
 TRACE tokio_threadpool::pool: signal_work -- notify; idx=1
 TRACE tokio_threadpool::pool: signal_work -- spawn; idx=1
-[body-image-futio/src/futio_tests/server.rs:73] Tuner::new().set_buffer_size_fs(17).finish()
-= Tunables {
+[body-image-futio/src/futio_tests/server.rs:73] Tuner::new().set_buffer_size_fs(17).finish() = Tunables {
     max_body_ram: 196608,
     max_body: 1073741824,
     buffer_size_ram: 8192,
@@ -102,9 +121,9 @@ TRACE tokio_threadpool::worker:   sleeping -- push to stack; idx=1
 TRACE tokio_threadpool::worker:     -> starting to sleep; idx=1
 ```
 
-With `std::dbg!` released, there is intrinsic value in offering
-developers parity and convenience with an extension to `log` for
-inline expression and value logging.
+With `std::dbg!` released, there is now intrinsic value in at least
+offering developers parity and convenience with an extension to `log`
+for inline expression and value logging.
 
 # Detailed Design
 
@@ -303,3 +322,4 @@ None.
 [log RFC 296]: https://github.com/rust-lang-nursery/log/pull/296
 [implementation PR]: https://github.com/rust-lang-nursery/log/pull/316
 [2361-single]: https://github.com/rust-lang/rfcs/blob/master/text/2361-dbg-macro.md#accepting-a-single-expression-instead-of-many
+[issue 50297]: https://github.com/rust-lang/rust/issues/50297#issuecomment-388988381
--- a/rfcs/0001-inline-expression-value-logging.md
+++ b/rfcs/0001-inline-expression-value-logging.md
@@ -1,0 +1,191 @@
+# Summary
+
+Add support to `log` for inline expression and value logging, as a
+superset of the `dbg!` macro recently added to rust `std`.
+
+# Motivation
+
+The motivation is in part the same as for the accepted `dbg!` macro of
+[RFC 2361] and implemented in rust 1.32. To summarize:
+
+* Its convenient to be able to insert logging into larger expressions
+  without needing to restructure using additional `let` bindings, or
+  by duplicating sub-expressions.
+
+* For debug/trace level logging in particular, automatic formatting of
+  the expression with its value can give sufficient context in the log
+  (particularly with _target_, _file_ and _line_ information) and avoids
+  additional boilerplate.
+
+In a project where configuring `log` and an output logger
+implementation hasn't (yet) happened, one can conveniently use
+[`std::dbg!`](https://doc.rust-lang.org/std/macro.dbg.html), with no
+initial setup cost, for `expression = value` printing to STDERR. From
+the linked rustdoc:
+
+> Note that the macro is intended as a debugging tool and therefore
+> you should avoid having uses of it in version control for longer
+> periods. Use cases involving debug output that should be added to
+> version control may be better served by macros such as `debug!` from
+> the `log` crate.
+
+Indeed, a major point of the `log` package and `Level`s, is the
+ability to keep `debug!` and `trace!` logging in place for further
+use, including by other contributors, without paying a cost for
+unlogged messages in release builds.
+
+It follows that for projects that _do already have_ `log` and an
+output logger dependency and configuration, and particularly with some
+debug/trace logging already in place, use of `std::dbg!` would be
+unwelcome in PRs, and likely less productive than a `log`-based
+equivalent. For parity of convenience, the `log` API needs an
+extension for logging expressions and passing through values inline.
+
+# Detailed Design
+
+In addition to the existing set of _formatted logging_ macros, by
+level, e.g. `trace!`, `debug!`, etc., we add a new set of _inline
+expression and value logging_ macros, with a “_-v_” suffix: `tracev!`,
+`debugv!`, and so forth for all levels. The _-v_ macros take a _single_
+expression argument, which is evaluated exactly once, regardless of if
+the logging level is enabled or not, and returned:
+
+```rust
+let n = 12;
+let m = debugv!(n / 2) - 1;
+//      ^-- debug log message: "n / 2 = 6"
+assert_eq!(m, 5);
+```
+
+The _default_ format string for the _-v_ macros is `"{} = {:?}"`,
+where the `stringify!`-ed expression and resulting value are passed,
+in that order.  Note that the `std::dbg!` macro currently uses `"{} =
+{:#?}"`—the value is "pretty-printed", potentially over multiple
+lines.  Given the line-orientation of logging output, the default
+format for the _-v_ macros avoids this. However, a custom format may
+also be passed before the expression, which adds considerably more
+output flexibility:
+
+```rust
+let i = 32;
+tracev!("{} = {}", i);            // use `Display` instead of `Debug`
+tracev!("{} = {:x}", i);          // hexadecimal format value
+tracev!("{} = {:#?}", i);         // use pretty, multi-line format
+
+let rem = infov!("{1:?} remaining ({0})", deadline - Instance::now());
+                                  // value first, with additional context
+```
+
+For symmetry with the existing `log!` macro, a `logv!` is also
+included, which allows passing a `Level` as parameter.
+
+Finally, like all the other public logging macros, the _-v_ macros
+allows overriding the default module-path target with a string
+literal:
+
+```rust
+let i = 33;
+let j = warnv!(target: "maths", (i-1) / 2);
+assert_eq!(j, 16);
+```
+
+See also the [implementation PR], which includes guide level
+documentation, in the form of rustdoc with doc-tests.
+
+# Alternatives
+
+## Multiple expression support
+
+[RFC 2173] included multiple-expression support for `std::dbg!` but
+was closed, at least in part due to perceived complexity with multiple
+expressions and tuple value returns, in preference to [RFC 2361] which
+was ultimately merged and implemented. While `log` is arguably a "more
+advanced" tool, the complexity of multiple expression support does not
+seem to be warranted by the functional gain or convenience win, over
+using the existing formatted logging macros for more complex
+formatting requirements.
+
+## DSL extension of existing log macros
+
+Instead of adding a new set of _-v_ macros, it would be possible to
+extend the existing logging macros by using some additional marker
+syntax, such as the following:
+
+```rust
+debug!(= n/2)
+trace!("{} = {:x}", =i);
+```
+
+Here the `=` signals that the expression should be `stringify!`-ed for
+the message and its value returned from the macro. This complicates
+the macro's, but more importantly, considerably complicates the
+necessary guide documentation for new and existing users to understand
+an evolving logging _DSL_ as new syntax.  This syntax isn't any more
+compact. It aids comprehension when the macro arguments are as
+function-like as possible, by introducing the _-v_ macros specific to
+the feature as designed.
+
+## Only customize the value part of the format
+
+As proposed above, the entire format string may be customized. Until
+understanding that the single `i` argument is expanded to two
+arguments for formatting, it is surprising to see the following, as
+used above, with two placeholders in the format string:
+
+``` rust
+tracev!("{} = {:x}", i);
+```
+
+A workable alternative would be to only allow customizing the value
+part of the format, and have the expression part be fixed as `"{} =
+"`:
+
+``` rust
+tracev!("{:x}". i);
+```
+
+While this alternative is more compact for the subset of compatible
+use cases, it offers _less_ customization options.  For example,
+supplying a custom literal prefix to the message, for additional
+context. It is also less symmetric with the the _formatted logging_
+macros, where the format string literal represents the _entire_ text
+message of the log `Record`. Finally, this alternatives adds a small
+but measurable amount of formatting overhead, as the value needs to be
+formatted to a temporary `fmt::Arguments` before formatting to the
+final message.
+
+## Release this as a separate crate, not in `log`
+
+The proposed additional logging macros play the same role as, for
+example, the existing `debug!` macro in `log`, which is just more
+convenient than using `log!(Level::Debug, …)` with an extra import.
+
+Ease of use was also an important part of the design and decision to
+add `dbg!` to rust `std` and the prelude.
+
+While the proposed additions would still require import for the
+macro(s), at least in 2018 edition projects, adding this feature to
+`log` avoids:
+
+* Needing to _discover_, add, and maintain an additional library
+  dependency. Discovery may be the biggest issue, and the inclusion of
+  `dbg!` in `std` and the prelude raises a high bar. This could be
+  partially mitigated by documented such an extension crate in the log
+  README and/or top-level library rustdoc.
+
+* The community effort to maintain such a separate library with
+  compatibility to the `log` crate, as it evolves.  Fro example, _-v_
+  macros will likely need to be adapted, for example, when structured
+  logging is implemented ([log RFC 296]).
+
+The [implementation PR] (as of this writing) adds just 74 lines
+non-test code, only of `macro_rules!`.
+
+# Unresolved Questions
+
+None.
+
+[RFC 2173]: https://github.com/rust-lang/rfcs/pull/2173
+[RFC 2361]: https://github.com/rust-lang/rfcs/pull/2361
+[log RFC 296]: https://github.com/rust-lang-nursery/log/pull/296
+[implementation PR]: https://github.com/rust-lang-nursery/log/pull/316
--- a/rfcs/0001-inline-expression-value-logging.md
+++ b/rfcs/0001-inline-expression-value-logging.md
@@ -149,7 +149,7 @@ use cases, it offers _less_ customization options.  For example,
 supplying a custom literal prefix to the message, for additional
 context. It is also less symmetric with the the _formatted logging_
 macros, where the format string literal represents the _entire_ text
-message of the log `Record`. Finally, this alternatives adds a small
+message of the log `Record`. Finally, this alternative adds a small
 but measurable amount of formatting overhead, as the value needs to be
 formatted to a temporary `fmt::Arguments` before formatting to the
 final message.
@@ -174,12 +174,12 @@ macro(s), at least in 2018 edition projects, adding this feature to
   README and/or top-level library rustdoc.
 
 * The community effort to maintain such a separate library with
-  compatibility to the `log` crate, as it evolves.  Fro example, _-v_
-  macros will likely need to be adapted, for example, when structured
-  logging is implemented ([log RFC 296]).
+  compatibility to the `log` crate, as it evolves.  For example, the
+  _-v_ macros will likely need to be adapted, when structured logging
+  is implemented ([log RFC 296]).
 
 The [implementation PR] (as of this writing) adds just 74 lines
-non-test code, only of `macro_rules!`.
+of non-test code, all of which is `macro_rules!`.
 
 # Unresolved Questions
 
--- a/rfcs/0001-inline-expression-value-logging.md
+++ b/rfcs/0001-inline-expression-value-logging.md
@@ -172,10 +172,7 @@ use cases, it offers less customization options.  For example, it does
 not support adding a custom literal prefix to the message, for additional
 context. It is also less symmetric with the the _formatted logging_
 macros, where the format string literal represents the _entire_ text
-message of the log `Record`. Finally, this alternative adds a small
-but measurable amount of formatting overhead, as the value needs to be
-formatted to a temporary `fmt::Arguments` before formatting to the
-final message.
+message of the log `Record`.
 
 ## Release this as a separate crate, not in `log`
 
--- a/rfcs/0001-inline-expression-value-logging.md
+++ b/rfcs/0001-inline-expression-value-logging.md
@@ -6,9 +6,9 @@ superset of the `dbg!` macro recently added to rust `std`.
 # Motivation
 
 The motivation is in part the same as for the accepted `dbg!` macro of
-[RFC 2361] and implemented in rust 1.32. To summarize:
+[RFC 2361], as implemented in rust 1.32. To summarize:
 
-* Its convenient to be able to insert logging into larger expressions
+* It is convenient to be able to insert logging into larger expressions
   without needing to restructure using additional `let` bindings, or
   by duplicating sub-expressions.
 
@@ -77,7 +77,7 @@ let rem = infov!("{1:?} remaining ({0})", deadline - Instance::now());
 ```
 
 For symmetry with the existing `log!` macro, a `logv!` is also
-included, which allows passing a `Level` as parameter.
+included, which allows passing the `Level` as a parameter.
 
 Finally, like all the other public logging macros, the _-v_ macros
 allows overriding the default module-path target with a string
@@ -145,9 +145,9 @@ the message and its value returned from the macro. This complicates
 the macro's, but more importantly, considerably complicates the
 necessary guide documentation for new and existing users to understand
 an evolving logging _DSL_ as new syntax.  This syntax isn't any more
-compact. It aids comprehension when the macro arguments are as
-function-like as possible, by introducing the _-v_ macros specific to
-the feature as designed.
+compact. Comprehension is aided when the macro arguments are as
+function-like as possible, with macros specific to the feature, as
+designed above.
 
 ## Only customize the value part of the format
 
@@ -161,16 +161,15 @@ tracev!("{} = {:x}", i);
 ```
 
 A workable alternative would be to only allow customizing the value
-part of the format, and have the expression part be fixed as `"{} =
-"`:
+part of the format, and with the expression part fixed as `"{} ="`:
 
 ``` rust
-tracev!("{:x}". i);
+tracev!("{:x}", i);
 ```
 
 While this alternative is more compact for the subset of compatible
-use cases, it offers _less_ customization options.  For example,
-supplying a custom literal prefix to the message, for additional
+use cases, it offers less customization options.  For example, it does
+not support adding a custom literal prefix to the message, for additional
 context. It is also less symmetric with the the _formatted logging_
 macros, where the format string literal represents the _entire_ text
 message of the log `Record`. Finally, this alternative adds a small
--- a/rfcs/0001-inline-expression-value-logging.md
+++ b/rfcs/0001-inline-expression-value-logging.md
@@ -38,8 +38,73 @@ It follows that for projects that _do already have_ `log` and an
 output logger dependency and configuration, and particularly with some
 debug/trace logging already in place, use of `std::dbg!` would be
 unwelcome in PRs, and likely less productive than a `log`-based
-equivalent. For parity of convenience, the `log` API needs an
-extension for logging expressions and passing through values inline.
+equivalent.
+
+## `std::dbg!` doesn't work well in projects with `log` configured
+
+In the context of unit tests, when using `cargo test`, _stderr_ (or
+_stdout_) configured log output isn't captured, while `dbg!`, at least
+when used within the tests, is captured.  For passing tests, only log
+output will be shown. For failing tests, log and `dbg!` output are
+shown out of order, with `dbg!` captured and buffered for output at
+the end of the test.
+
+A workaround is the rather elusive `cargo test -- --nocapture` flags,
+while selecting a single test by name or using `-j 1` (single thread)
+to prevent interleaving of output. But even with these workarounds in
+place, or in the context of examples or other executables:
+
+* The output is jarringly in two different formats: one configurable
+  by the logging implementation, the other hard-coded by `std::dbg!`
+  which acts effectively like its own micro-logging system.
+
+* `std::dbg!` is hard-coded to use "pretty" multi-line format (via
+  `{:#?}`), which is also jarring by normal logging conventions.
+
+* `std::dbg!` isn't beholden to log's level or other filtering and
+  thus can't be kept in a project.
+
+* Other log metadata or output options like module name (`target`) or
+  thread names are not available with `dbg!`
+
+Below, us an excerpt of a debugging session combining `log::trace!`,
+`log::debug!`, and `std::dbg!`:
+
+```txt
+TRACE mio::poll: registering with poller
+TRACE tokio_threadpool::builder: build; num-workers=2
+TRACE mio::poll: registering with poller
+TRACE tokio_threadpool::sender: execute; count=1
+TRACE tokio_threadpool::pool:     -> submit external
+TRACE tokio_threadpool::pool: signal_work -- notify; idx=1
+TRACE tokio_threadpool::pool: signal_work -- spawn; idx=1
+[body-image-futio/src/futio_tests/server.rs:73] Tuner::new().set_buffer_size_fs(17).finish()
+= Tunables {
+    max_body_ram: 196608,
+    max_body: 1073741824,
+    buffer_size_ram: 8192,
+    buffer_size_fs: 17,
+    size_estimate_deflate: 4,
+    size_estimate_gzip: 5,
+    size_estimate_brotli: 6,
+    temp_dir: "/tmp",
+    res_timeout: None,
+    body_timeout: Some(
+        60s
+    )
+}
+TRACE tokio_threadpool::sender: execute; count=2
+TRACE tokio_threadpool::pool:     -> submit external
+TRACE tokio_threadpool::pool: signal_work -- notify; idx=0
+TRACE tokio_threadpool::pool: signal_work -- spawn; idx=0
+TRACE tokio_threadpool::worker: Worker::sleep; worker=WorkerId(1)
+TRACE tokio_threadpool::worker:   sleeping -- push to stack; idx=1
+TRACE tokio_threadpool::worker:     -> starting to sleep; idx=1
+```
+
+With `std::dbg!` released, there is intrinsic value in offering
+developers parity and convenience with an extension to `log` for
+inline expression and value logging.
 
 # Detailed Design
 
@@ -51,29 +116,38 @@ expression argument, which is evaluated exactly once, regardless of if
 the logging level is enabled or not, and returned:
 
 ```rust
-let n = 12;
-let m = debugv!(n / 2) - 1;
-//      ^-- debug log message: "n / 2 = 6"
-assert_eq!(m, 5);
+use std::time::{Duration, Instant};
+
+let remaining = debugv!(deadline - Instant::now());
+//               ^-- debug log: deadline - Instant::now() → 950µs
+debugv!(remaining);
+// or            ^-- debug log: remaining → 950µs
 ```
 
-The _default_ format string for the _-v_ macros is `"{} = {:?}"`,
+The _default_ format string for the _-v_ macros is `"{} → {:?}"`,
 where the `stringify!`-ed expression and resulting value are passed,
 in that order.  Note that the `std::dbg!` macro currently uses `"{} =
 {:#?}"`—the value is "pretty-printed", potentially over multiple
 lines.  Given the line-orientation of logging output, the default
-format for the _-v_ macros avoids this. However, a custom format may
-also be passed before the expression, which adds considerably more
-output flexibility:
+format for the _-v_ macros avoids this.  Also we use U+2192 RIGHTWARDS
+ARROW (→) as a format separator, which is more easily distinguished
+from any commonly typed log message or rust expression. The log
+record can be customized via two optional parameters: a message prefix
+string, and a format specifier for the value. Note that the former is
+required, if passing the later:
 
 ```rust
 let i = 32;
-tracev!("{} = {}", i);            // use `Display` instead of `Debug`
-tracev!("{} = {:x}", i);          // hexadecimal format value
-tracev!("{} = {:#?}", i);         // use pretty, multi-line format
 
-let rem = infov!("{1:?} remaining ({0})", deadline - Instance::now());
-                                  // value first, with additional context
+infov!(i);
+infov!("", "{:?}", i);       // equivalent to above
+// ^------------------------ info log: i → 32
+infov!("index", i);          // contextual prefix specified
+infov!("index", "{}", i);    // use `Display` instead of `Debug`
+// ^------------------------ info log: index i → 32
+infov!("index", "{:#x}", i); // hexadecimal format value
+// ^------------------------ info log: index i → 0x20
+infov!("index", "{:#?}", i); // use pretty, multi-line format
 ```
 
 For symmetry with the existing `log!` macro, a `logv!` is also
@@ -91,6 +165,35 @@ assert_eq!(j, 16);
 
 See also the [implementation PR], which includes guide level
 documentation, in the form of rustdoc with doc-tests.
+
+# Expected Usage
+
+Like with `dbg!`, its appropriate to add, then shortly remove, uses of
+the _-v_ macros, or to iteratively replace _-v_ macros with the non-v
+macro's (low edit distance) for nicer formatting or more English
+context. The value of the feature does not hinge on the _-v_ macros
+being long lived in code.  The fact that _-v_ macros _could_ be long
+lived is just a bonus of inclusion in `log`.  Its also perfectly
+appropriate to use the _-v_ macros in statement position (including as
+per above design examples).
+
+With the feature in place, while developing and debugging code:
+
+1. Add `tracev!`, `debugv!`, and occasionally, `infov!` calls as
+   convenient for debugging and demonstrating correct behavior.
+
+2. When getting closer to release grade changes, refine your logging
+   by removing some _-v_ macro calls, and replacing some
+   with the existing message formatting macros in statement position,
+   making the messages more like English sentences.
+
+3. Check-in (`git commit`) code with `tracev!`, `debugv!` macro
+   calls in place. If those calls survived to this step, then they are
+   potentially useful in the future to you and other developers, just
+   like the current use of `trace` and `debug`.
+
+4. Iteratively repeat with step (1), possibly in parallel with other
+   developers.
 
 # Alternatives
 
@@ -112,8 +215,9 @@ aspect][2361-single]:
 In relation to the proposed design of this RFC, accepting multiple
 expressions would also be at odds with allowing an optional custom
 format string as a preceding parameter.  To support both would require
-an additional marker, e.g. `debugv!(format: "{} = {:x}", i, j)`, as
-further complication and bulk.
+an additional markers, e.g.
+`debugv!(prefix: "context", format: "{:x}", i, j)`, for further
+complication and bulk.
 
 As suggested in RFC 2361, explicitly passing a multiple expression
 tuple works when desired, and avoids complications to both the
@@ -122,9 +226,9 @@ syntax and macro implementation:
 ```rust
 let j = 19;
 let (q, r) = debugv!((j/4, j%4));
-\\           ^-- debug log message: (j / 4, j % 4) = (4, 3)
-let (q, r) = debugv!("quarter {} = {:?}", (j/4, j%4));
-\\           ^-- debug log message: quarter (j / 4, j % 4) = (4, 3)
+\\           ^-- debug log message: (j / 4, j % 4) → (4, 3)
+let (q, r) = debugv!("quarter", (j/4, j%4));
+\\           ^-- debug log message: quarter (j / 4, j % 4) → (4, 3)
 assert_eq!(q, 4);
 assert_eq!(r, 3);
 ```
@@ -137,7 +241,7 @@ syntax, such as the following:
 
 ```rust
 debug!(= n/2)
-trace!("{} = {:x}", =i);
+trace!("index", =i);
 ```
 
 Here the `=` signals that the expression should be `stringify!`-ed for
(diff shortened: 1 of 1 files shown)


KodrAus:
Thanks for putting this together @dekellum! I'll have a proper read shortly, but I've posted a link [to the subreddit](https://www.reddit.com/r/rust/comments/alhder/an_rfc_for_dbg_style_macros_in_log/) that points back here, just in case we get any comments there.

rfcs/0001-inline-expression-value-logging.md
```diff
@@ -0,0 +1,191 @@
...
+use cases, it offers _less_ customization options.  For example,
+supplying a custom literal prefix to the message, for additional
+context. It is also less symmetric with the the _formatted logging_
+macros, where the format string literal represents the _entire_ text
+message of the log `Record`. Finally, this alternative adds a small
+but measurable amount of formatting overhead, as the value needs to be
+formatted to a temporary `fmt::Arguments` before formatting to the
+final message.
+
+## Release this as a separate crate, not in `log`
```
KodrAus: This is the big question for me. Are users of the `log` crate going to get enough value out of `logv` to warrant building and supporting a parallel API?

Your point about `debug`, `info`, `warn` etc already being effectively parallel APIs is reasonable though.

rfcs/0001-inline-expression-value-logging.md
```diff
@@ -0,0 +1,191 @@
...
+also be passed before the expression, which adds considerably more
+output flexibility:
+
+```rust
+let i = 32;
+tracev!("{} = {}", i);            // use `Display` instead of `Debug`
+tracev!("{} = {:x}", i);          // hexadecimal format value
+tracev!("{} = {:#?}", i);         // use pretty, multi-line format
+
+let rem = infov!("{1:?} remaining ({0})", deadline - Instance::now());
```
KodrAus: I quite like how this has turned out :+1:

rfcs/0001-inline-expression-value-logging.md
```diff
@@ -0,0 +1,191 @@
...
+* Needing to _discover_, add, and maintain an additional library
+  dependency. Discovery may be the biggest issue, and the inclusion of
+  `dbg!` in `std` and the prelude raises a high bar. This could be
+  partially mitigated by documented such an extension crate in the log
+  README and/or top-level library rustdoc.
+
+* The community effort to maintain such a separate library with
+  compatibility to the `log` crate, as it evolves.  For example, the
+  _-v_ macros will likely need to be adapted, when structured logging
+  is implemented ([log RFC 296]).
```
KodrAus: How this interacts with structured logging will be interesting. I think we should be able to decorate a `Record` with at least the stringified expression as a structured value, and use the `Debug` representation of the value if the user hasn't supplied an explicit format function to `logv`.

rfcs/0001-inline-expression-value-logging.md
```diff
@@ -0,0 +1,191 @@
...
+
+See also the [implementation PR], which includes guide level
+documentation, in the form of rustdoc with doc-tests.
+
+# Alternatives
+
+## Multiple expression support
+
+[RFC 2173] included multiple-expression support for `std::dbg!` but
+was closed, at least in part due to perceived complexity with multiple
```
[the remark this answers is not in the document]
dekellum: Thanks @Centril. I will refine this alternative subsection to (a) presume less about why RFC 2173 was superseded; (b) point out that _both_ multiple arguments _and_ a  custom format string are at least somewhat at odds, and (c) note that the following still works, as currently designed, by explicitly passing a tuple:

``` rust
let i = 16;
assert_eq!(debugv!((i/4, i/8)), (4, 2));
assert_eq!(debugv!("{} = {:?}", (i/4, i/8)), (4, 2));
```

Regarding your multiple argument example macro: It would be surprising and undesirable in a _log_ context, I think, to write more than one log record for a single _-v_ macro invocation. To "lock stderr only once" sounds like a closely related goal.

rfcs/0001-inline-expression-value-logging.md
dekellum: This is, in addition to allowing the same sort of structured key+values, at the end of the _-v_ macro arguments,?  Here *I think* you are providing evidence in favor of including this feature in _log_ itself, so these features can be more closely integrated as appropriate.

rfcs/0001-inline-expression-value-logging.md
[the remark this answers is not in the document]
Centril: Thanks @dekellum; I agree with you about the distinction between logging and debugging wrt. the usefulness of multiple arguments; `dbg!(a, b, c)` seems more useful than `debugv!(a, b, c)`. Fwiw, I think this RFC is a nice addition. :)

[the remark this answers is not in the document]
dekellum: Updated this alternatives subsection, per above, and with further elaboration.

rfcs/0001-inline-expression-value-logging.md
KodrAus: Yeh, I was running through the exercise of whether we'd be worse off for adding this feature and then having to introduce structured logging without breakage to it later, which I think we'll be able to do just fine. So no issues there.

rfcs/0001-inline-expression-value-logging.md
```diff
@@ -0,0 +1,191 @@
...
+macro(s), at least in 2018 edition projects, adding this feature to
+`log` avoids:
+
+* Needing to _discover_, add, and maintain an additional library
+  dependency. Discovery may be the biggest issue, and the inclusion of
+  `dbg!` in `std` and the prelude raises a high bar. This could be
+  partially mitigated by documented such an extension crate in the log
+  README and/or top-level library rustdoc.
+
+* The community effort to maintain such a separate library with
```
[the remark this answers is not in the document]
KodrAus: I think the issue is more tied to the amount of context than to the amount of code to support. It just means it's another feature to keep in mind when evolving the library.

I find your argument about the duplicated macros we already have for `info`, `warn` etc pretty convincing though. It's not like we don't already have a suite of similar macros.

KodrAus:
cc @rust-lang-nursery/libs (apologies for the ping) does anyone have any thoughts on adding `dbg!`-like macros in the `log` crate?

As a rough FCP, I'd be on-board with merging and supporting this.

KodrAus:
Alrighty, let's try a bit of a _community final comment period_. I'm on board with this feature as proposed. If you'd also like it, please drop a :+1: on this or the original comment. If you have any reservations, please leave a comment here outlining your concerns.

If we haven't registered any concerns by February 15 then I'll merge this one in and we'll ship it in `log`!

ssokolow:
I like what I see in the RFC but I want to qualify that with a caution that I've had far too many "Oh, yeah. I never considered that issue." moments while reading through other, more heavily commented-on RFCs.

dekellum:
So can we merge this and move on to considering the implementation PR? 🙈 🤞

dtolnay:
Thanks for the ping @KodrAus. Looks like I missed your deadline but my reaction anyway:

I don't think this RFC is sufficiently motivated as written. The std dbg macro, as explained in its documentation that is also quoted in this RFC, is intended for ephemeral debugging use cases. The thing about ephemeral debugging is that being easy to type and untype is valued above almost everything else. It seems to me like this RFC is trying to use easy-to-type as the primary motivation for a use case *other than* ephemeral debugging, which seems misconceived. If a log line, even at the debug or trace level, is going to exist for longer than one ephemeral debugging session, I think you will find that it is practically always possible and worthwhile to put in an extra 10 seconds of effort to add more context than just the name of a variable or stringify of an expression. At that point I would find it clearer for the log to occur in a statement by itself rather than being dbg-style nested inside a bigger expression or used as a return value.

I have not been involved in API design of the log crate so far so this is just my opinion as an outsider, but I would currently consider myself quite opposed to this change.

One thing that might change my opinion is if @dekellum or anyone else could share a single ironclad real world use case, maybe an extract of ~15 lines of code for context, in which they believe one of these new macros is the best and clearest way to express what the code needs to do. In other words show me the perfect use case for any one of these macros. I don't see that in the RFC text; all the code seems to be toy examples intended to demonstrate syntax only, not intended to justify why the new macros improve on the status quo.

Centril:
> I think you will find that it is practically always possible and worthwhile to put in an extra 10 seconds of effort to add more context than just the name of a variable or stringify of an expression.

If we take an application such as [rfcbot](https://github.com/anp/rfcbot-rs/search?q=debug%21%28&unscoped_q=debug%21%28) there are some places where `debug!(..)` is used and a custom message is provided. I can only say that from my perspective, having the origin of the logging be printed would have been useful when debugging rfcbot since files + line numbers tend to be *more unique* and therefore makes it easier to find the relevant places.

Getting the value you gave back is also nice since it doesn't force you to make temporaries just for the purpose of logging.

dtolnay:
> I can only say that from my perspective, having the origin of the logging be printed would have been useful when debugging rfcbot since files + line numbers tend to be _more unique_ and therefore makes it easier to find the relevant places.

File and line number are already present in the log record (https://docs.rs/log/0.4.6/log/struct.Record.html). It sounds like rfcbot will want to configure its logger to include that file and line number in its output.

As far as I can tell from the RFC and the implementation PR, this RFC does not propose doing anything different with file and line information than any of the existing log macros. Without using a logger that emits file and line in the output, rfcbot would continue not having file and line in its log output even if it switches to use logv macros. Sorry if I missed your point -- my interpretation of your response does not seem related to my concerns.

> Getting the value you gave back is also nice since it doesn't force you to make temporaries just for the purpose of logging.

This is the easy-to-type argument that I explained I find not at all compelling. To be more explicit, my perspective is that:

```rust
// I would never use this because file + line + "n / 2" is not sufficient
// context to make this log line meaningful to me long in the future.
let m = debugv!(n / 2) - 1;

// I would never use this because it would be more clearly expressed with
// the log being its own statement, despite taking more lines.
let m = debugv!("context context {} = {:?}", n / 2) - 1;
```

Centril:
> File and line number are already present in the log record (https://docs.rs/log/0.4.6/log/struct.Record.html). It sounds like rfcbot will want to configure its logger to include that file and line number in its output.

Oh... right; then my point is wrong... Thanks for the correction!

> This is the easy-to-type argument that I explained I find not at all compelling. To be more explicit, my perspective is that:

From my perspective, it's not really about being more easy-to-type but rather about having to invent temporary names for things that aren't semantically significant; temporary `let` bindings sometimes also require introducing a new block (e.g. compare `pat => expr` vs. `pat => { let tmp = expr; ... }`) and so they encourage right-ward drift. Now, I'm not philosophically opposed to `let` bindings or anything like that... I just think you shouldn't be forced to use them when they are not semantically important. Of course, readability is often quite subjective... :)

KodrAus:
> Looks like I missed your deadline but my reaction anyway

That date was just something I pulled out of thin air to try make things a bit more tangible. I think it's worth working through your feedback first @dtolnay.

> This is the easy-to-type argument that I explained I find not at all compelling.

That's a fair point, the main motivation for this is as sugar for capturing the evaluation of some expression. I've been thinking of it as just an alternative way to write log statements that appeals more to some people.

As for not using a raw `logv!` statement because it doesn't have enough context, I agree, but in languages that support it I find myself avoiding pushing additional context through log statements directly, and relying on some ambient context in a thread-local or something attached to a logger to carry that extra info. That way you don't clutter up the local statement with trivia, and you don't lose that trivia when you're faced with just the formatted log record later.

KodrAus:
So maybe what we need to do is look at a project like `rfcbot` and see whether we could really use these new macros to improve its logging?

dtolnay:
I would like to wait for @dekellum to share a use case that they feel perfectly calls for one of these new macros (and include it in the RFC text). I think that is the crucial missing piece before we can form a meaningful shared opinion on this RFC.

<br>

@Centril:

> From my perspective, it's not really about being more easy-to-type but rather about having to invent temporary names for things that aren't semantically significant

This doesn't make sense to me because if something is **not semantically significant** then it shouldn't be in log messages. If something *is semantically significant* then it is possible to pick a variable name as well as describe in words in the log message what significance the logged value holds.

I think your argument would be easier to articulate and evaluate in the context of a concrete snippet of real world code, which is why I am still pushing for that.

> temporary `let` bindings sometimes also require introducing a new block (e.g. compare `pat => expr` vs. `pat => { let tmp = expr; ... }`) and so they encourage right-ward drift.

I acknowledge that rightward drift is a thing. In the absence of a concrete use case, I don't acknowledge that doing more things in the same expression solely to avoid indentation is going to result in the clearest solution.

I think this argument too would be easier to evaluate in the context of a real world code snippet.

<br>

@KodrAus:

> in languages that support it I find myself avoiding pushing additional context through log statements directly, and relying on some ambient context in a thread-local or something attached to a logger to carry that extra info. That way you don't clutter up the local statement with trivia, and you don't lose that trivia when you're faced with just the formatted log record later.

I used the word context but didn't explain what I mean. I don't mean context as in other values in scope (like what file path we tried to open that failed, or what size of buffer is being used, etc). I mean human language that explains why some value appears in my log. For example this pair of log lines in rfcbot: [[1]](https://github.com/anp/rfcbot-rs/blob/16d64daa187d4bd9d6858db71881c04487814259/src/github/nag.rs#L975) [[2]](https://github.com/anp/rfcbot-rs/blob/16d64daa187d4bd9d6858db71881c04487814259/src/github/nag.rs#L991). Which would you rather see a year after writing the code?

```
[nag.rs:975] Command is to resolve a concern (actually-a-black-box)
[nag.rs:991] Found a matching concern (actually-a-black-box)
```

```
[nag.rs:975] concern_name = "actually-a-black-box"
[nag.rs:991] concern_name = "actually-a-black-box"
```

You need context to know what is going on with a value that makes it semantically significant (in @Centril's phrasing). Usually this means a verb. In fact every log message in that file uses a verb as context.

> So maybe what we need to do is look at a project like `rfcbot` and see whether we could really use these new macros to improve its logging?

I think the RFC needs to include a snippet of real world code that illustrates one of the macros being used in an ideal way -- where the author believes it to be the best solution. Code from rfcbot would meet my bar for real world code.

KodrAus:
> human language that explains why some value appears in my log

Right, I guess I've been thinking of this with post-structured-logging blinkers, where you'd write:

```rust
let command_name = debugv!("Command is to resolve a concern.", get_concern_name());
```

and `get_command_name()` and `actually-a-block` would end up attached as extra data. That snippet wouldn't be useful today, you'd have to include format replacements for the expression and value or you won't see them.

The output of final records not being as useful in the `logv` macros is a valid concern though that we should have an answer to.

dtolnay:
I don't see this discussion being productive until someone has proposed a single actual use of any one of these macros. I am unsubscribing for now, but kindly ping me again when anybody has demonstrated a concrete use case that they believe is a good fit -- or go ahead with the RFC if the consensus among others is that hypothetical use cases are sufficient motivation.

I am looking for a single use case that you couldn't write more clearly some other way without the new macros, such as in the case of @KodrAus's last comment which would be more clearly written as:

```rust
let concern_name = get_concern_name();
debug!("...", concern_name);
```

I would insist on the use case being drawn from real world code as opposed to hypothetical or contrived or abstract, and with enough surrounding code to evaluate whether the proposed macros are the most appropriate solution.

rfcs/0001-inline-expression-value-logging.md
KodrAus: Thinking about real-world examples more, I'm guessing these macros would be useful as a way to 'annotate' an expression to trace your way through interesting operations. That suggests we might actually want the expression and format to be reversed in the macro so the most relevant info when scanning the code appears first:

```rust
let rem = deadline - Instance::now();
```

becomes:

```rust
let rem = tracev!(deadline = Instance::now(), "{value:?} remaining ({expr})");
```

where we could interpolate the expression and value as named args into the format.

dekellum:
I'm not sure I can offer what you are asking for, or otherwise convince you, @dtolnay. The proposed feature offers a few relatively minor conveniences and is subject, as has been suggested, to style preference. Obviously, every character of log message output can be also be achieved with the existing macros in statement position.

I won't partake in re-litigating the `std::dbg!` macro addition [RFC 2361]. Some of your argument for only using the existing logging macros seems to me to be similar to arguments made against that and a prior RFC, e.g. to just use `std::eprintln!` in all cases. Did you weigh in on that RFC with a consistent opinion? Would @Centril care to comment on that process, history, and summarize the compelling pros/cons given for `dbg! vs `eprintln!`?

[RFC 2361]: https://github.com/rust-lang/rfcs/pull/2361

## This is my limited, _real code_ experience on the matter:

`std::dbg!` doesn't work very well with a project that already has `log` configured, at least in a unit test context, which is very much my preferred context for debugging:

* With `cargo test`, STDERROR configured log output isn't captured, while `dbg!` is!?! (At least in my test setup. Apparently that cargo feature does not work at the level of redirecting the process STDERROR file handle? Hooked in at higher level?) In passing tests, unless you are aware of and use the rather elusive and cryptic `cargo test -- --nocapture`, your `dbg!` and log output _will not both be shown_. In failing test runs, the two output sources are printed out of order, not interleaved.

* If you figure out `-- --nocapture`, you'll notice (see below) that the output formatting is jarringly in two different formats: one configurable by the logging implementation, the other hard-coded by `std::dbg!` which acts effectively like its own micro-logging system.

* `std::dbg!` is hard-coded to use "pretty" multi-line format (via `{:#?}`) which is also jarring by normal logging conventions.

* `std::dbg!` isn't beholden to log's level or other filtering and thus can't be kept in a project.

* I can turn on thread-name logging in my logging configuration, but not for `dbg!`

An excerpt of a debugging session with `log::trace!`, `log::debug!`, and `std::dbg!`:

``` log
TRACE mio::poll: registering with poller
TRACE tokio_threadpool::builder: build; num-workers=2
TRACE mio::poll: registering with poller
TRACE tokio_threadpool::sender: execute; count=1
TRACE tokio_threadpool::pool:     -> submit external
TRACE tokio_threadpool::pool: signal_work -- notify; idx=1
TRACE tokio_threadpool::pool: signal_work -- spawn; idx=1
[body-image-futio/src/futio_tests/server.rs:73] Tuner::new().set_buffer_size_fs(17).finish() = Tunables {
    max_body_ram: 196608,
    max_body: 1073741824,
    buffer_size_ram: 8192,
    buffer_size_fs: 17,
    size_estimate_deflate: 4,
    size_estimate_gzip: 5,
    size_estimate_brotli: 6,
    temp_dir: "/tmp",
    res_timeout: None,
    body_timeout: Some(
        60s
    )
}
TRACE tokio_threadpool::sender: execute; count=2
TRACE tokio_threadpool::pool:     -> submit external
TRACE tokio_threadpool::pool: signal_work -- notify; idx=0
TRACE tokio_threadpool::pool: signal_work -- spawn; idx=0
TRACE tokio_threadpool::worker: Worker::sleep; worker=WorkerId(1)
TRACE tokio_threadpool::worker:   sleeping -- push to stack; idx=1
TRACE tokio_threadpool::worker:     -> starting to sleep; idx=1
```

As this _real world example_ shows, it _can actually be useful to use verbatim, a rust code expression_, `stringify`-ed directly as the logging context. _I just want the same feature as `std::dbg!`, but directed through a single, configurable, logging system!_

My first reaction when I saw the rust RFCs was: `dbg!` vs `debug!`, why so important to save 2 characters?  Why isn't this hooking into `log`?

With the `std::dbg!` djinni out of the bottle, it would now be helpful for `log` to offer a more direct, but easily discovered super-set.

## @dtolnay, in initial comment:
> It seems to me like this RFC is trying to use easy-to-type as the primary motivation for a use case other than ephemeral debugging, which seems misconceived. 

This exposes a weakness in my terse RFC motivation section, _but not the feature_. Its perfectly appropriate to add, then shortly remove, uses of the `-v` macros, or to iteratively replace `-v` macros with the non-v macro's (only one character to delete) for nicer formatting or context. Its also perfectly appropriate to use the `-v` macros in statement position, for example:

``` rust
tracev!(some_long_function_call(x, y, z));
\\-- trace log message: some_long_function_call(x, y, z) = ()

\\ or...
tracev!(some_long_function_call(x, y, z));
\\-- trace log message: some_long_function_call(x, y, z) = "return value lost"

\\ or...
let r = tracev!(some_long_function_call(x, y, z));
\\-- trace log message: some_long_function_call(x, y, z) = "return value bound"

```

> If a log line, even at the debug or trace level, is going to exist for longer than one ephemeral debugging session, I think you will find that it is practically always possible and worthwhile to put in an extra 10 seconds of effort to add more context than just the name of a variable or stringify of an expression.

If your world view is that each log line (`Record`) should be fully formed, Queen's English sentences with punctuation, then you will most certainly _not_ want to keep use of the `-v` macros around.  Personally I think that might be a goal with `info` or `warn` or `error` level messages, but not so much with `debug` or `trace`. The above real world logging excerpt shows TRACE output that is pretty consistent with arbitrary code expressions, as does this excerpt:

``` log
TRACE tokio_reactor: loop process - 1 events, 0.000s
TRACE hyper::proto::h1::conn: flushed({role=server}): State { reading: Init, writing: Init, keep_alive: Idle, error: None }
TRACE tokio_threadpool::worker:     -> wakeup; idx=1
TRACE hyper::proto::h1::conn: Conn::read_head
DEBUG hyper::proto::h1::io: read 0 bytes
TRACE hyper::proto::h1::io: parse eof
TRACE hyper::proto::h1::conn: State::close_read()
DEBUG hyper::proto::h1::conn: read eof
TRACE hyper::proto::h1::conn: read_keep_alive; is_mid_message=true
TRACE hyper::proto::h1::conn: flushed({role=server}): State { reading: Closed, writing: Init, keep_alive: Disabled, error: None }
```

How I intend to use it, and what I think will practically happen, with this feature in place, while developing and debugging code, is:

1. Add `tracev!`, `debugv!`, and occasionally, `infov!` calls as convenient for debugging and demonstrating correct behavior.
2. When getting closer to release grade changes, refine your logging by removing some _-v_ macro calls, and replacing some (but not all) with the existing message formatting macros in statement position, making the messages more of _The Queen's English_.
3. Check-in (`git commit`) your code with `tracev!`, `debugv!` macro calls in place. If those calls survived to this step, then they are potentially useful in the future to you and other developers, just like the current use of `trace` and `debug`.
4. Continue by repeating these steps at (1) and in parallel with other developers doing the same.

(Am I the only one who uses the same process above, with the existing `debug!` and `trace!` macros?)

> At that point I would find it clearer for the log to occur in a statement by itself rather than being dbg-style nested inside a bigger expression or used as a return value.

Sounds like a general procedural vs functional programming style argument? Some will certainly like to use `-v` more than others. Still others will just stick with Haskell!

## Q: So `dbg!` was added to rust 1.32 `std` and the prelude (for real). Why do we have react to that, by adding something similar to log?

A: Because (see above) `dbg!` doesn't work well once logging is configured, and new users will expect that much parity and consistency!  I don't want to have tell someone learning/working with me, not to use `dbg!`, with no direct replacement in `log`!

## Q: Do the benefits justify the cost of maintaining the feature?

A: (Good question, me.) Outside of this RFC litigation cost, documentation and tests, it costs nearly nothing! Its 65 lines of macro rules code in total, that funnels into the existing `log!` macro. It might be nice if it integrated with structured logging, if/when that lands, but that's not even a requirement—the answer could be to use the existing (non-v) macro's only, if you want to use structured logging.

**You are not** going to count the number of public macros and get upset that the '-v' macros doubles them.  Why aren't you? Because if that was your concern, you would never have added a macro for each of the 5-logging levels, you would have just stuck with the one `log!(Level,…)` macro!

_So **yes**, ultimately, I think the convenience benefits for all ecosystem users, justify the costs of adding this to `log`!_

(I'm not so sure my personal costs have been justified at this point, however.)

## Conclusion

**TL;DR**: I'm _not_ misguided.  `std::dbg!` has shortcomings when used with a project that has `log` configured. The value of the feature does _not_ hinge on the _-v_ macros being long lived in code.  The fact that _-v_ macros  *could* be long lived (e.g as debug or trace level) is just a _bonus_ with inclusion in _log!_.

More seriously, another potential point of compromise on this, to reduce potential future costs more, would be to add the '-v' macros in under a default-on feature flag, with a disclaimer in the rustdoc that the `log` crate offers _no stability guarantees_ with them and they may become default-off, deprecated, and/or removed in the future—in particular if they are rarely used.  (Fuck. I mean, in most other languages, the _0.4.6_ crate version number would have already told you that! …I digress. It would be better to cover that in a blog post.) This also would allow `log` consumer project leads to govern enabling/disabling the feature in there own Cargo.toml.

I'll await your feedback before adding the non-snarky parts of this to the RFC motivation section. Thanks, as always, for your empathy and consideration!

rfcs/0001-inline-expression-value-logging.md
dekellum: Agreed. I touch on this a little in my below response to @dtolnay, going to resolve these here.

rfcs/0001-inline-expression-value-logging.md
dekellum: I actually had them named in my implementation branch at one point. It's easy to do. As you likely know, you can still use the position references,  `"{1} {0}"` even with named arguments, so I could resurect that in the implementation or you could add it later.

dekellum: Also, in my own limited experimentation I like using "{0} -> {:?}" as the format string and I wonder if that would be a better default format, as "->" doesn't appear in any expressions I'm aware of, but can of course be used for assignment in an expression position. Thoughts?

Centril:
> Would @Centril care to comment on that process, history, and summarize the compelling pros/cons given for `dbg! vs `eprintln!`?

The `dbg!(x)` macro is meant to be reachable from anywhere (e.g. the playground and *in seconds*), be maximally easy to write and be unimposing wrt. existing code and typing patterns. The reason why it takes `x` and then returns it back is so that you can do things like this:

```rust
    match 1 {
        x if dbg!(x) == 1 => {},
        _ => {}
    }

fn factorial(n: u32) -> u32 {
    if dbg!(n <= 1) {
        dbg!(1)
    } else {
        dbg!(n * factorial(n - 1))
    }
}
```
You can get back the original `factorial` function by simply removing `dbg!` calls. You don't need to remove any let bindings or such; conversely, to introduce `dbg!`, it's not necessary to introduce let bindings in a debugging session.

> `std::dbg!` doesn't work very well with a project that already has `log` configured, at least in a unit test context, which is very much my preferred context for debugging:

Integration with `log` was a non-goal and I think we never considered unit tests (they are well served by asserts as primary mechanism and `dbg!(..)` can successfully be used while developing tests).

> My first reaction when I saw the rust RFCs was: `dbg!` vs `debug!`, why so important to save 2 characters? Why isn't this hooking into `log`?

Easy... it takes more time to configure `log` than it takes to type `println!(...)` so the latter wins almost every time. `dbg!(x)` has to be more reachable and faster to use than `println!(...)` and also provide context; it does all of those things. As for `debug!` vs. `dbg!` and length, we simply chose the latter because `debug!` was "already taken".

> Sounds like a general procedural vs functional programming style argument?

Haskell has both `let x = expr in expr` bindings and `where` clauses. This is more a distinction between pointy and pointfree style. Tho it's more common to find people of the pointfree persuasion in Haskell circles.

---------------------------------

As for real code, having the features here available, I think I would rewrite:
```rust
// we want to panic if we're unable to find any of the usernames
let parsed_teams = teams::SETUP.team_labels().collect::<Vec<_>>();
info!("parsed teams: {:?}", parsed_teams);
```
as:
```rust
// we want to panic if we're unable to find any of the usernames
let parsed_teams = infov!(teams::SETUP.team_labels().collect::<Vec<_>>());
```
or:
```rust
// we want to panic if we're unable to find any of the usernames
let parsed_teams = teams::SETUP.team_labels().collect::<Vec<_>>();
infov!(parsed_teams);
```

I would rewrite:
```rust
        match Pool::builder()
            .max_size(CONFIG.db_pool_size)
            .build(manager)
        {
            Ok(p) => {
                info!("DB connection pool established.");
                p
            },
            Err(why) => {
                error!("Failed to establish DB connection pool: {}", why);
                panic!("Error creating connection pool.");
            }
        }
```
as:
```rust
        match Pool::builder()
            .max_size(CONFIG.db_pool_size)
            .build(manager)
        {
            Ok(p) => infov!(p, "DB connection pool established, {:?}."),
            Err(why) => {
                error!("Failed to establish DB connection pool: {}", why);
                panic!("Error creating connection pool.");
            }
        }
```

dekellum:
@Centril:
> we simply chose the latter \[`dbg!`\] because `debug!` was "already taken".

And for that in particular, and for all the other details you are providing, I am forever in your debt!

dekellum:
@Centril, to continue this 💓fest, this works too:

``` rust
    panic!(errorv!("fatal attraction💓!"));
```

**DEFINATELY NOT** in this RFC or its implementation PR, but seperately, I forsee the creation of the ellusive `fatal!` macro. Another one, yah!

KodrAus:
In this latest round of examples, we've been a bit fast-and-loose with whether or not the expression and value are actually interpolated into the format. Using the `format_args!` macro we wouldn't be able to do that (unless std was changed to emit a warning instead of an error for unused arguments that you could ignore). It seems like a bit of a papercut to have to include them if you don't want to. Without relying on a proc macro, we could take a message independently of the format and put them together ourselves:

```rust
match Pool::builder()
            .max_size(CONFIG.db_pool_size)
            .build(manager)
        {
            Ok(p) => infov!(p, "DB connection pool established."),
            Err(why) => panic!(errorv!("Failed to establish DB connection pool", why)),
        }
```

where `infov!(p, "DB connection pool established.")` expands to something like:

```rust
format_args!("{msg} ({fmt})", msg = "DB connection pool established", fmt = format_args!("{expr} -> {value:?}", expr = "p", value = p));
```

with an optional extra argument for the expression format:

```rust
infov!(p, "DB connection pool established.", "`{expr}` produced `{value:?}`");
```

If you just call `infov!(p)` then you'd just get the `{expr} -> {value:?}`.

dekellum:
Interesting idea.

In my last half joking case, that's why I suggested/threatened to propose a new `fatal!` macro, which could handle the gap in my composed example: `panic!(errorv!("msg"))` which, as you may have noticed, with current impl. will produce (msg = "msg") as the log message, before panic'ing, which is well short of perfect.
 
If there was a way to make that just work, that would certainly be a plus.  Having waited a while for this here, I don't think we should wait for a `format_args_lenient!` macro to stabilize in rust core that is lenient about missing placeholders in the format string, but if that ever happened, it would be backward compatible to start using it, wouldn't it?

In your proposed alternative, since both prefix and format are optional, wouldn't we need a marker token like '=', ';', "prefix:", "format:" or the like to distinguish them?  But If the normal case doesn't need it, it might not be too bad.. There is already the precedence of the "target:" marker.

KodrAus:
> In your proposed alternative, since both prefix and format are optional, wouldn't we need a marker token like '=', ';', "prefix:", "format:" or the like to distinguish them?

I would suggest we simply require a textual message be given if you also want to override the format for the expression and value.

dekellum:
But `format_args_lenient!` would solve it and avoid needing two parameters, wouldn't it?  It wouldn't be backward compatible to go back from two parameters to one, though.

KodrAus:
Hmm, I wouldn't rely on a `format_args_lenient!` becoming a thing in the standard library that we could plug into, and as an external proc macro I think we could go back from two parameters without breaking existing users of one parameter.

Having separate parameters for a message and format at least gives us some clear indication that there's a separate magic part of the format that uses format arguments you aren't explicitly providing.

dekellum:
Also, just to use a trick I just learned from @Centril, this works as intended:

```rust
    let panic_msg = format!("fatal attraction {}!", "💓");
    panic!(errorv!(panic_msg));
```

KodrAus:
@dekellum splitting the message into text and a format lets you write:

```rust
infov(p, "A message. We don't need to include any format args");
```

and guarantee that it will work on the compiler version that `log` currently targets and produce something like:

```
INFO: A message. We don't need to include any formst args (p = 42)
```

It reduces the barriers to associating a useful message with a `logv` call, which is important for longterm usefulness. Requiring a message before you can alter the value format assumes that customising the format of the value itself (display, debug, multiline etc) is of secondary importance to annotating the expression and value that gets logged.

dekellum:
OK, now I understand your position. I'd like to think about that more, but looks like it at least deserves a place in the alternatives of the RFC.

dekellum:
Note, also in the RFC, the order of parameters is currently reversed. To change that I would like to hear a detailed justification. My justification for current order is that it is consistent with the existing message logging (non-v) macros.

KodrAus:
My reasoning for reversing the order is that when scanning the code visually it keeps the actual expression on the left where it's closer to the assignment:

```rust
let x = infov!(1 / 2, "noise noise noise noise noise noise noise noise noise noise noise noise");
```

vs:

```rust
let x = infov!("noise noise noise noise noise noise noise noise noise noise noise noise", 1 / 2);
```

I don't necessarily think inconsistency with ordering for the message logging macros is a big deal here. For the message logging macros, the message itself is the significant part of the statement, so coming before args makes sense. For expression logging macros, the expression itself is the significant part of the statement.

dekellum:
I'd like to address the latest suggested _-v_ macro interface changes in an incremental fashion, here, starting with the idea to split the customization parameters into a message prefix parameter and format parameter.  The suggestion to change the parameter order is mostly orthogonal, so I will address that subsequently.

I agree that a contextual message prefix is going to be more common than changing the format. I also like this change, because it makes it possible to do as per an existing RFC alternative: _Only customize the value part of the format_. In other words, adding the message prefix parameter obviates the need to specify the entire format, and the library can _own_ the expression part of the format, also making value format customization easier to use, when it _is_ needed.

This change does come at the cost of some complexity. It _was_ 64 lines of macro rules. This adds another 22. Also in working through documentation (to make sure it was easily explainable), I added another 7 lines of documentation.  See an excerpt of this below:

```rust
//! The _default_ format string for the _-v_ macros is `"{} = {:?}"`, where
//! the `stringify!`-ed expression and resulting value are passed, in that
//! order. If the log record is not filtered out, the `Debug` implementation
//! for the type of the given expression value is used (`"{:?}"`).
//!
//! The log record can be customized via two optional parameters: a message
//! prefix string, and a format specifier for the value. Note that the former
//! is required, if passing the later:

let i = 32;
infov!(i);
infov!("", "{:?}", i);       // equivalent to above
infov!("index", i);          // non-empty prefix with additional context

infov!("index", "{}", i);    // use `Display` instead of `Debug`
infov!("index", "{:#x}", i); // hexadecimal format value
// ^-- info log message: index i = 0x20
infov!("index", "{:#?}", i); // use pretty, multi-line format
```

I primarily wanted to mention the size increases in case some other roving commentator shows up next month, to say this is too complicated, and shouldn't be added!

Having to specify at least a blank (`""`) message prefix in order to customize the value format is a minor nuisance.

Also it adds the possibility of some other mis-usage, one without a compiler error, and some other possible strange compiler errors:

``` rust
debugv!("bad format in prefix {:?}", i);
// --> debug log message: bad format in prefix {:?} i = 32
//  note: literal format placeholder in output

// compiler error: argument never used
debugv!("bad", "format", i);
// ^^^^^^^^^^^^^^^^^^^^^^^^^
// |
// argument never used
// formatting specifier missing

// compiler error: 5 positional arguments in format string, but there are 4 arguments
debugv!("bad", "{} = {:?}", i);
```

So I see pros and cons.  If we stayed with the existing RFC, '-v' macro interface for now, this could be revisited later/elsewhere, with a procedural macro that checked for 0 or 2 format placeholders in the single argument custom format argument. If it has 0 placeholders then we treat it like just the message prefix. If it has 2 then it is used as the _entire_ format. If it has any other number, issue a sane compiler error.  A procedural macro could also presumably make all of the above cases have sane compiler errors, with the 2-argument design, but we'd have to live with these quirks in the interim.

Thoughts on this design consideration?

KodrAus:
Thanks for working through this @dekellum!

> adding the message prefix parameter obviates the need to specify the entire format, and the library can _own_ the expression part of the format, also making value format customization easier to use, when it _is_ needed.

Ah that's an interesting point. In the larger Rust codebases I've worked on I've always ended up defining my own logging macros anyways, so it doesn't seem unreasonable to want to do this.

> Having to specify at least a blank (`""`) message prefix in order to customize the value format is a minor nuisance.

It is a bit of a nuisance, but spun another way it's an encouragement to improve the utility of your log macros by adding a message. If I saw an empty string in a logging expression in a code review I'd probably be a little suspicious.

> I primarily wanted to mention the size increases in case some other roving commentator shows up next month, to say this is too complicated, and shouldn't be added!

That's fair! The log macros are already a bit obtuse, but I think that's just how macros end up, so am not too worried about that.

Personally, over the long term I can see us completely rebuilding the macros in `log` at some point, either internally or by shifting them out into a separate crate, but it makes sense to try offer the best end-user experience with the tools we have today. We'll have to live with them for a long time in any case. Both approaches have pros and cons, but in my opinion the split message/format doesn't surface its cons in the happy-path. Still, I'm interested to hear what other folks think about that.

dekellum:
I've updated the RFC to include some additional motivation, and for the design change for the two customization parameters.

dekellum:
Regarding the design question of _-v_ macro parameter ordering:

The macro's can be made to work with either order, so I believe this is purely a question of clarity of design, including least-surprise, as well as long term ease of use.  My position remains to stay with the existing design and RFC (before and as recently updated). Here is a more detailed justification:

(1) If you consider the optional target, which is first in _every_ public `log` macro, the proposed reversed ordering has required parameters sandwiched between optional parameters on both sides, e.g:

``` text
debugv!([target: "target",] <expression>, ["prefix",] ["value-format"])
```

This would make explaining the parameters that much harder. Or would you prefer, for this alternative, to also move the parameter target to the right? (And if so, where exactly?)

Whereas, the current RFC design has the only required parameter on the far right:

``` text
debugv!(\[target: "target",\] \["prefix",\] \["value-format"\], <expression>)
```

(2) Compared with the existing message formatting macros, parameters:

``` text
debug!(\[target: "target",\] "format string literal", \[<expression>,\] \[...\])
```

As previously noted, the current _-v_ macro design is more consistent with the existing message formatting macros. This extends to `logv!` consistency with `log!`.  Beyond just perceived consistency, as per the [Expected Usage] mentioned here and since added to the RFC: converting from _-v_ macros to non-v formatting macros, and also in the other direction, is expected. The rust code edit distance is minimized with the current argument order. For example, starting from:

``` rust
debugv!("refine", some_result);
```

…and converting to:

``` rust
debug!("refine some result: {:?}", some_result);
```

or vice versa.

(3) The current argument order matches the output order of the message:

``` rust
debugv!("prefix in code and in output", logged_at_end());
```

(4) Your prior example, @KodrAus, is at least 101 characters long, so rustfmt would yield:

``` rust
    x = infov!(
        "noise noise noise noise noise noise noise noise noise noise noise noise",
        1 / 2,
    );
```

Which doesn't obscure the expression. Also, I suspect that in common usage the prefix context will be much shorter given that the expression is also logged. If it really needs to be this long, it probably makes sense to factor out the logging into a `info!` macro on its own line, see (2).  Note that, IMO, its quite common to split out separate statements, adding let bindings, etc. as needed, to avoid long lines (that rustfmt might otherwise butcher).

(5) I'm not aware of any existing precedence in Rust for the proposed alternative ordering, i.e. a parameter that will be returned, should be first?  I wouldn't be surprised if some other language does have some precedence for some order, but it may or may not be relevant to Rust.

[Expected Usage]: https://github.com/dekellum/log/blob/rfc-v-macros/rfcs/0001-inline-expression-value-logging.md#expected-usage

rfcs/0001-inline-expression-value-logging.md
dekellum: Naming format args is no longer relevant (current RFC) given that only the value format is customizable.   Parameter order under consideration below.

KodrAus:
Hmm, I think we're risking becoming a bit of an echo-chamber on the ordering question. The parameter ordering is starting to seem like the kind of design decision we'd be better off determining through actual usage, right now it's a bit arbitrary and we could justify it in either direction. So based on that I'd be happy to stick with the order you've already specified, but reserve the right to change it based on user feedback before 'stabilization'.

Which brings up the question of how we'd like to roll this out. We don't really have a history of unstable features in `log`, but I think for `logv`, we could start by providing them through an opt-in unstable feature-gate, with the expectation that someone in a project like `rustc`, `cargo`, `rfcbot` or something else will spike out using these macros to see how effective they are and in turn potentially tweaking the macros in breaking ways based on that experience. Once they're actually seeing use then we could stabilize them by removing the unstable feature gate.

Any thoughts on that?

rfcs/0001-inline-expression-value-logging.md
```diff
@@ -0,0 +1,305 @@
...
+output logger dependency and configuration, and particularly with some
+debug/trace logging already in place, use of `std::dbg!` would be
+unwelcome in PRs, and likely less productive than a `log`-based
+equivalent.
+
+## `std::dbg!` doesn't work well in projects with `log` configured
+
+In the context of unit tests, when using `cargo test`, _stderr_ (or
+_stdout_) configured log output isn't captured, while `dbg!`, at least
+when used within the tests, is captured.  For passing tests, only log
```
dtolnay: I would like to understand this better before using it as motivation. How is it that a logger configured to stderr or stdout circumvents the test harness's capture of stderr and stdout? Is that a bug that needs to be fixed, eliminating the motivation?

There is value in successful tests remaining quiet.

KodrAus: This might be specifically directed at `env_logger`, which uses `termcolor` to write to the terminal by default. `termcolor` in turn doesn't use the `print!` or `eprint!` macros so it doesn't hook into some special unstable APIs those macros and `libtest` use to figure out if output should be captured. We're working around it in `env_logger` at the moment, but I'd call this a point-in-time issue.

dtolnay:
I have re-read the RFC. This is much improved from the original draft -- nicely done. The proposed design has improved and the motivation has been fleshed out.

Still I don't think this feature should be accepted. I understand that it is something many people would like to use, and I see that it makes certain code easier to type, and it results in output that is prettier than mixing log+dbg, but I don't think the cost–benefit is where it would need to be for this change to be a net positive on the experience of logging in Rust.

### Mastery

Mainly I am concerned that we are adding decisions and knowledge where there didn't used to be decisions and knowledge required. I see a lot of value for new and experienced log users alike from never having to pause and think whether code would be "cleaner" if we logged a warning the *other* way of logging a warning. I never want there to be a debate on a code review, or even a single comment asking the (probably less experienced) author to please do it the other way. I don't want people to try to articulate elaborate "best practices" for when to logv in expression position, when to logv in statement position, when to normal log that all just boil down to personal preference. I don't want people to stumble on logv performing a move of its arguments when as a beginner we all had to learn as a special case that println (and by extension, log) do not move arguments despite writing no `&` on them like for function calls.

In my view the entire experience of there now being more to learn, more steps before attaining complete mastery of the crate, and more *things to think* when writing code, make the feature not worth its benefits. I like that the log crate today is easy to master. I find it freeing -- because when you want to log, for the most part you just do it and it's clear how.

### Analogy

My position may sound weird because if people want to use a thing and it makes code easier to write and maybe read, is an obvious generalization of existing log and std APIs, and all the other benefits, then of course we should accept it. I think over time I have come to wish for more restraint in the design process than that.

I will point out one past design question where I think restraint came up in a way that shares a lot of similarities with this one -- the issue of `#{expr}` interpolation in the quote crate, https://github.com/dtolnay/quote/pull/88. I don't want for this to be a big distraction, but I bring it up because I think it will help convey the shape of my opinion on logv by analogy with something that the others on this thread don't already have an invested opinion on, so may find it easier to recognize the point of view.

See the thread, but basically the quote crate supports interpolation of single variables with `#var` and various people over time have proposed interpolating expressions with `#{expr}` to support e.g. `extern #{f.abi} fn #{f.name}`. This was:

- proposed independently by multiple users;
- an obvious extension of existing functionality;
- trivial to implement (4 lines);
- negligible increase in the surface area of the crate;
- easy to learn / remember / understand / teach;
- makes code easier to write, sometimes much easier;
- makes certain code easier to read;
- supported by Alex, one of the most frequent and most experienced users of the crate;

that is to say almost every good thing that you could say about a feature, and everything that we traditionally look for when evaluating and accepting features. And yet when I considered the aggregate experience of using the crate, and everything that entails, I felt that the feature would lead to a worse experience for users. In other words it makes the crate worse.

I have a very similar feeling on this RFC. I think it is good for so many reasons, and yet altogether it is a bad thing for this crate.

### Counterproposal

We keep using the existing log macros, we use dbg for ephemeral debugging, we live with the differences in formatting between them, and we fix env_logger (https://github.com/sebasmagri/env_logger/pull/127) to behave well under the test harness.

rfcs/0001-inline-expression-value-logging.md
dekellum: My observations are with the `fern` logger implementation, which I chose for its features with a CLI binary, and then figured would be best to stick with for test logging.  It ultimately uses `write!(std::io::Stderr,…)` I believe.   So it seems that the test harness isn't actually capturing stderr/out at the process level. Perhaps it can't without also abandoning test threads in a single process.  It seems pretty fragile.  Will it even work when `dbg!` or `eprinln!` are found in lib code, outside of test cfg?

dekellum: Digging around a bit more, I found this summary of the capture weakness of libtest pretty helpful:
 
https://github.com/rust-lang/rust/issues/50297#issuecomment-388988381

It suggests it is also vulnerable when other threads beyond the test thread log.

dekellum:
Yeah I didn't think I'd convince you of anything, @dtolnay! 

I do appreciate your intuition, but I think it embeds a certain logical fallacy regarding the current state of the ecosystem, circa rust 1.32.

* Existing `log` users, as easy to "master" as I agree it is, are not going to have much new to take in with the _-v_ macros, which they can slowly experiment with and adopt, or not, as they may choose.

* The new users that _might_ struggle with the extra choice of _-v_ vs non-v macros, or understanding the move semantics of _-v_, are now starting to use `std::dbg!`, likely before even reaching for `log`.  BTW `std::dbg!` seems to be quite well received.  When and if they do reach for `log` 0.4.6, they are going to be disappointed with the `log` + `std::dbg` experience and lack of a `log` equivalent for `std::dbg!`.  If instead the _-v_ macros are available, they would be able to take their `std::dbg!` learnings and simply re-apply them, more productively with `log`.

Said another way, your position on this would be much more compelling, to me at least, if `dbg!` hadn't, quite miraculously given its own obstacles, already made it into rust `std`.

Finally, the `log` + `std::dbg` experience can't be well fixed with changes to `env_logger`, since it neither the only logger, nor in control of the problems.  Hopes for making test output capture reliable, has been subsumed into a complete Test Framework Overhaul, rust-lang/rust#50297, and my guess from reading, is that is a bit far off. When and if that is in place, and at best, we still would be stuck with `std::dbg!`'s hard-coded format and lack of other features.

I want to make it clear, I'm not bashing `std::dbg!` here. I think its a good feature, and it makes some sense to do it the way they did it, like a micro-logging system. I (and I think others) just want parity in a single `log` API and system.

dekellum:
If you are following along… I was finally able to use some of my open source time budget for using `debugv` in a real setting. In this case its improve the logging directly in tests, and I intend to keep these `debugv` uses in place once merged. Some notes:

* Removes 46 lines, adds 27 lines of test code (includes whitespace, YMMV).
* I didn't feel the need to use either of the `debugv` customization parameters in any of these cases.
* In some cases, I made a variable name more human friendly for use as the expression in the log. That neither seems like cheating nor a bad thing to promote, actually.
* I found `debugv` composed nicely with try-`?` and `Option` or `Result::unwrap()` chaining, as commonly used in these tests. For example:

``` rust
    let record = debugv!(reader.read(&tune))?.unwrap();
```

This `BarcReader::read` can return `None` on end-of-file, so inserting `debugv` in that position gracefully logs that case, before it would panic on unwrap, and where the remainder of the test works with `Some(Record)`.  Yes, I know I could have also used `expect("not to read None")` or similar, but this also logs the happy path and keeps things compact.

Otherwise the changes are pretty boring, actually, which I think is a useful thing to demonstrate with a feature such as this, and the recent suggestion that it could be a distraction, or something worse.  There is no currying, or other forms of leftist propaganda in these changes!  In this setting, its much more of a 1-to-1 mapping of the intended use for `std::dbg!`, but better integrated and for keeping around.  See the diff. below:

https://github.com/dekellum/body-image/compare/dev...debugv-in-tests-0

KodrAus:
My starting point for this RFC was _Rust is an expression-oriented language, so it seems reasonable to offer expression-oriented log macros_. In that context, having an equivalent set of `logv` macros makes sense, but as @dtolnay has pointed out, that also comes with problems. Creating decisions where there were none before and inviting valueless discussion are problems when I think logging is something that is most effective when it's free of ceremony and roadblocks to adoption. Anything that's not clear is a roadblock, and is likely to result in things not being logged.

If we're going to continue down this path then I think we could alleviate some of the possible confusion around whether you should be using `log` or `logv` by only offering the `debugv` macro (like @dekellum's original PR). That way the decision process is: use `debugv` instead of `dbg` if it's available, and use `debugv` instead of `debug` if you're tracing your way through an expression and it's convenient. You should associate a contextual message with your `debugv` calls.

Centril:
If the sole purpose of the new macro is for debugging, then one possibility to consider is that `log` may expose `log::dbg!(..)` which take the exact same inputs as `std::dbg!(..)` (moving the inputs, move semantics, requiring `Debug`, etc.) and returns the same output.

If `use log::dbg` occurs and the macro is in scope, then `log`'s methodology for logging happens, otherwise the macro is chosen from the prelude. This eases the transition between std and log such that you don't really have to care which one you are using and so that you can switch at a moments notice. 

However, one important aspect of this is that `log` should commit to only using the input and output forms that the standard library provides. If the `log` crate moves outside of this, this may set expectations for the standard library, negate smooth transitioning, or be just confusing.

dekellum:
Related to the feature gate sub-thread, which I still owe you a response to @KodrAus, we _could_ separate `debugv!`and `tracev!` as feature _logv_ and the remaining _-v_ macro levels as feature _logv_extra_ (thinking out loud on feature names). The point being that project leads get to opt in to what extent they would like to have these in their own codebase.

@Centril, I would prefer to not, by default, shadow your `std::dbg!` name, but thank you for offering it!  Of course individual consumer projects could still decide to alias `debugv!` as `dbg!` if that is their preference.  They _could_ also make that a fixed one-argument macro, or accept our additional current 0-3 customization parameters, which IMO are nice if not always used addition in a _log_ context.

dtolnay:
I like the idea to add just log::dbg with the same input patterns as std::dbg. I would expect an RFC with justification explaining the advantages over simply using std::dbg for that use case (given that dbg use cases tend to be about getting the job done, not about beauty of output, so differences in the formatting of the filename/line between log output and dbg output doesn't seem so detrimental). But the idea seems like a promising direction to me!

dekellum:
> My starting point for this RFC was Rust is an expression-oriented language, so it seems reasonable to offer expression-oriented log macros. In that context, having an equivalent set of `logv` macros makes sense,

I've consistently agreed with that, and believe the current RFC is a good expression of, and design for, that goal.  Yes `tracev!` and `debugv!` happen to fill an important and immediate gap in relation to `std::dbg!` and testing, but I also took @Centril's proposed use cases for `infov!`, etc. seriously. Should I not have, @Centril?  I know you are perfectly capable of writing your own `infov!`-like macro if you want one, but I had treated your use cases as being indicative of how other users would start using the _-v_ feature in `log`.

> as @dtolnay has pointed out, that also comes with problems. Creating decisions where there were none before and inviting valueless discussion

See my response above, but TL;DR: Overblown. Could be applied to anything *new* in the logging space, like for example, _Structured Logging_.

Just to be clear at this point, I'm not going to write a new alternative RFC that dumbs this down to a single `log::dbg` macro, that:

1. Purposefully makes the log `API` subordinate to current and future libstd decisions regarding `std::dbg!`. This could lead to later making non-PATCH safe changes and release of `log`, for example, if @Centril manages to re-introduce the multi-value input, tuple-output support in `dbg!` in some later rust release.  Or, @Centril, is your motivation different? Really any future change to `std::dbg!` could require some complicated rust-version aware dance in `log!`. Have you all really thought this through?

2. Introduce a `log::dbg!` with no continuity with the rest of the `log` API. Presumably a `log::dbg!` could not even include an optional `target:` parameter, and would need to be hard-coded for pretty-print multi-line output despite that not being a `log` norm.  And for this purpose, you would also have me just pretend that the `log::tracev!` level didn't exist?

@dtolnay, can you explain how you are somehow more comfortable with this `log::dbg!`alternative?  Is it just that the imposed discontinuity avoids violating your envisaged sanctity of the existing `log` output macros?

I tried to offer several points of comprise, but honestly even the whole feature gate, non-stable concept is over extending. The _-v_ feature is purely macro-rules, and users that don't want these macros don't have to import them.  How much of this conservatism ("more restraint") just stems from this uniquely-rust-inflicted Semver cul-de-sac that the the `log` crate finds itself in? Specifically, the need to release a new shim "0.4.7" (which is non-PATCH safe from a Semver perspective) just to make a "0.5" point release?

dtolnay:
@dekellum:

> See my response above, but TL;DR: Overblown. Could be applied to anything _new_ in the logging space, like for example, _Structured Logging_.

Selecting features where the argument cannot be applied is not the point. The goal is selecting features whose benefits outweigh their costs. Despite an argument being applicable to a feature, it is possible for that feature to be worthwhile because it brings greater benefits.

> How much of this conservatism ("more restraint") just stems from this uniquely-rust-inflicted Semver cul-de-sac that the the `log` crate finds itself in? Specifically, the need to release a new shim "0.4.7" (which is non-PATCH safe from a Semver perspective) just to make a "0.5" point release?

None, I think. I'm not sure how that would have anything to do with it. My view continues to be that the RFC would be a net bad thing for this crate. I would hold the same opinion if we were designing the crate from scratch with no versioning considerations.

Similar to how I think the benefits of `#{expr}` do not outweigh the costs to include in a quote crate, even a brand new one. Similar to how the benefits of `++` increment operators do not outweigh the cost of including them in Rust.

All of these things would have undeniable benefits if added. All of them are things that people would happily use. Any of these, if it gets added, would have people excitedly thinking "finally! I have been waiting to use that". People would use them and love them because they only see the benefits; these features are *easy to want*. It is much harder in general for people to perceive drawbacks of things like any of these three (logv, #{expr}, ++) because it requires them to think (and have experience thinking) about the entire experience of a feature, rather than individual instances of using it.

`++` in particular is strikingly similar to this RFC. Both of them boil down to, why would you do in two lines what you could do in one line and save indentation? That's looking at the benefit, but does not take into account the drawbacks.

```diff
  match op {
-     Op::Write(byte) => {
-         outbuf[pos] = byte;
-         pos += 1;
-     }
+     Op::Write(byte) => outbuf[pos++] = byte,
```

dekellum:
@dtolnay:  I do hesitate to drag _Structured Logging_ in for example purposes, my apologize @KodrAus , but then again, it _at least_ has to do with logging!  If I understand @KodrAus correctly on that feature, he is going to just release a new set of logging macro's in another crate that are structured logging friendly, for example, preserving the message template and named argument as structure.   So this just moves the problem up beyond `log`, as a decision for new users and new projects as to _what_ logging API they are going to use.  Ode to those proverbial _good old days_, when it was easy to just use `log`, positioned as the one true rust logging API. Now people are going to need to debate the merits of different APIs in different crates.

Looks like tokio-trace might be including `std::dbg!` parity, but I'm not sure how general purpose a logging API that actually is?  I can't decide which to use!

https://github.com/tokio-rs/tokio-trace-nursery/blob/ad5e1eb4de578bc0223a402c455d005da3b44ca3/tokio-trace-macros/src/lib.rs#L5

With your position @dtolnay, are we effectively relegating `log` to a much smaller role as keeper of a `Logger` static atomic (e.g. `set_logger`) for a legacy logging system?

dtolnay:
I don't see it playing out that way, no. If it does play out that way, it won't be for lack of logv macros.

dtolnay:
On a re-reading of the thread, I noticed that you conflate the position of being against these new macros as 
"conservatism" -- https://github.com/rust-lang-nursery/log/pull/317#issuecomment-468768068. That may be me not having explained well enough how I meant "restraint". Let me try again so that I maybe alleviate your worry of becoming a legacy logging system in https://github.com/rust-lang-nursery/log/pull/317#issuecomment-468818537.

Conservatism is opposition to change or innovation. What I meant by restraint is different: critically evaluating tradeoffs. The example of `++` differentiates them. Including `++` in Rust would have been conservative because every other similar language has it. Critically evaluating the benefits and drawbacks and deciding not to include `++` was restraint but quite opposite to conservatism. Maybe "diligence" would be better word choice.

*Accepting features* is not a requirement for staying a successful and relevant library long-term. *Evaluating tradeoffs* is a requirement.

dekellum:
@dtolnay: Many trade-offs were evaluated in just taking this design and RFC to its current point.  You needn't lecture further on that as a concept. Unfortunately, I think you and me are just going to have to agree to disagree on our respective risk vs. benefit analysis of the current RFC.  

I'm still perplexed and missing an import facet of your position, regarding the above mentioned `log::dbg!` is different from the RFC and might somehow be acceptable to you, but I fear exploring that might be a further waist of both our time, so I won't ask again.

Sincerely, I thank you for your time and consideration.

rfcs/0001-inline-expression-value-logging.md
dekellum: Refined this motivation section in the RFC to link to the same and better describe the issue, and its awkward workarounds.

dekellum:
More _-v_ macro usage in a realistic setting: dekellum/body-image@0708dba7a

The _-v_ form composes well with futures 0.1 and its typical method chaining, here with some usage of the optional message prefix.   Still just in test code, but of course a well tested project will have a lot of test code.

KodrAus:
@Centril:

> If the sole purpose of the new macro is for debugging, then one possibility to consider is that log may expose log::dbg!(..) which take the exact same inputs as std::dbg!(..) (moving the inputs, move semantics, requiring Debug, etc.) and returns the same output.

That's an interesting idea that sounds promising to me too 🤔 If we did something like that it would probably be instead of something like `debugv` though so it might be worth exploring a little here. It seems like a fair alternative that's easy to motivate and has a clear single usecase, but I can think of a few tradeoffs to consider too:

- Being tied to the same API as `std::dbg` creates a risk that changes or improvements to `std::dbg` that we can't replicate here in `log` means it's no longer a true drop-in replacement for `std::dbg`. That could be a non-issue if we try to be `log`-aware when discussing changes that affect `std::dbg`.
- Not having a contextual message with the expression means `log::dbg` would be as ephemeral as `std::dbg`, but by virtue of living in `log` it could be interpreted as a non-quick-and-dirty alternative to `std::dbg`.

@dekellum:

> See my response above, but TL;DR: overblown.

Hmm, I wouldn't quite go that far. Creating confusion by offering a parallel and not obviously more applicable API is a concern that's tricky to measure and a bit speculative, but it's a valid one nonetheless.

> release a new set of logging macro's in another crate that are structured logging friendly ... So this just moves the problem up beyond log, as a decision for new users and new projects as to what logging API they are going to use.

The goal of that work is a little different. It's a point-in-time effort to unblock the benefits for the growing list of structured frameworks to integrate with `log` without blocking on a feature that needs a lot more design work. It should eventually find its way back into `log`, or we'll restructure `log`. It's not really important just yet.

> I tried to offer several points of comprise, but honestly even the whole feature gate, non-stable concept is over extending.

I have to disagree with this point. It does seem unnecessary if we think of this API purely in terms of its immediate impact on a codebase that imports `log`, but is important to give us a chance to see how they play out in practice before committing to that exact API proper. The feature-gate would be a tool for the maintainers of `log` to have some wiggle room, rather than a tool for consumers of `log` to ignore APIs they're not interested in. The point about restraint is wider than just `log`, but we do have some unique constraints that make breaking changes costly.

> we could separate debugv!and tracev! as feature logv and the remaining -v macro levels as feature logv_extra (thinking out loud on feature names).

That sounds like a fair starting point to me, but I would put either just `tracev` or `debugv` behind one gate and the rest behind another. That way we could get some idea of who's interested in the full set of expression-oriented macros vs just the `std::dbg` equivalent.

-----

Also, I realise we've never explicitly stated what our expectations on RFCs are here in `log`, so it's possible we're all on different pages with respect to what accepting this means. That's my bad. I haven't been considering these RFCs as a kind of binding contract to ship exactly what they outline. I've been considering them as a focal point for discussion about new features in `log`, and as a future historical artifact that captures that discussion that we can come back to. In that discussion/documentation context, we'll want to give ourselves a chance to change/add/remove things before actually committing to supporting the feature over the long term, even if that ends up looking a little differently to what's described in the RFC.

dekellum:
So I think the current RFC is a well reasoned and complete design for the _expression and value logging_ feature as conceived. Thanks for all your input helping to make the RFC and design better!

This seems now stalled, perhaps awaiting more compelling input, on the existential question: _Will users be confused by having the choice of using this?_

(TL;DR: Ultimately I think that prospective users are smarter than that, and will be well prepared by use of `std::dbg!`.)

To obtain more feedback, as well as to _get on with things_, I've gone ahead and implemented your earliest advice, @KodrAus, by publishing the work as the standalone [tao-log] crate.  Editing a project's Cargo.toml to add `tao-log` as a dependency (possibly replacing the `log` dep.) may be about as much work as setting additional non-default `log` features, were they available.

This is released under the same license as `log` and I've made it clear that this is derivative work of this RFC process and the original PR #316 under `log`.  I still believe  it would be better to include this in `log` proper, and I am willing to offer some time to do that, if this RFC is eventually approved.

Of course I would also appreciate any feedback you might have on _tao-log_ and its [documentation], here or via the [dekellum/tao-log] repo.  For example, its currently unclear if my changes for a new overview of log output macros would be well received here in `log`, or how that might need to change for non-default feature gates.

[tao-log]: https://crates.io/crates/tao-log
[documentation]: https://docs.rs/crate/tao-log
[dekellum/tao-log]: https://github.com/dekellum/tao-log

pravic:
What is holding us from approving this RFC?

dtolnay:
@pravic I am still strongly opposed to accepting this into the crate. See my writeups above, which still reflect my current opinion.

The only new information relative to the 2019 discussion is that the proposed macros have been available on crates.io for &gt;2 years (https://github.com/rust-lang/log/pull/317#issuecomment-469855550) and so far nobody outside the RFC author has found it worth using that crate.

Probably if that crate were to become somewhat widely used then we could reevaluate here.

KodrAus:
It's been a while since we've visited this. It doesn't look like something we'll accept in `log` itself at this stage, but the work is still valid and has been good to explore. Thanks for all the effort in putting this together!

I'll go ahead and close this one now to keep our open set of things closer to what's actively on our plates.

WalterSmuts:
> Probably if that crate were to become somewhat widely used then we could reevaluate here

@dtolnay Given that I specifically wanted this feature for the past year and gave a couple of attempts at trying to find it, I don't think this is a valid way to judge if folks are looking for this feature.

> and so far nobody outside the RFC author has found it worth using that crate.

That sounds a bit catty... Also wrong... Given that it's got 7K downloads

PR #325, opened 2019-03-18
Thumbv6 support

jonas-schievink:
As discussed in https://github.com/rust-lang-nursery/log/issues/285, this adds another initialization function, `set_logger_racy`, which doesn't use `compare_and_swap` and is unsafe.

This allows `log` to support platforms that lack CAS support, which includes `thumbv6`. I haven't added any other architecture since I don't know how the exact support matrix for atomics looks.

Closes https://github.com/rust-lang-nursery/log/issues/285

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1102,7 +1102,7 @@ pub fn max_level() -> LevelFilter {
 /// An error is returned if a logger has already been set.
 ///
 /// [`set_logger`]: fn.set_logger.html
-#[cfg(feature = "std")]
+#[cfg(all(feature = "std", atomic_cas))]
 pub fn set_boxed_logger(logger: Box<Log>) -> Result<(), SetLoggerError> {
     set_logger_inner(|| unsafe { &*Box::into_raw(logger) })
 }
@@ -1115,6 +1115,13 @@ pub fn set_boxed_logger(logger: Box<Log>) -> Result<(), SetLoggerError> {
 /// This function does not typically need to be called manually. Logger
 /// implementations should provide an initialization method that installs the
 /// logger internally.
+///
+/// # Availability
+///
+/// This method is available even when the `std` feature is disabled. However,
+/// it is currently unavailable on `thumbv6` targets, which lack support for
+/// some atomic operations which are used by this function. However, even on
+/// those targets, [`set_logger_racy`] is available.
 ///
 /// # Errors
 ///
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -1178,6 +1189,36 @@ where
...
+/// A thread-unsafe version of [`set_logger`].
+///
+/// This function is available on all platforms, even those that do not have
+/// support for atomics that is needed by [`set_logger`].
+///
+/// In almost all cases, [`set_logger`] should be preferred.
+///
+/// # Safety
+///
+/// This function is only safe to call when there is no chance that any of its
```
sfackler: I think the unsafety is more limited than this - it's just forbidden to call this method concurrently, right?

jonas-schievink: It's worded this way because AFAIK any access to `LOGGER` while this function runs would be a data race and thus UB. This includes any of the initialization functions, but also any functions doing any logging.

Thomasdezeeuw: I think logging while this function is running is fine. Because all logging function check `STATE` first and only return an actual logging when its `INITIALIZED`, at which point the logging is set.

jonas-schievink: Ah, I see, that makes sense.

jonas-schievink:
@sfackler Does this looks better to you? I'm not terribly good at writing docs.

src/lib.rs
```diff
@@ -1178,6 +1189,37 @@ where
...
+/// A thread-unsafe version of [`set_logger`].
+///
+/// This function is available on all platforms, even those that do not have
+/// support for atomics that is needed by [`set_logger`].
+///
+/// In almost all cases, [`set_logger`] should be preferred.
+///
+/// # Safety
+///
+/// This function is only safe to call when no other logger initialization
```
sfackler: I think it's probably a good idea to explicitly mention that it is safe to concurrently log while this function runs.

src/lib.rs
```diff
@@ -1116,6 +1116,13 @@ pub fn set_boxed_logger(logger: Box<Log>) -> Result<(), SetLoggerError> {
 /// implementations should provide an initialization method that installs the
 /// logger internally.
 ///
+/// # Availability
+///
+/// This method is available even when the `std` feature is disabled. However,
+/// it is currently unavailable on `thumbv6` targets, which lack support for
+/// some atomic operations which are used by this function. However, even on
```
sfackler: I'd get rid of the However here since the previous sentence started with that as well.

sfackler:
LGTM with 2 more little doc nits!

chocol4te:
Any update?

sfackler:
Oh sorry, didn't notice that last commit!

PR #333, opened 2019-05-16
Respect padding for LevelFilter Display

johndrinkwater:
This mirrors a similar commit fe073054a0999d1bb238cb0293757741e95b3588
that changed output for Level, it seems sensible if the user wishes to
output the logging filter level to afford the same formatting choices.


sfackler:
Thanks!

PR #338, opened 2019-07-02
Remove duplicate Rust 2018 documentation

dereckson:
This is a follow-up for changes 0da424552703 and 0f3ffb9d051c.


KodrAus:
Thanks @dereckson!

PR #339, opened 2019-07-03
Make it easier to create a Value from owned or erased external types

KodrAus:
This PR allows consumers to work around `Value`'s required lifetime constraints in two ways:

1. allowing us to create a value with any lifetime from some owned primitive, like u8, bool etc. This is what the `Primitive` type is all about.
2. allowing us to create a value from some type that _can_ produce a concrete value we can use, but doesn't guarantee it will live as long as the value itself. This is what the `Fill` and `Slot` types are all about.

The `Fill` API is a more advanced initialization strategy (it's _actually_ a kind of visitor that just tries very hard to look like initialization), so it isn't exported from the `log::kv` module. Instead, this PR makes the `value` module public and leaves `Fill` and `Slot` in there. We need something like this to be able to translate a `&'kvs dyn tracing::Value` into a `log::kv::Value<'kvs>`.

### DIFF
--- a/src/kv/mod.rs
+++ b/src/kv/mod.rs
@@ -3,7 +3,7 @@
 mod error;
 mod source;
 mod key;
-mod value;
+pub mod value;
 
 pub use self::error::Error;
 pub use self::source::{Source, Visitor};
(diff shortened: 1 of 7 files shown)


KodrAus:
I'll be a good citizen and add some tests for `Fill`...

PR #340, opened 2019-07-06
Prepare for 0.4.7 release

KodrAus:
[Current changeset since `0.4.6`](https://github.com/rust-lang-nursery/log/compare/0.4.6...master)

It's been a while since we've done a release and there's a few things sitting on `master` that people want to be able to use.

I've added a note to our current structured logging stuff so it's clear that we're not committing to exactly what's there stability-wise. There are hoops we could make users jump through to access that API besides an unstable feature gate, but that didn't seem really necessary to me.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,20 @@
 
 ## [Unreleased]
 
+## [0.4.7] - 2019-07-06
+
+### New
+
+* Support for embedded environments with thread-unsafe initialization.
+* Initial unstable support for capturing structured data under the `kv_unstable`
+feature gate. This new API doesn't affect existing users and may change in future
+patches (so those changes may not appear in the changelog until it stabilizes).
+
+### Improved
+
+* Docs for using `log` with the 2018 edition.
+* Error messages for macros missing arguments.
+
 ## [0.4.6] - 2018-10-27
 
 ### Improved
@@ -112,7 +126,8 @@ version using log 0.4.x to avoid losing module and file information.
 
 Look at the [release tags] for information about older releases.
 
-[Unreleased]: https://github.com/rust-lang-nursery/log/compare/0.4.6...HEAD
+[Unreleased]: https://github.com/rust-lang-nursery/log/compare/0.4.7...HEAD
+[0.4.7]: https://github.com/rust-lang-nursery/log/compare/0.4.6...0.4.7
 [0.4.6]: https://github.com/rust-lang-nursery/log/compare/0.4.5...0.4.6
 [0.4.5]: https://github.com/rust-lang-nursery/log/compare/0.4.4...0.4.5
 [0.4.4]: https://github.com/rust-lang-nursery/log/compare/0.4.3...0.4.4
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ exclude = ["rfcs/**/*", "/.travis.yml", "/appveyor.yml"]
 build = "build.rs"
 
 [package.metadata.docs.rs]
-features = ["std", "serde"]
+features = ["std", "serde", "kv_unstable"]
 
 [[test]]
 name = "filters"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "log"
-version = "0.4.6" # remember to update html_root_url
+version = "0.4.7" # remember to update html_root_url
 authors = ["The Rust Project Developers"]
 license = "MIT OR Apache-2.0"
 readme = "README.md"
--- a/src/kv/mod.rs
+++ b/src/kv/mod.rs
@@ -1,4 +1,16 @@
-//! Structured key-value pairs.
+//! **UNSTABLE:** Structured key-value pairs.
+//! 
+//! This module is unstable and breaking changes may be made
+//! at any time. See [the tracking issue](https://github.com/rust-lang-nursery/log/issues/328)
+//! for more details.
+//! 
+//! Add the `kv_unstable` feature to your `Cargo.toml` to enable
+//! this module:
+//! 
+//! ```toml
+//! [dependencies.log]
+//! features = ["kv_unstable"]
+//! ```
 
 mod error;
 mod source;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -267,7 +267,7 @@
 #![doc(
     html_logo_url = "https://www.rust-lang.org/logos/rust-logo-128x128-blk-v2.png",
     html_favicon_url = "https://www.rust-lang.org/favicon.ico",
-    html_root_url = "https://docs.rs/log/0.4.6"
+    html_root_url = "https://docs.rs/log/0.4.7"
 )]
 #![warn(missing_docs)]
 #![deny(missing_debug_implementations)]


dekellum:
Asking the obvious: Given the new albeit non-default _kv_unstable_ feature, from standard SemVer perspective, shouldn't this be released as 0.5 instead? Consider that any logging implementations _or_ applications that enable _kv_unstable_  will need to specify a minimum of 0.4.7 for the _log_ crate (which, as its a patch version, can be easily missed without minimum-version testing). 

Looking at the changes since 0.4.6 there are also a bunch of pure-fixes that are justifiably in a 0.4.7 patch release.  Should those be at least released first/separately for the sake of reserving a strict SemVer patch, in any case?

I do realize that releasing as 0.5, could cause other problems with cargo resolution of duplicate versions of log. This the standard justification for core rust crates releasing new features, or potentially breaking changes, in patch releases.

sfackler:
In Rust semver, 0.x.y versions treat y as the major version number, which is only incremented for breaking changes.

dekellum:
As exemplified by the phrase "Rust SemVer", if everyone has there own unique definition, then there is no useful convention. Is your definition, @sfackler,  somehow compatible with the below?

https://semver.org/

> Given a version number MAJOR.MINOR.PATCH, increment the:
>
>    MAJOR version when you make incompatible API changes,
>    MINOR version when you add functionality in a backwards-compatible manner, and
>    PATCH version when you make backwards-compatible bug fixes.

(From this, _kv_unstable_ should be a MINOR version change)

https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements

> This compatibility convention is different from SemVer in the way it treats versions before 1.0.0. While SemVer says there is no compatibility before 1.0.0, Cargo considers 0.x.y to be compatible with 0.x.z, where y ≥ z and x > 0.

dtolnay:
Rust semver is indeed different from semver.org. Below 1.0 we bump the MINOR for breaking changes and the PATCH for nonbreaking changes.

dekellum:
Over simplifying doesn't help your argument.  It pretty common, for rust crates in general, to put new features such as this _kv_unstable_, in a new MINOR (aka 0.5.0).  

Now maybe the fair answer is that for a core crate such as _log_; which though generally a private dep, breaks with duplicate versions, and that for such a new non-default feature, its just not worth the effort to bump the release to _0.5.0_ and force a large area of the ecosystem to upgrade.   

Still begs the question if the fully legitimate PATCH fixes should be released as 0.4.7, first.  It could be a useful release (and reserved releases number) if _kv_unstable_ 0.4.8 or 0.5.0 breaks something.

sfackler:
> It pretty common, for rust crates in general, to put new features such as this _kv_unstable_, in a new MINOR (aka 0.5.0).

That does not match my experience in any crate I'm aware of.

BurntSushi:
Me either. If a crate hasn't hit 1.0 yet, then I treat MINOR as if it were a MAJOR, and squash MINOR and PATCH into PATCH, effectively. This, of course, being motivated by "Rust SemVer."

dekellum:
Crates _clap_ (including when < 1.0.0), _actix(-web)_, and _hashbrown_ come to mind, having examples of (likely) compatible features queued for release in new MINOR versions.  Its true that pre-1.0, this is sometimes overshadowed by the fact that maintainers also have clearly breaking changes queued to introduce on the next MINOR (which is legal pre-1.0).

I think this is a fair question (should it be released as log 0.5.0) because once some dependency turns on a feature gate like _kv_unstable_, there is no turning it off.   No one has claimed this _couldn't_ be released as 0.5.0, so please make that case that its not worth the trouble to do so, vs. just releasing all queued changes as 0.4.7.

BurntSushi:
There is no good reason, as far as I can see, to put out a breaking change release just because an unstable feature has been added.

dekellum:
The usual, general reason as to why SemVer assigns compatible new features to MINOR releases, and more specifically here:

* In case enabling the _kv_unstable_ feature (for example, enabled by some log implementation, in its new PATCH release) breaks some projects in an unforeseen way.  Releasing it as MINOR version bump clearly makes less compatibility claims, and gives some opt-in time/flexibility downstream.

* Would preserve 0.4.x versions for currently queued (on master) and future clear PATCH fixes

Again, I'm not saying that this justifies the effort or other downsides to releasing as 0.5.0, but only that its _possible_ and could have advantages.

Thomasdezeeuw:
@dekellum I mostly agree with the points you're making, but can we be a bit pragmatic about this? It's a fairly minor, unstable feature that doesn't impact a great deal of code (outside the added module itself). Log is a crate used throughout the Rust ecosystem and updating all those crates is just a pain and not worth the effort my opinion. Crates that depends on this unstable feature can specify `log = ^0.4.7` as dependency and all others will continue to work.

KodrAus:
Hi @dekellum :wave: that is a valid point, and we are very careful about how we version this library.

For all intents and purposes, consumers of `log` should pretend the `kv_unstable` feature and the `kv` module don’t exist yet. There are a few competing needs that have been balanced into this approach:

- We can’t just land an API like structured logging support in a single go. It’s pretty much guaranteed that it’ll need big changes based on user feedback.
- We want people to contribute to the design and implementation of the structured logging support, and that’s much easier to do if the code lives on `master` where other contributions are made.
- We want people to try structured logging support. There are users who are happy to jump on the breakage bandwagon from `crates.io` because they’re looped in to the development process.
- We don’t want to block on-going maintenance of `log` itself on this new work. There’s a steady stream of contributions made that shouldn’t be held up by whatever state the unstable work is in.

We could make the unstable work harder to get at, which would prevent users from stumbling across it, but would also create contribution barriers. I’m open to new perspectives though!

dekellum:
In case its not clear, I agree that you should be able to comfortably/safely make log crate releases for preview and evolution of the structured logging feature. Its only a question of how.

> that is a valid point, and we are very careful about how we version this library.

Yes, so careful, that 0.4.0 should have been released, or some 0.4.x should be re-released as 1.0.0 (1).  Beyond being a correct representation of the practical state and expectations on this crate, it would not be a question that releasing new features should be done in a new MINOR, right?

But perhaps structured logging is also viewed, ironicly, as the single missing feature to be stabilized for that precious 1.0.0 release?  So how can we proceed, version wise?

> For all intents and purposes, consumers of log should pretend the kv_unstable feature and the kv module don’t exist yet. [...] It’s pretty much guaranteed that it’ll need big changes based on user feedback.

_Good intentions._ So your current plan is to release a series of 0.4.7+ patch releases interleaved with evolving changed to the non-default feature _and_ normal maintenance and legitimate PATCH changes.

But structured logging is indeed a useful feature. How long before a logging implementation enables _kv_unstable_ by default?   The enabled feature adds public fields to the core `log::Record` struct and builders, and interestingly, also raises the minimum supported rust version (MSRV) from 1.16.0 to 1.21.0.

Meanwhile there is no direct Cargo support for crate MSRV, much less crate×feature MSRV nor "unstable" features.  This 0.4.z PATCH versioning plan seems to be bending a lot of well established conventions, beyond just SemVer, to avoid bumping the MINOR version.

### Proposed alternative

Release this as 0.5.0 or, if you prefer, 0.5.0-alpha.0.

In this plan, I don't suggest hiding it behind a non-default feature gate. Any lib crates with a log dependency (including log output implementions) can, in new MINOR or MAJOR series release, choose to actively _use_ `kv` by depending on '^0.5' or to passively allow that use (even in a PATCH release) by depending on log version `>=0.4.6, <0.6".

_This, folks, is how SemVer is supposed to work._

And I offer my help, here or elsewhere, in addressing whatever rust-specific, weirdly-situational or self-inflicted challenges that make this hard to pull off or overly onerous.

(1) https://semver.org/#how-do-i-know-when-to-release-100

> How do I know when to release 1.0.0?
>
> If your software is being used in production, it should probably already be 1.0.0. If you have a stable API on which users have come to depend, you should be 1.0.0. If you’re worrying a lot about backwards compatibility, you should probably already be 1.0.0.

sfackler:
There are hundreds if not thousands of crates that directly depend on `log`. It is profoundly unreasonable to force every single one of them to go manually push their version constraints forward every time we add a new feature to `log`.

If you want to change the way that Cargo and the broader Rust ecosystem think about `0.x.y` versions, that conversation should probably be held in a broader forum than a PR in a single crate's repo.

PR #349, opened 2019-08-22
Fix documentation for static max_level_xxx filters

### LINKED ISSUE #348: Description of compile-time log filtering needs clarification
Reading this: https://docs.rs/log/0.4.8/log/#compile-time-filters, I stumbled across this phrasing:

> Log invocations at disabled levels will be skipped and will not even be present in the resulting binary unless the log level is specified dynamically

I don't understand how this can be true -- either the call site is completely eliminated when release_max_level_XXX is used, or it isn't. If it is, I don't see how any dynamic logging adjustment can make it re-appear...

So the question boils down to - if I have a `trace!` invocation, and I set `release_max_level_debug`, what will be present in the final binary instead of that `trace` call? A condition check? nothing?

vmalloc:
closes #348

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -195,9 +195,8 @@
 //! # Compile time filters
 //!
 //! Log levels can be statically disabled at compile time via Cargo features. Log invocations at
-//! disabled levels will be skipped and will not even be present in the resulting binary unless the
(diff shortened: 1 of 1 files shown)


sfackler:
Thanks!

PR #350, opened 2019-09-01
Add dyn Trait annotations, remove unneeded lifetime annotations.

bpfoley:
As of rust 1.37
* dyn Trait refs need to be explicitly annotated
* static variable decls don't need the 'static lifetime annotation.


sfackler:
We support Rust all the way back to 1.16.

bpfoley:
> We support Rust all the way back to 1.16.

Ah! My bad, I should have read the big print in the readme. Thanks.

PR #351, opened 2019-09-01
Remove derived trait Hash as PartialEq is explicitly implemented

bpfoley:
Found by clippy
https://rust-lang.github.io/rust-clippy/master/#derive_hash_xor_eq

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -334,7 +334,7 @@ static LEVEL_PARSE_ERROR: &'static str =
 /// [`log!`](macro.log.html), and comparing a `Level` directly to a
 /// [`LevelFilter`](enum.LevelFilter.html).
 #[repr(usize)]
-#[derive(Copy, Eq, Debug, Hash)]
+#[derive(Copy, Eq, Debug)]
 pub enum Level {
     /// The "error" level.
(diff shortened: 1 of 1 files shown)


sfackler:
We can't just not implement Hash anymore - that would break backwards compatibility. Our manual PartialEq impl is equivalent to what derive would generate.

bpfoley:
Ah... So what's the best thing to do here? Add `#[allow(clippy::derive_hash_xor_eq)]`? Leave it alone?

sfackler:
I'd probably just go with `#[allow(clippy::derive_hash_xor_eq)]`, yeah.

thomaseizinger:
> Our manual PartialEq impl is equivalent to what derive would generate.

Out of curiosity: I saw that we do a cast to `usize` instead of just using the generated impl. What is the difference and consideration in terms of backwards compatibility?

sfackler:
The derive-generated code was significantly larger and less efficient last time I checked. If that's no longer the case, we can derive them.

KodrAus:
It's been over a year now so I'll go ahead and close this one. Thanks for investigating this though @thomaseizinger!

PR #346, opened 2019-08-10
[WIP] Log macro kv

Thomasdezeeuw:
This is my attempt at modifying the `log` macro to add support for key values, as described in https://github.com/rust-lang-nursery/log/issues/343#issuecomment-513759469. I didn't get very far as I'm not the best at Rust macros.

Testing with `cargo test --features std kv_unstable` gives the following error:

```log
error: local ambiguity: multiple parsing options: built-in NTs tt ('arg') or 1 other option.
  --> tests/kv_macro.rs:23:87
   |
23 |     log!(target: __log_module_path!(), Level::Info, "some {} message {}", "great", 1, {
   |                                                                                       ^
```

For #343.

/cc @KodrAus @yoshuawuyts

Issue #343, opened 2019-07-22
key_value capable log macros


yoshuawuyts:
Continuation of https://github.com/rust-lang-nursery/log/issues/328#issuecomment-513588521.

## Description

`kv_unstable` is available in `log` and provides a `key_values` field on the `Record` to capture key-value pairs. However currently in order to write data to loggers `Record` types need to be constructed manually, rather than being able to use the log macros. As per [RFC#296](https://github.com/rust-lang-nursery/log/blob/master/rfcs/0296-structured-logging.md) we should experiment with logging macros out of tree, and use it to inform how to add key-value logging to log's macros.

## Constraints
[tbi]

@KodrAus could you perhaps clarify what the constraints are?

### Slog
The [slog crate](https://docs.rs/slog/2.5.1/slog/macro.log.html) uses `;"key" => "value"` pairs in the macros:
```rust
#[macro_use]
extern crate slog;

fn main() {
    let drain = slog::Discard;
    let root = slog::Logger::root(
        drain,
        o!("key1" => "value1", "key2" => "value2")
    );
    info!(root, "test info log"; "log-key" => true);
}
```

### Tracing
[tracing](https://docs.rs/tracing/0.1.3/tracing/macro.info.html) supports two kinds of log styles: "a message + key-value pairs", and "only key-value pairs":
```rust
use tracing::field;

let addr = Ipv4Addr::new(127, 0, 0, 1);
let conn = Connection { port: 40, speed: 3.20 };

info!({ port = conn.port }, "connected to {}", addr);
info!(
    target: "connection_events",
    ip = %addr,
    conn.port,
    ?conn.speed,
);
```
_Update this section with more examples_

## Creating a manual record
For reference, this is how to create a manual record with key-value pairs:
```rust
use log::kv::{Error, Visitor, Source};

struct Pair {
    val: String
}

impl Source for Pair {
    fn visit(&self, visitor: &mut dyn Visitor<'_>) -> Result<(), Error> {
        visitor.visit_pair("some key".into(), self.val.into())
    }
}

log::logger().log(&log::Record::builder()
    .args(format_args!("hello world!"))
    .key_values(&Pair { val: "some value".to_string() })
    .level(log::Level::Info)
    .target(module_path!())
    .module_path(Some(module_path!()))
    .file(Some(file!()))
    .line(Some(line!()))
    .build());
```

## Conclusion
[...]

Issue #328, opened 2019-04-29
Tracking issue for kv_unstable (structured logging)


KodrAus:
Tracking issue for the implementation of #296. See also #149 

- [x] Initial implementation (#324)
- [ ] Integration with frameworks and loggers:
  - [x] [`slog`](https://github.com/slog-rs/slog)
    - [x] Zero-allocation conversion between `slog::Value` and `log::kv::Value`. We should be able to do zero-allocation conversion from `log` to `slog`, but may need to allocate from `slog` to `log`.
  - [ ] [`tracing`](https://github.com/tokio-rs/tracing) (https://github.com/tokio-rs/tracing/issues/341)
    - [x] Zero-allocation conversion between `tracing::Value` and `log::kv::Value` in the `event!` macro (#339)
    - [x] Capture `std::error::Error`s as `Value`s
  - [ ] [`fern`](https://github.com/daboross/fern)
  - [ ] [`async-log`](https://github.com/rustasync/async-log)
  - [x] [`std-logger`](https://github.com/Thomasdezeeuw/std-logger) (https://github.com/Thomasdezeeuw/std-logger/pull/14)
    - [x] Some way to tell whether a `Source` has any key-value pairs (#329)
  - [ ] [`femme`](https://github.com/lrlna/femme)
  - [ ] [`env_logger`](https://github.com/sebasmagri/env_logger) (https://github.com/sebasmagri/env_logger/pull/137)
  - [x] [`antlog`](https://github.com/KodrAus/antlog) (an experimental wrapper to prove out requirements)
    - [x] A way to get specific values (#335)
    - [x] A way to coerce a value to a common type (#379)
  -  _Please suggest others to track!_
- [x] Explore macro support (#343)
  - [x] `log!("msg {}", interpolated, { key: value })` (#353) - We ended up having to revert this in `log`, but are experimenting externally
  - [x]  [`log_kv_macro`](https://docs.rs/kv-log-macro/1.0.6/kv_log_macro/)
  - [x] [`log_macros`](https://github.com/KodrAus/log_macros)
- [x] Integration with serialization frameworks:
  - [x] `serde`
    - [x] Conversion between `serde::Serialize` and `std::fmt::Debug` (https://github.com/rust-lang/rfcs/pull/2696). See [`serde_fmt`](https://github.com/KodrAus/serde_fmt).
  - [x] `sval` (#342)
- [ ] Micro benchmarks
- [ ] Documentation and examples
- [ ] Stabilization review
[...]

PR #353, opened 2019-09-01
kv macro support

yoshuawuyts:
Adds key-value support to the log macros. Supersedes #346. Refs #343 #328. Thanks!

## Syntax
```rust
femme::start(log::LevelFilter::Info).unwrap();
info!("hello");
info!("hello",);
info!("hello {}", "cats");
info!("hello {}", "cats",);
info!("hello {}", "cats", {
    cat_1: "chashu",
    cat_2: "nori",
});
```

## Tasks
- [ ]  write tests

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1416,32 +1416,8 @@ pub fn __private_api_log(
     args: fmt::Arguments<'_>,
     level: Level,
     &(target, module_path, file, line): &(&str, &'static str, &'static str, u32),
-    kvs: Option<&[(&str, &str)]>,
+    kvs: Option<&[(&str, &kv::ToValue)]>,
 ) {
-    // Ideally there would be a `From` impl available for this.
-    struct KeyValues<'a> {
-        inner: &'a [(&'a str, &'a str)],
-    }
-
-    impl<'a> kv::Source for KeyValues<'a> {
-        fn visit<'kvs>(&'kvs self, visitor: &mut kv::Visitor<'kvs>) -> Result<(), kv::Error> {
-            for pair in self.inner {
-                visitor.visit_pair(pair.0.into(), pair.1.into())?;
-            }
-            Ok(())
-        }
-
-        #[inline]
-        fn count(&self) -> usize {
-            self.inner.len()
-        }
-    }
-
-    let kvs = match kvs {
-        Some(kvs) => Some(KeyValues { inner: kvs }),
-        None => None,
-    };
-
     logger().log(
         &Record::builder()
             .args(args)
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1389,10 +1389,12 @@ pub fn __private_api_log(
     args: fmt::Arguments,
     level: Level,
     &(target, module_path, file, line): &(&str, &'static str, &'static str, u32),
-    kvs: Option<&[(&str, &str)]>
+    kvs: Option<&[(&str, &str)]>,
 ) {
     if kvs.is_some() {
-        panic!("key-value support is experimental and must be enabled using the `kv_unstable` feature")
+        panic!(
+            "key-value support is experimental and must be enabled using the `kv_unstable` feature"
+        )
     }
 
     logger().log(
@@ -1414,7 +1416,7 @@ pub fn __private_api_log(
     args: fmt::Arguments<'_>,
     level: Level,
     &(target, module_path, file, line): &(&str, &'static str, &'static str, u32),
-    kvs: Option<&[(&str, &str)]>
+    kvs: Option<&[(&str, &str)]>,
 ) {
     // Ideally there would be a `From` impl available for this.
     struct KeyValues<'a> {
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -1384,11 +1384,64 @@ pub fn logger() -> &'static Log {
 
 // WARNING: this is not part of the crate's public API and is subject to change at any time
 #[doc(hidden)]
+#[cfg(not(feature = "kv_unstable"))]
 pub fn __private_api_log(
     args: fmt::Arguments,
     level: Level,
     &(target, module_path, file, line): &(&str, &'static str, &'static str, u32),
+    kvs: Option<&[(&str, &str)]>,
```
KodrAus: We should use `&ToValue` here so that the structure of captured values can be retained. I sketched out the change in https://github.com/yoshuawuyts/kv-log-macro/pull/1

KodrAus:
Ah, we should also be careful not to regress the `info!("Hello, {value}", value = 42);` case, just based on [`kv-log-macro`](https://github.com/yoshuawuyts/kv-log-macro) it looks like our changes to the macro syntax may not support the `ident = expr` syntax?

It does also present a bit of a syntactic oddity, where values interpolated into the message would use the `ident = expr` syntax from `format_args`, but the captured key-value pairs use the `ident: expr` syntax like structs:

```rust
info!("hello {value}", value = "cats", {
    cat_1: "chashu",
    cat_2: "nori",
});
```

I don't think this is necessarily a problem though, but thought it was worth pointing out.

src/lib.rs
yoshuawuyts: Ohhhh, yesss! -- I wish I'd known about this sooner, haha.

yoshuawuyts:
@KodrAus in general it seems we don't have a lot of tests right now -- I was wondering how we might want to go about it?

KodrAus:
> we don't have a lot of tests right now

Yeh, we don't really for the macro syntax, do we. Maybe we could add an integration test file that can do something like this:

```rust
struct AssertRecord;

impl Log for AssertRecord {
    fn record(&self, record: &Record) {
        ASSERT.with(|a| a(record));
    }
}

#[test]
fn level_info() {
    // Sets the assert function in thread-local storage
    // then calls the log statement
    test_log!(
        info!("Hello, {value}!", value = "world"),
        |record| {
            assert_eq!("Hello, world!", record.args().to_string();
            assert_eq!(Level::Info, record.level());
        }
    );
}
```

So that we can build up a suite for the current syntax.

But I'm happy to sketch that out independently if you like? For now I think if we can try find a way to keep `ident = expr` interpolated values working then we _should_ be good!

yoshuawuyts:
@KodrAus ohhh, yeah looks like you were onto something there -- I found a bug in my impl which I've fixed, but more importantly the `name = val` interpolated args aren't indeed no longer working. I've added some basic tests for the macros to prevent this from regressing.

Though I'm not entirely sure how to fix it, so I'm all ears if you have suggestions!

KodrAus:
Hi @yoshuawuyts!

I'm just circling back to this, I'll spend some time playing with the macros too and see where I end up. We might end up having to tweak the syntax a little (which we could do to consider working more nicely with other extensions like #357).

yoshuawuyts:
@KodrAus did you have any luck getting this to work? If not, perhaps we could ask @dtolnay for help?

KodrAus:
Hi @yoshuawuyts! :wave:

Sorry I’ve been really scattered over the last few months so haven’t been able to make a lot of progress here. However I’m feeling less concerned about regressing this specific edge-case, and I also think [implicit inline format args](https://github.com/rust-lang/rfcs/pull/2795) makes the difference between our binding syntax for format args vs kv pairs much less significant.

So all that’s to say I’d be on board with pushing forward with this!

yoshuawuyts:
@KodrAus yay, that's excellent news! I'm very excited for this!

I just pushed a patch to remove the failing tests, meaning this patch should now be ready to be merged! It seems CI is currently not setup, so it can't verify this patch. But hopefully this is good!

KodrAus:
> It seems CI is currently not setup

Oh dear... Lemme just fix that real quick.

KodrAus:
@yoshuawuyts would you like to squash and rebase your commits on the current `master`? I'm thinking the reason actions aren't running here is because we're missing the `.github/workflows` directory on your branch.

yoshuawuyts:
@KodrAus rebased and force-pushed. Hoping this works! :crossed_fingers:

yoshuawuyts:
Heh, for odd reasons CI is no longer reporting on this branch. My guess is because the repo was migrated it caused some bug in GitHub's UI.

CI link is here: https://github.com/yoshuawuyts/log/runs/295531981 -- it seems `cfg_if` is failing to build?

KodrAus:
Ah it looks like `cfg-if` has removed their explicit MSRV, so I'm trying to get everything back into a good state in #362.

yoshuawuyts:
Not behind a computer right now but we should try running CI again. Looks like #362 got merged two weeks ago.

yoshuawuyts:
@kodraus Wooot! So happy this got merged! Thank *you*!

Do you have any idea when this could make it to a release on crates.io? I'm super stoked to start trying it out! ✨

KodrAus:
@yoshuawuyts Yeh this is an exciting step forward! I'll put together a release PR today and cc you on it. I might get a few other peeps from Libs to take a look at the diff too just to make sure we haven't missed anything.

PR #354, opened 2019-09-03
Switch from Travis/AppVeyor to GitHub Actions

alexcrichton:
Switches from running Windows builds in my personal AppVeyor with 1 concurrency to GitHub actions which has much more concurrency.


alexcrichton:
It's worth noting that the additions to CI here are:

* Rustfmt is run and verified to be clean
* Windows now runs the same test suites as Linux
* OSX was added as a test target

Other than that should be the same as before!

sfackler:
Worth caching the crates.io index?

alexcrichton:
Seems reasonable to me, but I don't think that actions has caching yet, so that may have to come later

PR #355, opened 2019-09-05
0.3.x: fix unknown path when possible

BusyJay:
Since 0.4..8, log can provide static string if it's available, so 0.3.x can utilize it to avoid annoying unknown string when they are using together.


KodrAus:
Thanks for backporting this @BusyJay, it doesn't look like we're supporting previous releases of `log` though so will close this one for now.

PR #359, opened 2019-10-12
Mark trait objects with 'dyn' (bare-trait-object)

Fenex:
The commit removes deprecated-type warnings.


Fenex:
Oh, support for to 1.16. Sry =\

PR #361, opened 2019-10-30
Reduce unsafe usage

tuxsoy:


src/lib.rs
```diff
@@ -1192,7 +1191,7 @@ pub fn set_max_level(level: LevelFilter) {
 /// [`set_max_level`]: fn.set_max_level.html
 #[inline(always)]
 pub fn max_level() -> LevelFilter {
-    unsafe { mem::transmute(MAX_LOG_LEVEL_FILTER.load(Ordering::Relaxed)) }
+    LevelFilter::from_usize(MAX_LOG_LEVEL_FILTER.load(Ordering::Relaxed)).unwrap()
```
sfackler: This will introduce a branch and unwind into every log location.

src/lib.rs
```diff
@@ -1210,7 +1209,7 @@ pub fn max_level() -> LevelFilter {
 /// [`set_logger`]: fn.set_logger.html
 #[cfg(all(feature = "std", atomic_cas))]
 pub fn set_boxed_logger(logger: Box<Log>) -> Result<(), SetLoggerError> {
-    set_logger_inner(|| unsafe { &*Box::into_raw(logger) })
+    set_logger_inner(|| logger.leak() )
```
sfackler: I don't believe this will compile as-is - it has to be called as `Box::leak(logger)`. It will also increase our minimum supported Rust version to 1.26.

tuxsoy: Is there a minimum Rust version that `log` and other official rust crates need to support?

sfackler: We currently support 1.16.0

PR #362, opened 2019-11-09
Use explicit version of cfg-if in msrv build

KodrAus:
Just to try fix up our CI.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,7 +45,7 @@ kv_unstable = []
 kv_unstable_sval = ["kv_unstable", "sval/fmt"]
 
 [dependencies]
-cfg-if = "=0.1.9"
+cfg-if = "0.1.2"
 serde = { version = "1.0", optional = true, default-features = false }
 sval = { version = "0.4.2", optional = true, default-features = false }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,7 +45,7 @@ kv_unstable = []
 kv_unstable_sval = ["kv_unstable", "sval/fmt"]
 
 [dependencies]
-cfg-if = "0.1.2"
+cfg-if = "=0.1.9"
 serde = { version = "1.0", optional = true, default-features = false }
 sval = { version = "0.4.2", optional = true, default-features = false }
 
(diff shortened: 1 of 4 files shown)


sfackler:
It wasn't worth pinning cfg-if?

KodrAus:
I didn't have a lot of success pinning `cfg-if` using the `cargo update` onn the `1.16` toolchain, but I didn't actually try setting a `=` bound in our own `Cargo.toml`. We didn't get a lot of mileage out of bumping our MSRV from `1.16` to `1.31` anyways, so I'll give that a go :+1:

KodrAus:
Ok, looks like that works out just fine!

Cargo.toml
```diff
@@ -45,7 +45,7 @@ kv_unstable = []
 kv_unstable_sval = ["kv_unstable", "sval/fmt"]
 
 [dependencies]
-cfg-if = "0.1.2"
+cfg-if = "=0.1.9"
```
sfackler: This will cause issues if log is used in a dependency graph with another crate with a non `=` dependency on cfg-if. Cargo will iirc just refuse to resolve versions. If we can't just `cargo update -p cfg-if --precise 0.1.9` in the MSRV build, it may just be best to bump our MSRV to whatever cfg-if uses.

KodrAus: Ah right. I’d assumed we’d possibly just end up compiling two versions of `cfg-if`.

I’ll look again at why the `cargo update` path didn’t work out for me and fall back to bumping back to `1.31`.

sfackler:
rustfmt's failing, but r=me otherwise. Realistically, no-one's going to do the downgrade dance with cfg-if anyway...

src/lib.rs
```diff
@@ -911,28 +911,17 @@ impl<'a> RecordBuilder<'a> {
...
-        return RecordBuilder {
-            record: Record {
-                args: format_args!(""),
-                metadata: Metadata::builder().build(),
-                module_path: None,
-                file: None,
-                line: None,
-            },
-        };
+            }
```
sfackler:
```suggestion
            },
```

Issue #372, opened 2019-12-16
Potential behaviour regression with named formatting arguments


kubkon:
Current release version `0.4.10` seems to have introduced a regression in behaviour in the sense that named formatting arguments, which were legal in the previous versions, are now illegal.

Minimal example:

```rust
fn main() {
    let name = "John";
    log::info!("Hello, {name}", name = name);
}
```

Is this intended behaviour or a bug?

kubkon:
NB, works with the previous release `0.4.8`.

bbqsrc:
This is quite broken. 0.4.8 works, newer version broken. Format args do not work and cause build errors.

estk:
@KodrAus or @sfackler  Maybe .10 should be yanked? I have the same issue.

maplant:
@estk agree that 0.4.10 should be yanked.

sfackler:
Yanked.

KodrAus:
Ok. It looks like we just haven’t built up the CI coverage yet to be able to make changes to the `log` macros without breaking things. I’ll add some more cases, revert our implementation back to what it was (but retain #366) and we can explore macro changes externally.

KodrAus:
`master` has been updated to use the old macro implementation and we've got a failing test on the `kv_macro` branch the new one lives on to track this. Any new releases we push out now will be using the same macros as `0.4.8`.

I'll go ahead and close this one now. Thanks for the report!

Issue #369, opened 2019-12-15
0.4.9 release seems to break tokio (which depends on log)


alex:
You can see an example of this here: https://github.com/alex/otp-cop/pull/374/checks?check_run_id=349720268

I didn't see a mention of breaking changes (besides MSRV) in the CHANGELOG, so I'm filing this here. Let me know if this should be filed on tokio instead though!

mojzu:
I am receiving similar errors trying to install cargo-make.

```
   Compiling cargo-make v0.24.1
error: macro expansion ignores token `if` and any following
   --> <::log::macros::log_impl macros>:13:23
    |
13  |       let lvl = $ lvl ; if lvl <= $ crate :: STATIC_MAX_LEVEL && lvl <= $ crate
    |                         ^^
    | 
   ::: /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/cargo-make-0.24.1/src/lib/cache.rs:106:47
    |
106 |   ...                   Err(error) => info!(
    |  _____________________________________-
107 | | ...                       "Error while writing to cache file: {:#?}, error: {:#?}",
108 | | ...                       &file_name, error
109 | | ...                   ),
    | |                       -
    | |                       |
    | |_______________________caused by the macro expansion here
    |                         help: you might be missing a semicolon here: `;`
    |
    = note: the usage of `log_impl!` is likely invalid in expression context
    = note: this error originates in a macro outside of the current crate (in Nightly builds, run with -Z external-macro-backtrace for more info)

...
```

hecrj:
It seems to break [`relevant`](https://github.com/omni-viral/relevant) and [`gfx-backend-vulkan`](https://github.com/gfx-rs/gfx/tree/master/src/backend/vulkan) too.

pythondude325:
This also breaks [`goblin`](https://github.com/m4b/goblin)

sfackler:
I've yanked the 0.4.9 release.

@KodrAus It looks like all of the code blocks in the log_impl macro need to be contained in a block - how did CI pass after the KV stuff landed?

mojzu:
Install cargo-make is working again, thank you for the very quick fix!

KodrAus:
@sfackler Hm that's a good question... I'll check this out now.

KodrAus:
Ok, it looks like with our macro tests being brand new we didn't have coverage for the macros in expression context so CI didn't pick this up. I also totally missed this when reviewing the original implementation.

[...]

PR #374, opened 2019-12-18
Revert the macro implementation

KodrAus:
In #369 and #372 we've run into a few regressions with the new macro implementation that supports key-value pairs. This PR shifts that implementation onto a new branch called [`kv_macro`](https://github.com/rust-lang/log/tree/kv_macro) so that we can work on it out-of-band.

cc @yoshuawuyts

r? @sfackler


tisonkun:
Any updates? It is a bit long from we support kvs in macro.

Issue #357, opened 2019-09-08
Logging std::error::Errors


theduke:
It would be really nice to have a specialized way of logging errors that implement `std::error::Error`. 

This would allow automatically capturing detailed information about an error, including the `Display` and `Debug` representations, and especially backtraces via the new `Error::backtrace`, which has an open PR.

The main purpose is to give the backend implementation the option to extract all desired information about an error and use it as appropriate for the output.

Not sure how this could fit into the current design, but I'd imagine an additional method on `Log` and a macro. 

```rust
trait Log {
  ...
  // The record serves the purpose of providing the location, and optionally an additional error message / metadata.
  // Question: Should the record be optional or required?
  fn error(&self, error: &dyn std::error::Error,  record: Option<&Record>);
}
```

Plus a macro. 

Naming for that macro would of course be somewhat awkward, considering `error!` exists.  Maybe `with_error!()`. Also there could be a variant of `error!` that allows specifying an error value.

One could of course ask if this is really necessary, since you could already write a custom macro or method that captures the required information as a `Record`. But I'd like to see a standardized way of doing this that gives the logger implementation the power to do what it wants, without forcing the user to make a decision here up-front.

KodrAus:
Personally, I think treating errors as first-class concepts in logs is a worthwhile thing to do. It looks like we might not have given this too much thought in the past, but now that the `std::error::Error` type is more useful and backtrace support is in the works it could be a good time to consider it.

We could add something like a `fn source(&self) -> Option<&(dyn Error + 'static)>` method to `Record` along with macro support similar to how `target` was added:

```rust
error!(source: err, "Something didn't work {}", err);
```

Alternatively we could design support into the key-values API and leave it up to convention to capture an error with a common key name. That would probably be my preferred approach, because it would avoid creating too many parallel syntaxes in the `log!` macros.

theduke:
I like the `source: err` syntax.

> We could add something like a fn source(&self) -> Option<&(dyn Error + 'static)> method to Record along with macro support similar to how target was added

I don't think that's feasible, since it would require the `Record` to take ownership of the error.
Often  you would want to log the error, but then forward it in a `Result<>`.

theduke:
I like the `source: err` syntax. That would also enable attaching an error to all the macros, and prevent clutter.

sfackler:
Yep, I definitely agree that it makes sense to allow Errors to be tacked onto log records - it will be a field that's only available with the `std` feature, but that should be fine.

[...]

PR #378, opened 2020-01-31
Add an error field to Record

KodrAus:
For #357 

Adds a `dyn Error + 'static` field to `Record` so that concrete errors can be associated with a log event. I'd thought about adding a `Send` requirement, but decided against it because `Record` already isn't `Send`, there are error types out there folks might want to log that aren't `Send`, and since you can downcast errors if your app needs to send log records to other threads you could use that to downcast and build your own sendable record type.

r? @sfackler 

# Future extension

We could extend the macro to allow something like this in caller code:

```rust
let o = do_some_work().map_err(|e| {
    error!(source: &e, target: "some target", "I tried so hard, and got so far");
    e
})?;
```

Macro support could be added in a fairly naive way to the log macros, or we could take another stab at a tt muncher to support arbitrary special fields before the arguments.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -728,6 +728,8 @@ pub struct Record<'a> {
     module_path: Option<MaybeStaticStr<'a>>,
     file: Option<MaybeStaticStr<'a>>,
     line: Option<u32>,
+    #[cfg(feature = "std")]
+    error: Option<&'a (dyn std::error::Error + 'static)>,
     #[cfg(feature = "kv_unstable")]
     key_values: KeyValues<'a>,
 }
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -816,6 +818,13 @@ impl<'a> Record<'a> {
         self.line
     }
 
+    /// The error associated with the message.
```
sfackler: This should mention it requires the std feature.

src/lib.rs
```diff
@@ -979,6 +992,14 @@ impl<'a> RecordBuilder<'a> {
         self
     }
 
+    /// Set [`error`](struct.Record.html#method.error)
```
sfackler: Same here

sfackler:
Seems reasonable to me to not have a Send bound - even if the fmt::Arguments field wasn't also there you can't really ship a Record around since the error is just borrowed.

I do think we should figure out what the macro change looks like before cutting a release with this, though.

KodrAus:
> I do think we should figure out what the macro change looks like before cutting a release with this, though.

Sure :+1: I might leave this PR sitting open then and switch over to #357 where we can work out how we want the macro to look. I don't think it'll be as risky to change the macros following the same scheme as `target` as the kv support was, because we're not messing with the token soup that gets forwarded to `format_args`.

KodrAus:
This has fallen pretty out of date now so I'll close it and revisit in the future.

PR #394, opened 2020-05-18
Avoid format_args expansion when logging is disabled

stevenroose:
Revisit of https://github.com/rust-lang/log/pull/297/files.

That PR got the following comments: 

> Wouldn't the removal of if lvl <= $crate::STATIC_MAX_LEVEL && lvl <= $crate::max_level() cause the build time removal of the log entries over a specific level to stop to work?

and 

> This imposes a nontrivial cost to all log events that don't have expensive arguments. Older versions of the log crate looked like this and we explicitly moved away from it under the idea that we should optimize for the common case of cheap arguments.

Though, it seems that currently even if a log is to be filtered out by the logger, the entire log message is constructed and allocated. Which doesn't seem "optimized for cheap arguments".

Also, I don't know about compiler internals, but the difference is that this check
```rust
```
is turned into this check
```rust
        lvl <= $crate::STATIC_MAX_LEVEL
            && lvl <= $crate::max_level()
            && $crate::__private_api_enabled(lvl, $target)
```
I think if the compiler can compile away code behind `if false {`, I suppose it can also compile away code behind `if false && ... {`, no?

The test provided currently fails as the argument is also expanded when the message is not logged.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,10 @@ harness = false
 name = "macros"
 harness = true
 
+[[test]]
+name = "expansion"
+harness = false
+
 [features]
 max_level_off   = []
 max_level_error = []
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -30,8 +30,9 @@
 #[macro_export(local_inner_macros)]
 macro_rules! log {
     (target: $target:expr, $lvl:expr, $message:expr) => ({
+        let target = $target;
         let lvl = $lvl;
-        if lvl <= $crate::STATIC_MAX_LEVEL && lvl <= $crate::max_level() {
+        if log_enabled!(target: target, lvl) {
             // ensure that $message is a valid format string literal
             let _ = __log_format_args!($message);
             $crate::__private_api_log_lit(
@@ -42,12 +43,13 @@ macro_rules! log {
         }
     });
     (target: $target:expr, $lvl:expr, $($arg:tt)+) => ({
+        let target = $target;
         let lvl = $lvl;
-        if lvl <= $crate::STATIC_MAX_LEVEL && lvl <= $crate::max_level() {
+        if log_enabled!(target: target, lvl) {
             $crate::__private_api_log(
                 __log_format_args!($($arg)+),
                 lvl,
-                &($target, __log_module_path!(), __log_file!(), __log_line!()),
+                &(target, __log_module_path!(), __log_file!(), __log_line!()),
             );
         }
     });
--- a/tests/expansion.rs
+++ b/tests/expansion.rs
@@ -1,0 +1,46 @@
+#[macro_use]
+extern crate log;
+
+use log::{Level, LevelFilter, Log, Metadata, Record};
+
+#[cfg(feature = "std")]
+use log::set_boxed_logger;
+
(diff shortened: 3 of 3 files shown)


stevenroose:
I see that this crate enforced rustfmt via CI but does not have a rustfmt config file provided.

sfackler:
> Though, it seems that currently even if a log is to be filtered out by the logger, the entire log message is constructed and allocated.

No, it does not allocate. It uses `format_args!`, not `format!`.

stevenroose:
> No, it does not allocate. It uses format_args!, not format!.

Hmm. Then what did I do wrong with my test? The method used in the args list is evaluated even when the log is not produced.

sfackler:
The arguments are evaluated just like any normal function call. That produces an [`Arguments`](https://doc.rust-lang.org/std/fmt/struct.Arguments.html) object which can then be formatted into a string or written to a file or whatever based on what the registered logging implementation wants to do.

The decision to not call `enabled` before logging was made intentionally in the 0.3 release. In many logging implementations, the logic needed for filtering needs to be duplicated in the log handling code for e.g. routing anyway, so calling enabled duplicates all of that work. Additionally, logging code can easily bitrot if arguments are not evaluated. Before that change was made, tons of debug logging statements in the standard library would panic with e.g. out of bounds array lookups if you ever tried enabling logging in the module.

Generally speaking, the vast majority of log calls don't contain expensive computation, and the few that do can manually wrap the call in a conditional.

stevenroose:
Hmm, I can hardly imagine that the enabled function will have a worst cost than allocating a bunch of objects based on variables. I did notice that the `SimpleLogger` example calls `enabled` inside its `log` method. Which seems weird. It feels that enabled should be called before.

Also, given the compile-static maximum log level, in most cases, neither the `enabled()` method or any of the argument expansions will be called with this change.

sfackler:
> Hmm, I can hardly imagine that the enabled function will have a worst cost than allocating a bunch of objects based on variables.

Please describe what "objects are being allocated based on variables" in this bit of code:

```rust
info!("shaving yak number {}: status {}", id, status);
```

> I did notice that the SimpleLogger example calls enabled inside its log method. Which seems weird. It feels that enabled should be called before.

> Note that enabled is not necessarily called before this method. Implementations of log should perform all necessary filtering internally.

https://docs.rs/log/0.4.8/log/trait.Log.html#tymethod.log

> Also, given the compile-static maximum log level, in most cases, neither the `enabled()` method or any of the argument expansions will be called with this change.

Static filtering is not enabled by default.

stevenroose:
> Note that enabled is not necessarily called before this method. Implementations of log should perform all necessary filtering internally.

Yeah I saw that. Just noting that that API could have been different.

> Static filtering is not enabled by default.

How so? Because one has to call `log::set_max_level()`?

KodrAus:
Just coming through some triage. It's been a while since we've had activity here. There is a decision to made between continuing to evaluate arguments when a static level precludes them and avoid potential breakage, and ignoring them to optimize things away. Both have upsides and downsides, but `log` has decided on the former for its macros. At this stage I don't think that's likely to change, but it's still possible to use `log` with alternative entrypoints, and those entrypoints may make different choices.

PR #396, opened 2020-05-22
Rework structured value casting

KodrAus:
This is a spike to see if we can improve the way we cast values to concrete types. It does more work when capturing `Debug` and `Display` to retain the underlying primitive so that a new macro could reasonably default to creating values through these std traits while retaining numbers.

It does some wacky build script code generation to create a static sorted set of type ids we care about so they can be binary searched quickly.


build.rs
```diff
@@ -3,12 +3,21 @@
...
     let target = env::var("TARGET").unwrap();
 
     if !target.starts_with("thumbv6") {
         println!("cargo:rustc-cfg=atomic_cas");
     }
 
+    #[cfg(feature = "kv_unstable")]
+    into_primitive::generate();
+
+    println!("cargo:rustc-cfg=src_build");
```
KodrAus, line 20: I’m using this so the `into_primitive` module can tell if it’s in the build script or not. Is there a cfg that exists already for this?

src/lib.rs
```diff
@@ -276,6 +276,9 @@
 #![cfg_attr(rustbuild, feature(staged_api, rustc_private))]
 #![cfg_attr(rustbuild, unstable(feature = "rustc_private", issue = "27812"))]
 
+#![feature(is_sorted)]
```
KodrAus, line 279: This seems nice for debug assertions. I’ll look into where stabilisation is at.

src/lib.rs
```diff
@@ -276,6 +276,9 @@
...
+#![feature(const_type_id)]
```
KodrAus, line 280: I think we just stabilise these const methods on-demand. I can’t see why this one shouldn’t be stabilised, but will look into it.

KodrAus: See: https://github.com/rust-lang/rust/pull/72488

src/lib.rs
KodrAus: See: https://github.com/rust-lang/rust/issues/53485

src/kv/value/internal/cast/into_primitive.rs
```diff
@@ -0,0 +1,109 @@
+// NOTE: With specialization we could potentially avoid this call using a blanket
```
KodrAus, line 1: Specialization is possible because we use `impl Trait` instead of `dyn Trait` for inputs. This approach would still be necessary if we switched to `dyn Trait`. The trade-off would be that with `dyn Trait` we could generate less code in calling libraries at the cost of still requiring this runtime check.

KodrAus: cc @hawkw sorry for the random ping but thought you might find this interesting for `tracing` 🙂

This is using a static set of type ids sorted at build time and binary searches them for interesting types given a generic `T: Debug + ‘static` input and treats the input as that concrete type. The runtime cost per argument is small (I measured ~9ns on my machine for a hit and ~5ns for a miss) but is not negligible. It could also use specialisation to remove that runtime cost, which I think might land as `min_specialization` sometime over the next year. Maybe you could consider it for something like the `#[instrument]` macro, to keep the nice compatible `Debug` bound, but also retain numbers and booleans as structured values?

hawkw: This is fascinating, thanks for pinging me!

KodrAus: So...

If https://github.com/rust-lang/rust/pull/72437 and https://github.com/rust-lang/rust/pull/72488 are both merged then you’d be able to do this at compile time without specialization :)

There’s an example in the `const_type_id` PR.

Cargo.toml
```diff
@@ -47,6 +47,7 @@ std = []
 # this will have a tighter MSRV before stabilization
 kv_unstable = []
 kv_unstable_sval = ["kv_unstable", "sval/fmt"]
+kv_unstable_const_primitive = []
```
KodrAus, line 50: Just testing different strategies for destructuring.

src/kv/value/internal/cast/into_primitive.rs
```diff
@@ -0,0 +1,159 @@
+#[cfg(all(src_build, feature = "kv_unstable_const_primitive"))]
+pub(super) fn into_primitive<'v, T: 'static>(value: &'v T) -> Option<crate::kv::value::internal::Primitive<'v>> {
+    use std::any::TypeId;
+
+    use crate::kv::value::internal::Primitive;
+    
+    trait ToPrimitive where Self: 'static {
```
KodrAus, line 7: So strangely enough this approach using const seems slightly slower than the runtime binary search. I haven’t looked into why that is yet.

KodrAus: Just a few missing inlines.

src/kv/value/internal/cast/primitive.rs
```diff
@@ -0,0 +1,250 @@
+// Use consts to match a type with a conversion fn
+// Pros: fast, will work on stable soon (possibly 1.45.0)
+// Cons: requires a `'static` bound
+#[cfg(all(src_build, feature = "kv_unstable_const_primitive"))]
```
KodrAus, line 4: @hawkw I’ve got a few different strategies for capturing the value from a generic `T` here now.

For `tracing`, if this is something you wanted to do you might end up needing to look more at the specialization one, because the others require a `’static` bound in order to get the type id, which from what I can see isn’t compatible with how the `#[instrument]` macro currently works.

src/lib.rs
```diff
@@ -275,6 +275,8 @@
 // an unstable crate
 #![cfg_attr(rustbuild, feature(staged_api, rustc_private))]
 #![cfg_attr(rustbuild, unstable(feature = "rustc_private", issue = "27812"))]
+// TODO: Remove once https://github.com/rust-lang/rust/pull/72488 is merged
```
KodrAus, line 278: I'll wait for https://github.com/rust-lang/rust/pull/72488 before merging this PR so this unstable feature can be removed.

KodrAus:
Once a new `nightly` goes out this should be good to merge!

PR #400, opened 2020-06-04
Respect fmt trait and flags in Value

### LINKED ISSUE #395: kv::value::Value constructed from String uses Debug output for Display
@isgj reported an issue in Tide https://github.com/http-rs/tide/pull/495#discussion_r426254201 that led us to file in issue in Femme https://github.com/lrlna/femme/issues/12 about output that was being printed with extra quotes:

```json
{"level":30,"time":1589717147052,"msg":"Response sent","method":"\"GET\"","status":"200","duration":"\"157ms\"","path":"\"/api/v1/users\""}
```

The printing logic in `femme` looks like this:

```rust
    impl<'kvs, 'a> kv::Visitor<'kvs> for Visitor<'a> {
        fn visit_pair(
            &mut self,
            key: kv::Key<'kvs>,
            val: kv::Value<'kvs>,
        ) -> Result<(), kv::Error> {
            write!(self.string, "\n    {} {}", style(key).bold(), val).unwrap();
            Ok(())
        }
```

It turns out that the problem stems from the `Display` path using the `Debug` logic: https://docs.rs/log/0.4.8/src/log/kv/value/internal.rs.html#111-117. Currently in master this code seems refactored, but still inherently the same: https://github.com/rust-lang/log/blob/master/src/kv/value/internal/mod.rs#L49-L51.

We should probably introduce a new `Display` path for values so that they can correctly be displayed. Thanks!

cc/ @KodrAus

KodrAus:
Fixes #395 

This PR makes the `fmt::Debug` and `fmt::Display` impls on `Value` respect the trait being used and any flags present. I've added a separate visitor for `Display` that uses the right trait, and pass the value directly to the formatter instead of going through `format_args`.

There are some extra tests for formatting and I've shifted a few pieces around to be consistent. ~~Tests won't pass until I fix https://github.com/sval-rs/sval/issues/84 as well.~~

r? @yoshuawuyts

### DIFF
--- a/src/kv/error.rs
+++ b/src/kv/error.rs
@@ -67,10 +67,4 @@ mod std_support {
             Error::boxed(err)
         }
     }
-
-    impl From<Error> for io::Error {
-        fn from(err: Error) -> Self {
-            io::Error::new(io::ErrorKind::Other, err)
-        }
-    }
 }
--- a/src/kv/value/internal/sval.rs
+++ b/src/kv/value/internal/sval.rs
@@ -42,6 +42,48 @@ impl<'s, 'f> Slot<'s, 'f> {
 
 impl<'v> sval::Value for kv::Value<'v> {
     fn stream(&self, s: &mut sval::value::Stream) -> sval::value::Result {
+        struct SvalVisitor<'a, 'b: 'a>(&'a mut sval::value::Stream<'b>);
+
+        impl<'a, 'b: 'a, 'v> Visitor<'v> for SvalVisitor<'a, 'b> {
+            fn debug(&mut self, v: &dyn fmt::Debug) -> Result<(), Error> {
+                self.0
+                    .fmt(format_args!("{:?}", v))
+                    .map_err(Error::from_sval)
+            }
+
+            fn u64(&mut self, v: u64) -> Result<(), Error> {
+                self.0.u64(v).map_err(Error::from_sval)
+            }
+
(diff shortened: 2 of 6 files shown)


src/kv/value/internal/sval.rs
```diff
@@ -191,4 +191,20 @@ mod tests {
...
+
+        impl sval::Value for TestSval {
+            fn stream(&self, stream: &mut sval::value::Stream) -> sval::value::Result {
+                stream.u64(42)
+            }
+        }
+
+        assert_eq!(
+            format!("{:04?}", 42u64),
+            format!("{:04?}", kv::Value::from_sval(&TestSval)),
```
KodrAus, line 207: This doesn't pass yet.

src/kv/value/internal/sval.rs
```diff
@@ -107,48 +149,6 @@ impl Error {
     }
 }
 
-struct SvalVisitor<'a, 'b: 'a>(&'a mut sval::value::Stream<'b>);
```
KodrAus, line 110: Just moved up into the `impl` block, like the visitors in the `fmt` module.

src/kv/error.rs
```diff
@@ -67,10 +67,4 @@ mod std_support {
             Error::boxed(err)
         }
     }
-
-    impl From<Error> for io::Error {
```
KodrAus, line 71: We removed a similar external `From` impl in #398 that was causing problems with inference.

PR #405, opened 2020-07-27
Allow Level and LevelFilter to deserialize from the index_variant

diondokter:
This PR allows the Level and LevelFilter enums to deserialize from the index_variant.

This solves https://github.com/jamesmunns/postcard/issues/24.

Postcard is made to serialize into a compact byte representation, so it takes the index_variant that is provided by the [Serialize implementation](https://github.com/rust-lang/log/blob/master/src/serde.rs#L23). There was however no way to change that back into the enum.

### DIFF
--- a/src/serde.rs
+++ b/src/serde.rs
@@ -60,6 +60,17 @@ impl<'de> Deserialize<'de> for Level {
 
                 self.visit_str(variant)
             }
+
+            fn visit_u64<E>(self, v: u64) -> Result<Self::Value, E>
+            where
+                E: Error,
+            {
+                let variant: &str = LOG_LEVEL_NAMES[1..]
+                    .get(v as usize)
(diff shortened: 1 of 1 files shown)


src/serde.rs
```diff
@@ -60,6 +60,17 @@ impl<'de> Deserialize<'de> for Level {
 
                 self.visit_str(variant)
             }
+
+            fn visit_u32<E>(self, v: u32) -> Result<Self::Value, E>
```
KodrAus, line 64: It looks like `serde` uses `visit_u64` for this, which makes sense, since `visit_u32` forwards to it by default, so let's target `visit_u64` instead.

KodrAus: Ah, I see your second commit changes this to `visit_u32`. I think sticking with the more compatible `visit_u64` like you originally had is a good idea :+1:

diondokter: Done

PR #409, opened 2020-08-23
Add Level::as_str

Thomasdezeeuw:
Returns the string representation of the `Level`.


Thomasdezeeuw:
If acceptable, maybe we also want to add it to `LevelFilter`.

sfackler:
LGTM, but let's add it to `LevelFilter` as well!

Thomasdezeeuw:
@sfackler done.

sfackler:
Thanks!

PR #410, opened 2020-08-27
Use `std::panic::Location` for file & line information

pwoolcoc:
this commit adds a feature, `track_caller`, that allows the user to opt-in to using location information from `std::panic::Location` instead of the built-in `file!` and `line!` macros. This will allow log record location information to be manipulated in the same way that `#[track_caller]` allows panic location information to be manipulated

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,6 +48,9 @@ std = []
 kv_unstable = []
 kv_unstable_sval = ["kv_unstable", "sval/fmt"]
 
+# requires 1.46.0
+track_caller = []
+
 [dependencies]
 cfg-if = "0.1.2"
 serde = { version = "1.0", optional = true, default-features = false }
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -249,9 +249,12 @@ macro_rules! __log_module_path {
 #[macro_export]
 macro_rules! __log_file {
     () => {
-        if cfg!(feature = "track_caller") {
+        #[cfg(feature = "track_caller")]
+        {
             ::std::panic::Location::caller().file()
-        } else {
+        }
+        #[cfg(not(feature = "track_caller"))]
+        {
             file!()
         }
     };
@@ -261,9 +264,12 @@ macro_rules! __log_file {
 #[macro_export]
 macro_rules! __log_line {
     () => {
-        if cfg!(feature = "track_caller") {
+        #[cfg(feature = "track_caller")]
+        {
             ::std::panic::Location::caller().line()
-        } else {
+        }
+        #[cfg(not(feature = "track_caller"))]
+        {
             line!()
         }
     };
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -249,7 +249,11 @@ macro_rules! __log_module_path {
 #[macro_export]
 macro_rules! __log_file {
     () => {
-        file!()
+        if cfg!(feature = "track_caller") {
+            ::std::panic::Location::caller().file()
+        } else {
+            file!()
+        }
     };
 }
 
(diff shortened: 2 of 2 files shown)


Thomasdezeeuw:
Is there a benefit to doing this?

Thomasdezeeuw:
Ah, sorry didn't see the second part of your comment, ignore my previous comment.

pwoolcoc:
It allows you to manipulate location information for log messages in the same way that you can for `panic!` messages. For example, one of my applications has a generic "retry-er" that takes a closure and a retry policy, and will run the closure until it gets a successful return value from the closure or until it hits it's max retries. Currently, any logging we do from the retryer (like "got error Foo, retry 1/10") will include location information from inside the retryer, but no identifying information about exactly which closure generated the error. Of course, we could simply pass more information to the retryer to help us identify the piece of code that is erroring out, but I thought it would be possible to use `#[track_caller]` for this instead, to keep the API cleaner. Turns out I couldn't use `log::error!`, `log::warn!`, etc, because those macros use `file!()` and `line!()`. I can do it if I write my own `log::Record`-generating code, but I thought having this functionality built-in to `log` might be useful for others.

src/macros.rs
```diff
@@ -249,14 +249,22 @@ macro_rules! __log_module_path {
 #[macro_export]
 macro_rules! __log_file {
     () => {
-        file!()
+        if cfg!(feature = "track_caller") {
+            ::std::panic::Location::caller().file()
+        } else {
+            file!()
+        }
```
jdm, line 256: This will break compilation of log with rustc < 1.46, even without the track_caller feature enabled. The following will allow compilation to succeed without that feature enabled:
```
        #[cfg(feature = "track_caller")]
        {
            ::std::panic::Location::caller().file()
        }
        #[cfg(not(feature = "track_caller"))]
        {
            file!()
        }
```

src/macros.rs
```diff
@@ -249,14 +249,28 @@ macro_rules! __log_module_path {
 #[macro_export]
 macro_rules! __log_file {
     () => {
-        file!()
+        #[cfg(feature = "track_caller")]
+        {
+            ::std::panic::Location::caller().file()
```
sfackler, line 254: This cfg is evaluated in the context of the crate that invokes a log macro, not the context of the log crate so this won't work.

pwoolcoc: thanks @sfackler, I pushed an update so they're evaluated in the correct context

jdm:
My earlier concern has been addressed, but otherwise I'm just a bystander :)

pwoolcoc:
no problem, thanks!

pwoolcoc:
bump, can someone take a look at this?

Cargo.toml
```diff
@@ -48,6 +48,9 @@ std = []
 kv_unstable = []
 kv_unstable_sval = ["kv_unstable", "sval/fmt"]
 
+# requires 1.46.0
+track_caller = []
```
KodrAus, line 52: I think we could detect `1.46.0` in our `build.rs` and quietly enable this.

KodrAus (commented):
Thanks @pwoolcoc! If MSRVs weren't a concern this is probably something we'd want unconditionally, wouldn't we? If that's the case I think we could use our `build.rs` to detect `1.46.0` support and just enable it quietly instead of using an explicit feature.

KodrAus:
Just coming in through some triage. I think the only outstanding thing here was using our build script to automatically use `track_caller` if we're on at least `1.46.0`.

jrose-signal:
Since this PR was left off, the MSRV has been bumped to 1.60. I think that means this can/should be turned on unconditionally!

Thomasdezeeuw:
@pwoolcoc is this something you're still interested in? We've updated our MSRV to 1.60 so this can be done now.

pwoolcoc:
@Thomasdezeeuw yes, but as I deleted my original fork, I can't seem to update this PR, so I'll have to open a new PR.

pwoolcoc:
Closing in favor of https://github.com/rust-lang/log/pull/599

PR #413, opened 2020-09-12
atomics: Add support for targets without atomics

alistair23:
Some targets (such as RV32IMC) don't have atomic support. This PR allows the log crate to build for those targets.

As the platform doesn't have atomics we can't implement an Atomic version of AtomicUsize, so this version is not atomic. Any platform without atomics is unlikely to have multiple cores, so this shouldn't be a problem.
    
This is somewhat based on implementations in: https://github.com/japaric/heapless

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -20,10 +20,16 @@ fn main() {
         None => return,
     };
 
-    // If the target isn't thumbv6 then we can use atomic CAS
-    if !target.starts_with("thumbv6") {
-        println!("cargo:rustc-cfg=atomic_cas");
-    }
+    match &target[..] {
+        "thumbv6m-none-eabi"
+        | "msp430-none-elf"
+        | "riscv32i-unknown-none-elf"
+        | "riscv32imc-unknown-none-elf" => {}
+
+        _ => {
+            println!("cargo:rustc-cfg=atomic_cas");
+        }
+    };
 
     // If the Rust version is at least 1.46.0 then we can use type ids at compile time
     if minor >= 47 {
--- a/build.rs
+++ b/build.rs
@@ -31,6 +31,14 @@ fn main() {
         }
     };
 
+    match &target[..] {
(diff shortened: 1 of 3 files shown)


alistair23:
Ping

sfackler:
This seems like a plausible approach, though I think we'll need to have CI target at least one of these architectures so the not(has_atomics) code doesn't rot.

alistair23:
Done, I have added a CI test to build RISC-V without atomics.

sfackler:
Thanks!

cc @KodrAus does this seem okay to you as well? It's a bit scary, but it seems reasonable that a platform with no atomics can't really have threads either.

dtolnay (changes requested):
I would like the rationale from the PR description to be documented on the unsafe impl.

```rust
// Any platform without atomics is unlikely to have multiple cores, so
// writing via Cell will not be a race condition.
```

build.rs
```diff
@@ -20,10 +20,24 @@ fn main() {
         None => return,
     };
 
-    // If the target isn't thumbv6 then we can use atomic CAS
-    if !target.starts_with("thumbv6") {
-        println!("cargo:rustc-cfg=atomic_cas");
-    }
+    match &target[..] {
```
KodrAus, line 23: Maybe we could extract these into little functions so they're a bit easier to follow:

```rust
if target_has_atomic_cas { .. }

..

if target_has_atomics() { .. }
```

KodrAus:
I think I'd feel just a bit safer if we encapsulated the global logger code into a module, so that we limit the reachability of our "fake" `AtomicUsize` as much as possible, but I don't think that's a blocker.

alistair23:
I think I have addressed all comments, let me know if there is anything else.

sfackler:
Thanks!

alistair23:
Thanks for merging. Do you know when the next release will be?

PR #417, opened 2020-10-12
Update to cfg-if 1.0

mbrubeck:


jyn514:
Oops, I started working on this without noticing your PR. In mine I allowed using either 0.1 or 1.0 to avoid duplicate dependencies - do you mind doing that here?

```diff
$ git show
commit 610df5155f906f00e446c9ad2f7a02041869f68c (HEAD -> cfg-if)
Author: Joshua Nelson <jyn514@gmail.com>
Date:   Wed Oct 14 10:18:06 2020 -0400

    Update cfg-if to 1.0
    
    There are many crates the depend on log, and if they depend on `cfg-if 1.0`, depending on 0.1 will introduce a duplicate dependency. This
    allows removing the duplication.
    
    To be even more flexible, since `log` compiles with both cfg-if 0.1 and 1.0,
    this allows compiling with either, avoiding more duplication for crates
    which have not yet upgraded to 1.0.

diff --git a/Cargo.toml b/Cargo.toml
index 4021fb0..e62d7e8 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,7 +49,7 @@ kv_unstable = []
 kv_unstable_sval = ["kv_unstable", "sval/fmt"]
 
 [dependencies]
-cfg-if = "0.1.2"
+cfg-if = ">=0.1.2, ~1"
 serde = { version = "1.0", optional = true, default-features = false }
 sval = { version = "0.5.2", optional = true, default-features = false }
```

mbrubeck:
> ```diff
> +cfg-if = ">=0.1.2, ~1"
> ```

This doesn't work as intended.  When there are multiple requirements, Cargo will only allow versions that meet *all* of the requirements.  So this is compatible with 1.0.0 (which matches both parts), but not with 0.1.2 (which matches only the first part):

```text
$ cargo update -p cfg-if --precise 0.1.2
    Updating crates.io index
error: failed to select a version for the requirement `cfg-if = ">=0.1.2, ~1"`
candidate versions found which didn't match: 0.1.2
```

(Even if Cargo used “or” instead of “and” to combine the requirements, this would still be incorrect, because then it would allow versions like 2.0.0, which matches the first requirement but not the second.)

jyn514:
Oh, I see. It looks like cargo doesn't allow `or` which is unfortunate.

> (Even if Cargo used “or” instead of “and” to combine the requirements, this would still be incorrect, because then it would allow versions like 2.0.0, which matches the first requirement but not the second.)

Sure, you could fix that with `(>=0.1.2 AND < 1) OR ~1` if cargo had a syntax for it.

KodrAus (approved):
I'm ok with this as-is. It is a little unfortunate to have 2 versions compiling in a depdency graph, but that's mostly a point-in-time issue since `cfg-if` has been around so long and is so widely used. The best way to fix it is to upgrade crates to `cfg-if` `1.0` (it looks like there was only one small patch difference https://github.com/alexcrichton/cfg-if/pull/37)

emilio:
@KodrAus do you know when this will be released? Thanks!

PR #418, opened 2020-10-14
Remove use of unstable const_type_id feature (fixes CI failures on 1.47+)

mbrubeck:
The `const_type_id` feature was planned to stabilize in Rust 1.47, but this was reverted in rust-lang/rust#77083.

`build.rs` currently assumes the feature is stable in Rust 1.47 and later. This now causes errors when building `log` with the `kv_unstable` feature on beta and nightly, which in turn is causing CI to fail on master.

This patch simply removes the code that uses this feature, since it is not clear when it might be stabilized.  Alternatively, the code could be kept around behind a second, nightly-only Cargo feature if it is still useful to someone.  CC @KodrAus who added this code in #396.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -2,19 +2,13 @@
 //! atomics and sets `cfg` flags accordingly.
 
 use std::env;
-use std::process::Command;
-use std::str::{self, FromStr};
+use std::str;
 
 #[cfg(feature = "kv_unstable")]
 #[path = "src/kv/value/internal/cast/primitive.rs"]
 mod primitive;
 
 fn main() {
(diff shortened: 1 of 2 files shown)


KodrAus:
Ah! My bad 😄 I saw your message come through my notifications but thought it was a comment rather than a PR so I just pushed through #419 to fix this.

I have an in-progress branch that refactors the guts of the `value` module with an external crate that handles this all in a cleaner way so didn't go as far as actually removing the internal usage that you've tracked down.

Thanks for patching this up! :bow:

PR #419, opened 2020-10-15
remove dependency on reverted const type id

KodrAus:
We were going to depend on `const_type_id`, but that feature was reverted before it was stabilized.

I'm working on a refactoring that changes this, but in the meantime anybody depending on `kv_unstable` and building on `1.47.0` is broken. This just removes the expectation that `1.47.0` will allow `const_type_id`.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -10,11 +10,6 @@ use std::str::{self, FromStr};
 mod primitive;
 
 fn main() {
-    let minor = match rustc_minor_version() {
-        Some(minor) => minor,
-        None => return,
-    };
-
     let target = match rustc_target() {
         Some(target) => target,
         None => return,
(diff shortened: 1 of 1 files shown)


Karrq:
Is this perhaps the expected error to find on 1.47?
I'm currently trying to investigate who enables the kv feature in my workspace

```
error[E0635]: unknown feature `const_type_id`
 --> /home/user/.cargo/registry/src/github.com-1ecc6299db9ec823/value-bag-1.0.0-alpha.5/src/lib.rs:3:54
  |
3 | #![cfg_attr(value_bag_capture_const_type_id, feature(const_type_id))]
  |                                                      ^^^^^^^^^^^^^

error: aborting due to previous error

For more information about this error, try `rustc --explain E0635`.
error: could not compile `value-bag`.
```

PR #426, opened 2020-12-21
Key-value serializer adapters

KodrAus:
This PR does two things:

- Exposes the `Value::downcast_ref` method that was missed in #423 
- Adds adapters so that `Source`s can be treated like a map or sequence for formatting or serialization

This means `Log` implementors can now write something like:

```rust
use log::kv::source;

#[derive(Serialize)]
pub struct MyRecordAsMap<'a> {
    msg: &'a str,
    #[serde(flatten)]
    #[serde(with = "source::as_map")]
    kvs: &'a dyn Source,
}
```

to define a type that can be serialized using `serde` as a map with key values flattened.


passcod:
Would be nice to add that code sample to the kv docs as it's not super obvious how to use it otherwise

PR #427, opened 2020-12-21
Run tests in our MSRV

KodrAus:
This is mostly important for our integration tests that exercise the macros. If our main crate tests don't build on the MSRV I'll tweak it to only run integration tests.


KodrAus:
Note: I'll merge this with a squash to remove that merge commit

PR #364, opened 2019-11-17
Some small changes for readability and performance

rodoufu:
The only problem on the build is on the dependency `cfg-if 0.1.10`

PR #428, opened 2020-12-21
Some small changes for readability

rodoufu:
Related to: #364

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "log"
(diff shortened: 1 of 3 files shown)


KodrAus:
Hi @rodoufu :wave: Thanks for the PR and sorry for letting this sit for so long! I think there are some good changes in here, but we try to keep `log` dependency-free by default so can't really bring in something like `rustversion`.

I have let this go very stale on you though, so I'll go ahead and close this one, but if you (or anybody) wanted to do this without the `TryFrom` impl that would be a welcome addition.

PR #431, opened 2021-01-05
Add a prelude module

ibraheemdev:
This PR adds a `prelude` module that makes it easier to import the logging macros. You can now glob import `prelude` as opposed to naming the macros explicitly:
```rust
use log::prelude::*;

// instead of
use log::{warn, trace, debug, log, info, error}
```

The current workarounds are to use the old `extern crate` syntax:
```rust
#[macro_use]
extern crate log;
```

Or to use them without an import:
```rust
log::warn!("...");
```

However, many people prefer not having to write the longer name, and a `prelude` module makes this easier:
```rust
use log::prelude::*;

warn!("...");
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -297,7 +297,7 @@ mod serde;
 pub mod kv;
 
 pub mod prelude {
-    pub use macros::{log, debug, error, info, trace, warn};
+    pub use crate::{log, debug, error, info, trace, warn};
 }
 
 #[cfg(has_atomics)]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -296,6 +296,10 @@ mod serde;
 #[cfg(feature = "kv_unstable")]
 pub mod kv;
 
+pub mod prelude {
+    pub use macros::{log, debug, error, info, trace, warn};
+}
+
 #[cfg(has_atomics)]
 use std::sync::atomic::{AtomicUsize, Ordering};
 


LunarLambda:
I want to add to this that it can also help with `pub use`ing the macros in a crate implenenting the log facade, which *may* help avoid confusion for new users (needing both `log` and whatever crate actually does the logging), and prevent potential version conflicts

BurntSushi:
I think adding a prelude to this crate is not a great idea. Preludes encourage glob imports which, generally speaking, may make code easier to write but typically makes it harder to read. While preludes can sometimes rise above this and become worth it, I think it's a very high bar and I don't think it's passed here.

dekellum:
You could consider using [tao-log][https://docs.rs/tao-log/1.0.1/tao_log/], or doing something similar yourself, either directly in a module of the project where you want it, _or_ in your own crate:

See this first few examples about glob importing/or not.

dekellum:
BTW, the "wrapper crate" ("new crate" pattern?) is going to likely be a lot easier way to make progress improving the "log ecosystem".  

That said, I don't want to discourage you from (also?) pursuing it here,  given the above "rust libs team" effective-veto comment. Best of luck, either way! :1st_place_medal: :heart: :shamrock:

LunarLambda:
> You could consider using [`tao-log`](https://docs.rs/tao-log/1.0.1/tao_log/), or doing something similar yourself, either directly in a module of the project where you want it, _or_ in your own crate:
> 
> See this first few examples about glob importing/or not.

This is what I did in my own crate, `outlog`, but I forgot to re-export the `log` and `log_enabled` macros.

There may have been some confusion about this issue. We meant specifically a macro prelude for `log`, i.e. a `use`able alternative to `#[macro_use] extern crate log;`

dekellum:
I think I understand that. Good luck!

LunarLambda:
> BTW, the "wrapper crate" ("new crate" pattern?) is going to likely be a lot easier way to make progress improving the "log ecosystem".
> 
> That said, I don't want to discourage you from (also?) pursuing it here, given the above "rust libs team" effective-veto comment. Best of luck, either way! 🥇 ❤️ ☘️

I don't think I'm familiar with that pattern. Is there somewhere I can read more about it?

dekellum:
Well, I just made it up! :smile:  But its very clean way to go, particularly for macro only crates.

PR #434, opened 2021-01-08
check std feature against kv_unstable

### LINKED ISSUE #433: no method named `to_str` found for struct `ValueBag<'v>` in the current scope
after upgrade to log 0.4.12

i don't know if related to https://github.com/sval-rs/value-bag/issues/15
```
error[E0599]: no method named `to_str` found for struct `ValueBag<'v>` in the current scope
   --> /home/fede/.cargo/registry/src/github.com-1ecc6299db9ec823/log-0.4.12/src/kv/value.rs:460:24
    |
460 |             self.inner.to_str()
    |                        ^^^^^^ method not found in `ValueBag<'v>`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0599`.
error: could not compile `log`

```

KodrAus:
Closes #433 

There's an incorrect `#[cfg(feature = "std")]` in `kv::value` that's broken users upgrading to `0.4.12`. This PR fixes that up and runs a few more permutations in CI.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -42,6 +42,8 @@ jobs:
     - run: cargo test --verbose --features kv_unstable
     - run: cargo test --verbose --features "kv_unstable std"
     - run: cargo test --verbose --features "kv_unstable_sval"
+    - run: cargo test --verbose --features "kv_unstable_sval std"
+    - run: cargo test --verbose --features "kv_unstable kv_unstable_std kv_unstable_sval kv_unstable_serde"
     - run: cargo run --verbose --manifest-path test_max_level_features/Cargo.toml
     - run: cargo run --verbose --manifest-path test_max_level_features/Cargo.toml --release
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -40,9 +40,8 @@ jobs:
     - run: cargo test --verbose --features serde
     - run: cargo test --verbose --features std
     - run: cargo test --verbose --features kv_unstable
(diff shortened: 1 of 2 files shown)


KodrAus:
It looks like this slipped through because the feature mismatch was masked by dev dependencies.

Byron:
Thanks a lot for the quick fix! I was just investigating why `smol` fails to compile.

KodrAus:
Once CI goes through I'll publish a `0.4.13` version. In the meantime `0.4.12` has been yanked.

Byron:
Thanks so much - another `cargo update` later and the world is fixed :).

Byron:
Thinking about it - in case you believe this could possibly happen again, then [`cargo all-features`](https://github.com/frewsxcv/cargo-all-features) might be the tool to enforce testing all permutations.

KodrAus:
It's definitely something worth looking into! We do something like this over in `env_logger`. One of the challenges here is that the dependencies whose features we were misusing also appear as `dev-dependencies`, which is why permutations of feature flags in tests didn't pick this up. So now we check them in `nightly` using the unstable `avoid-dev-deps` flag.

Byron:
Even though I don't truly follow 😅 I would like to point out that the tool can be configured with [options](https://github.com/frewsxcv/cargo-all-features#options) to handle special cases better. It seems this feature was entirely contributed, so I would hope that other needs that `log` might have could be contributed there as well.

KodrAus:
Ah great I'll take a look! If `cargo-check-all-features` passes through our additional `-Z avoid-dev-deps` args then that should be everything we need.

PR #438, opened 2021-01-08
Implement ToValue for String

Thomasdezeeuw:
Using the `str` implementation.


Thomasdezeeuw:
Nevermind I now see this has been moved behind the `kv_unstable_std` flag.

PR #439, opened 2021-01-08
Remove usage of AtomicUsize::compare_exchange

Thomasdezeeuw:
It is going to be deprecated in 1.50, but it's already producing
warnings on nightly.


KodrAus (approved):
Thanks @Thomasdezeeuw!

PR #446, opened 2021-01-23
Remove the `__private_api_log_lit` special case

### LINKED ISSUE #445: `#![feature(format_args_capture)]` support
It would be great if this crate supports [`#![feature(format_args_capture)]`](https://github.com/rust-lang/rust/issues/67984). It currently does not ([playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=9237562522889e215864f9df9ea42c19)):
```rust
#![feature(format_args_capture)]

fn main() {
    env_logger::init();
    let x = 0;
    log::error!("{x}");
}
```
The above program logs `{x}` *without `unused-variables` warning* because `log!` treats single argument specially and uses `__log_format_args!` to validate the argument:
https://github.com/rust-lang/log/blob/1cd39c01fd991e30a7c6eb4ae2b89d9d8af268db/src/macros.rs#L35-L36

I don't understand why `log!` treats single argument specially. The `__private_api_log` code path and `__private_api_log_lit` path should do the same thing (both of them construct `fmt::Arguments`):
https://github.com/rust-lang/log/blob/1cd39c01fd991e30a7c6eb4ae2b89d9d8af268db/src/lib.rs#L1481
https://github.com/rust-lang/log/blob/1cd39c01fd991e30a7c6eb4ae2b89d9d8af268db/src/lib.rs#L1462

If the (redundant?) single argument special treatment is removed, `#![feature(format_args_capture)]` will be supported. That is not a breaking change because `log!` already uses `__log_format_args!` to validate the argument.

I'd like to help if this request is accepted.

hyd-dev:
https://github.com/rust-lang/log/issues/445#issuecomment-765965279:
> That special case existed to minimize the amount of code generated, but it's not compatible with format_args_capture as you pointed out. It should just be removed.

Closes #445.

I did not add a test for `format_args_capture`. I think I need to create a custom logger to capture records and add a testing-only crate feature or `--cfg` (to tell if `#![feature(format_args_capture)]` is allowed to use) to test it. `format_args_capture` is *unstable* now, I'm not sure if it worths it. Feel free to ask me to add such test if it's necessary.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1471,25 +1471,6 @@ pub fn __private_api_log(
 
 // WARNING: this is not part of the crate's public API and is subject to change at any time
 #[doc(hidden)]
-pub fn __private_api_log_lit(
-    message: &str,
-    level: Level,
-    &(target, module_path, file, line): &(&str, &'static str, &'static str, u32),
-) {
-    logger().log(
-        &Record::builder()
-            .args(format_args!("{}", message))
-            .level(level)
-            .target(target)
-            .module_path_static(Some(module_path))
-            .file_static(Some(file))
-            .line(Some(line))
-            .build(),
-    );
-}
-
-// WARNING: this is not part of the crate's public API and is subject to change at any time
-#[doc(hidden)]
 pub fn __private_api_enabled(level: Level, target: &str) -> bool {
     logger().enabled(&Metadata::builder().level(level).target(target).build())
 }
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -29,18 +29,6 @@
 /// ```
 #[macro_export(local_inner_macros)]
 macro_rules! log {
-    (target: $target:expr, $lvl:expr, $message:expr) => ({
-        let lvl = $lvl;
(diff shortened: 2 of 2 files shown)


hyd-dev:
The [CI failure](https://github.com/rust-lang/log/runs/1754550362#step:4:5) looks like rust-lang/cargo#8517:
```
    Updating crates.io index
warning: spurious network error (2 tries remaining): error inflating zlib stream; class=Zlib (5)
warning: spurious network error (1 tries remaining): error inflating zlib stream; class=Zlib (5)
error: failed to get `cfg-if` as a dependency of package `log v0.4.13 (D:\a\log\log)`

Caused by:
  failed to load source for dependency `cfg-if`

Caused by:
  Unable to update registry `https://github.com/rust-lang/crates.io-index`

Caused by:
  failed to fetch `https://github.com/rust-lang/crates.io-index`

Caused by:
  error inflating zlib stream; class=Zlib (5)
Error: Process completed with exit code 1.
```

PR #449, opened 2021-03-25
Allow use of deprecated atomic::spin_loop_hint

Thomasdezeeuw:
It's replaced by `hint::spin_loop`, however that has only been stable
since 1.49.0. Since we still support 1.31.0 we can't use it. Once the
MSRV is updated to 1.49 we can use `hint::spin_loop`.


KodrAus (approved):

KodrAus:
Thanks @Thomasdezeeuw!

PR #450, opened 2021-04-01
Add `unset_logger`, `unset_boxed_logger` and `reset_max_level` functions

pablosichert:
Hey there!

Specifically for testing, it would be handy if the globals could be restored to their initial state.

This PR adds an `unset_logger`, `unset_boxed_logger` and `reset_max_level` function to accomplish that.

---

Btw. I was wondering what the purpose of the `test_max_level_features` folder is, since the `tests/filters.rs` file seems to accomplish the same.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1435,16 +1435,31 @@ impl error::Error for SetLoggerError {}
 /// Unsets the global logger.
 ///
 /// When the global logger has been set using [`set_boxed_logger`], no
-/// destructors will be called and memory of the pervious logger will leak.
+/// destructors will be called and memory of the previous logger will leak. To
+/// prevent leaking memory in that case, use [`unset_boxed_logger`] instead.
 ///
 /// This function is typically only used in tests to reset global state.
 #[cfg(atomic_cas)]
 pub fn unset_logger() {
-    unset_logger_inner()
+    unset_logger_inner(false)
+}
+
+/// Unsets the global logger.
+///
+/// This function is typically only used in tests to reset global state.
+///
+/// # Safety
+///
+/// When the logger has been initialized before calling this function, it is
+/// only safe to call this function if the logger has been initialized via
+/// [`set_boxed_logger`].
+#[cfg(all(feature = "std", atomic_cas))]
+pub unsafe fn unset_boxed_logger() {
+    unset_logger_inner(true)
 }
 
 #[cfg(atomic_cas)]
-fn unset_logger_inner() {
+fn unset_logger_inner(#[cfg(feature = "std")] boxed: bool, #[cfg(not(feature = "std"))] _: bool) {
     let old_state = match STATE.compare_exchange(
         INITIALIZED,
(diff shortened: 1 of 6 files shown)


pablosichert:
Is it possible that the test runner changed between `1.31.0` (MSRV) and `1.51.0` (stable, as of today) to not run integration tests concurrently across files any more?

I'm getting intermittent failures because (I assume) `set_boxed_logger` is called from different files concurrently:

```
' panicked at 'called `Result::unwrap()` on an `Err` value: SetLoggerError(())'
```

src/lib.rs
```diff
@@ -1421,6 +1432,60 @@ impl fmt::Display for SetLoggerError {
...
+        Ordering::SeqCst,
+    ) {
+        Ok(s) | Err(s) => s,
+    };
+    match old_state {
+        INITIALIZED => {
+            #[cfg(feature = "std")]
+            {
+                if boxed {
+                    unsafe { Box::from_raw(LOGGER as *const _ as *mut dyn Log) };
```
sfackler, line 1477: This is not memory safe. It can free the logger while another thread is actively using it.

sfackler: Older versions of the log crate used to have the ability to deregister a logger, but that was removed due to the overhead required to make the operation safe.

pablosichert: That's a good catch, will add that to the safety description of `unset_boxed_logger`.

I understand that this API is unsafe for several reasons and should be avoided in normal operation. However, it would have some really practical benefits for testing.

Is there chance this could be approved when it is behind a `#[doc(hidden)]`?

pablosichert: I tried adding a `#[cfg(test)]`, but that doesn't seem to be visible across crate boundaries.

sfackler: I don't really understand how you would use this safely in a test, given that they will be executing concurrently.

If you want to test different configurations of a logger, I would recommend having some kind of shim in your code that allowed you to install a thread local logger.

pablosichert: > given that they will be executing concurrently

I made sure the test module asserts on the logger state runs all tests sequentially (guarded by a mutex).

If you want the full context, it is concerning this PR: https://github.com/timberio/vector/pull/6926.

sfackler: Again, I think there are better approaches than adding this unsafe API.

pablosichert: Ok, I'll check out if I can find a workaround based on your suggestion.

If I understand correctly, I'd need to get hold of all threads where logs might be emitted and install the logger there. Given that in our case the Tokio runtime controls the thread pool, it doesn't seem very straight-forward.

sfackler: If a thread local approach doesn't work, you can use something else (e.g. make a logger that just buffers up log messages and lets you grab all of them at once, and then run tests sequentially).

pablosichert:
Closing this for now, we'll investigate the approach @sfackler laid out.

PR #459, opened 2021-08-27
Expose logging levels 

WaDelma:
I am using `FromStr` implementation of `LevelFilter` to parse logging level given by user via command line flag. In case user gives invalid logging level I want to print list of allowed values. Currently I have to replicate `LOG_LEVEL_NAMES` array in my code and remember to keep it in sync (It is probably unlikely it changes that much though).

One question I have is that is there a reason it is `static` instead of `const`?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -562,6 +562,8 @@ impl Level {
     }
 
     /// Iterate through all supported logging levels
+    ///
+    /// The order of iteration is from more severe to less severe log messages
     pub fn iter() -> impl Iterator<Item = Self> {
         (1..).flat_map(Self::from_usize)
     }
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -560,6 +560,11 @@ impl Level {
     pub fn as_str(&self) -> &'static str {
         LOG_LEVEL_NAMES[*self as usize]
     }
+
+    /// Iterate through all supported logging levels
+    pub fn iter() -> impl Iterator<Item = Self> {
```
KodrAus, line 565: It might be worth documenting something about the order of these fields. Using `from_usize` we’re effectively returning them in priority order.

What do you think?

KodrAus (approved):
Thanks for working on this @WaDelma!

PR #467, opened 2021-11-14
Add a simple zero-dependency visitor for kv values

### LINKED ISSUE #440: Simple zero-dependency visitor for kv::Value
## Motivation

Users want to be able to visit simple primitives captured in `kv::Value`s but currently resort to chaining calls like:

```rust
if let Some(v) = value.to_i64() {
    ..
} else if let Some(v) = value.to_u64() {
    ..
} ..
```

This is flaky and not fun to write. The goal of the current design is that if you're serializing values you probably already have a serialization framework handy (either `std::fmt`, `serde`, or `sval`) that you'll want to use.

There are reasons to want to visit values without pulling one of these in though. For example, https://github.com/Thomasdezeeuw/std-logger/pull/29#discussion_r553903486 wants to write custom formatting logic to make use of `IoSlice`s, and others don't want the additional mental overhead in learning another framework.

## Proposal

Add a simple zero-dependency visit API for `log::kv::Value` that can be used instead of chaining `to_*` methods or learning a full serialization framework. It intentionally won't support complex values, but should be enough for simple cases to get off the ground with. A new trait, `kv::value::Visitor` will be added with the following methods:

```rust
pub mod value {
    pub trait Visitor<'v> {
        fn visit_i64(&mut self, v: i64) -> Result<(), Error> { self.visit_any(v.to_value()) }
        fn visit_u64(&mut self, v: u64) -> Result<(), Error> { self.visit_any(v.to_value()) }
        fn visit_bool(&mut self, v: bool) -> Result<(), Error> { self.visit_any(v.to_value()) }
        fn visit_str(&mut self, v: &str) -> Result<(), Error> { self.visit_any(v.to_value()) }
        fn visit_borrowed_str(&mut self, v: &'v str) -> Result<(), Error> { self.visit_str(v) }
        fn visit_error(&mut self, v: &(dyn std::error::Error + 'static)) -> Result<(), Error> { self.visit_any(Value::from_dyn_error(v)) }
        fn visit_borrowed_error(&mut self, v: &'v (dyn std::error::Error + 'static)) -> Result<(), Error> { self.visit_error(v) }

        // NOTE: This is the only required method
        fn visit_any(&mut self, v: Value) -> Result<(), Error>;
    }
}
```

with a default implementation looking something like:

```rust
impl<'v> value::Visitor<'v> for MyVisitor {
    fn visit_any(&mut self, v: Value) -> Result<(), Error> {
        write!(self.buf, "{}", v)?;
        Ok(())
    }
}
```

A `Value` can be visited using an inherent `visit` method:

```rust
impl<'v> Value<'v> {
    pub fn visit(&self, visitor: impl Visitor<'v>) -> Result<(), Error> {
        // Here we know what kind of value we're looking at,
        // so can call the most specific method on `value::Visitor`
    }
}
```

Other methods can be overridden to customize their behavior.

## Migrating from `value::Visitor` to `sval` or `serde`

If callers want to switch from `value::Visitor` to `sval` or `serde` to support more complex serialization needs, the change will be from calling `Value::visit` to either `<Value as sval::Value>::stream` or `<Value as serde::Serialize>::serialize`.

### Why not add these to `source::Visitor`?

The `source::Visitor` trait is centred around `visit_pair`, which lets us build higher-level methods on top, like `get`. If we add other methods to `source::Visitor` it may become confusing how the trait should be implemented to support these higher-level methods.

### Why does `visit_any` take a `Value`?

The `Value` type is a carrier for a bunch of trait implementations, `fmt::Debug`, `fmt::Display`, and possibly `sval::Value`, and `serde::Serialize`. By passing in a `Value` here we give the caller flexibility to decide what trait they want to use without adding a lot of complexity to the `value::Visitor` trait.

KodrAus:
Closes #440

This surfaces a zero-dependency visitor for `kv::Value`s that can be used instead of having to chain `to_*` calls when a full serialization framework like `serde` isn't needed.

r? @Thomasdezeeuw

### DIFF
--- a/src/kv/error.rs
+++ b/src/kv/error.rs
@@ -11,6 +11,7 @@ enum Inner {
     #[cfg(feature = "std")]
     Boxed(std_support::BoxedError),
     Msg(&'static str),
+    Value(value_bag::Error),
     Fmt,
 }
 
@@ -21,6 +22,24 @@ impl Error {
             inner: Inner::Msg(msg),
         }
     }
+
+    // Not public so we don't leak the `value_bag` API
+    pub(super) fn from_value(err: value_bag::Error) -> Self {
+        Error {
+            inner: Inner::Value(err),
+        }
+    }
+
+    // Not public so we don't leak the `value_bag` API
+    pub(super) fn into_value(self) -> value_bag::Error {
+        match self.inner {
+            Inner::Value(err) => err,
+            #[cfg(feature = "kv_unstable_std")]
+            _ => value_bag::Error::boxed(self),
+            #[cfg(not(feature = "kv_unstable_std"))]
+            _ => value_bag::Error::msg("error inspecting a value"),
+        }
+    }
 }
 
 impl fmt::Display for Error {
@@ -29,6 +48,7 @@ impl fmt::Display for Error {
         match &self.inner {
             #[cfg(feature = "std")]
             &Boxed(ref err) => err.fmt(f),
+            &Value(ref err) => err.fmt(f),
             &Msg(ref msg) => msg.fmt(f),
             &Fmt => fmt::Error.fmt(f),
         }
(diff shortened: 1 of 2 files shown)


Thomasdezeeuw (approved):
LGTM.

PR #471, opened 2021-11-15
Add Log implementation for &impl Log and Arc<impl Log>

### LINKED ISSUE #458: implement `Log` on more foreign types
We already have `impl <T: Log> Log for Box<T>`. It would be cool to have this for more types, for example `Cow<'static, T: Log>`, `Rc` etc.

I'd have to double-check, but this may even be generalizable to `impl <T, R> Log for T where T: Borrow<R>,  R: Log`

piegamesde:
Closes #458. Turns out that `Rc<Log>` isn't possible because `Log: Sync`. Other candidates for adding:

- `Cow`
- `Option`
- `()` (this would allow us to get rid of the internal `NopLogger`)
- `Pin` (?)
- Probably forgot something

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1284,24 +1284,6 @@ where
 }
 
 #[cfg(feature = "std")]
-impl<P, T> Log for std::pin::Pin<P>
-where
-    P: std::ops::Deref<Target = T> + std::marker::Send + std::marker::Sync,
-    T: Log + ?Sized,
-{
-    fn enabled(&self, metadata: &Metadata) -> bool {
-        (**self).enabled(metadata)
-    }
-
-    fn log(&self, record: &Record) {
-        (**self).log(record)
-    }
-    fn flush(&self) {
-        (**self).flush()
-    }
-}
-
-#[cfg(feature = "std")]
 impl<T> Log for std::sync::Arc<T>
 where
     T: ?Sized + Log,
@@ -1900,7 +1882,7 @@ mod tests {
     fn test_foreign_impl() {
         use super::Log;
         #[cfg(feature = "std")]
-        use std::{pin::Pin, sync::Arc};
+        use std::sync::Arc;
 
         fn assert_is_log<T: Log + ?Sized>() {}
 
@@ -1910,9 +1892,6 @@ mod tests {
         assert_is_log::<Box<dyn Log>>();
 
         #[cfg(feature = "std")]
-        assert_is_log::<Pin<Box<dyn Log>>>();
-
-        #[cfg(feature = "std")]
         assert_is_log::<Arc<dyn Log>>();
 
         // Assert these statements for all T: Log + ?Sized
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1250,6 +1250,22 @@ impl Log for NopLogger {
     fn flush(&self) {}
 }
 
+impl<T> Log for &'_ T
+where
+    T: ?Sized + Log,
+{
+    fn enabled(&self, metadata: &Metadata) -> bool {
+        (*self).enabled(metadata)
```
KodrAus, line 1258: Normally I'd write something like `(**self).enabled(metadata)` when forwarding implementations on references. I just thought I'd check to make sure this isn't a possible infinite recursion 🙂

piegamesde: I'd forgotten the * and I had a warning, then I added it and the warning went away. I could add a test that would otherwise trigger the infinite recursion if you like.

KodrAus:
Thanks @piegamesde! I think this is a great starting point. I think ideally we'd stick to implementations that can unconditionally forward to another implementation, like `Box<T>`, `Arc<T>`, `&T`. Just to avoid any possible confusion over how something like `Option<T>` or `()` should actually behave.

piegamesde:
I'd argue that the intent for `Option<T>` is rather clear and for `()` more or less as well. But I agree that this can be overdone (looking at you, Haskell) and one has to draw the line somewhere.

This still leaves `Pin` and `Cow` open. I'd be in favor of `Pin` because I don't see any reason *not* to (the first person coming up with a legitimate use case for `Pin<Box<Log>>` wins a cookie). For `Cow`, I don't know. *Technically* the implementation would simply forward along its `Borrow` so its good?

KodrAus:
Yeh, there are subtleties in the semantics that I don't think the benefit outweighs for `Option` or `()`. `Pin` and `Cow` both do seem reasonable. Would you like to include them in this PR or follow-up with another?

src/lib.rs
KodrAus: The more tests the merrier!

piegamesde:
> Would you like to include them in this PR or follow-up with another?

Let's do all that make sense right now. I added `Pin`, but I had to back off `Cow` because it's `Clone` constraint on `T`. This is something that does make little sense for loggers. This is not the first time I've wished for some `MaybeOwned` in Rust, i.e. a `Cow` without the "clone" and the "write" part.

For the potential recursion issue mentioned in review, I added `#![deny(unconditional_recursion)]` instead of a test.

KodrAus:
Ah it looks like we might need to back `Pin` out too 🙂 It was actually still unstable in 1.31.0

piegamesde:
Yes, I've seen the CI fail too. Is there a way to feature guard this only for newer Rust versions?

KodrAus:
Ah unfortunately we can’t really introduce any unconditional dependencies such as `rustversion` to `log`. We could do some version detection in the build script, but the best option for now might be to simply roll back `Pin`.

piegamesde:
I removed the `Pin` implementation in a second commit so that you can simply revert it after the next MSRV bump. Feel free to squash if you prefer it that way.

KodrAus (approved):
Thanks for working on this @piegamesde!

Issue #436, opened 2021-01-08
Stabilize kv_unstable


KodrAus:
This is part of #328 

I think we've explored enough of this design space now and the API has been in use within the `tide` project for some time now with their own kv-capable logging macro.

By stabilizing just the backend (non-macro-supported) `kv` module we can let users depend on this with confidence so that projects like `tracing` can also utilize it.

# Migration path

The `kv` API will remain an opt-in optional API. All features that currently begin with `kv_unstable` will be changed to `kv`. We can keep these old unstable features around for a while if we need.

Any other changes that are needed should be included here.

# Steps

- [x] Public API review of the `kv` module
- [x] Stabilize [`sval`](https://github.com/sval-rs/sval)
- [x] Stabilize [`value-bag`](https://github.com/sval-rs/value-bag)
- [ ] Documentation and examples

cc @yoshuawuyts

Thomasdezeeuw:
I don't know what the plan is for the `kv_unstable_std` feature, but can that be removed? Surely most users mean to enable `kv_unstable_std` if `std` and `kv` are enabled.

KodrAus:
@Thomasdezeeuw unfortunately it’s a limitation of Cargo features. Since it’s not possible to activate a feature of a dependency when two unrelated features are active we need one that covers them both.

Thomasdezeeuw:
@KodrAus in the code you can use `#[cfg(all(feature = "kv", feature = "std"))]` in stead of `#[cfg(feature = "kv_unstable_std")]`, removing the `kv_unstable_std` feature.

KodrAus:
@Thomasdezeeuw we can do that in code, but we also have dependencies that need their own `std` features enabled too.

Thomasdezeeuw:
@KodrAus could you make another release to try the new macro syntax?

KodrAus:
@Thomasdezeeuw Will do! I'll get the ball rolling on this now.

Thomasdezeeuw:
@KodrAus any update on a release? We've been using the git version without issues for a while (including the new macro syntax).

KodrAus:
@Thomasdezeeuw I've been waiting for some feedback from the wider project on https://github.com/rust-lang/log/issues/343#issuecomment-1002851191, but it's been a bit quiet. I'm pretty confident in the direction we're going though and that we're not heading down a path we'll regret later so will push ahead.

KodrAus:
[...]

PR #476, opened 2021-12-07
Allow macros to be more consistent with kvs

KodrAus:
Part of #436 

This tweaks the new key-value macro syntax slightly so that it can be more consistent in cases where key-values are supplied and in cases where they're not. All the old syntax is still supported, but we can now write `log!` macros like:

```rust
info!(target = "my_target"; "format {string}");
```

which then supports key-value pairs as:

```rust
info!(target = "my_target"; key1 = value1, key2 = value2; "format {formatted1}", formatted1 = value1);
```

For key-values, the `target = x;` form is required. It splits our macro up into 3 logical parts, separated by `;`s:

1. `target = "my_target"`: The portion of the log record that's considered by top-level filtering. It's equivalent to `log_enabled!`.
2. `key1 = value1, key2 = value2`: The structured data associated with the log record.
3. `"format {formatted1}", formatted1 = value1`: The unstructured message associated with the log record.

I think for `log` here we should also remove the `capture_*` methods on `Value` for now, because they lead to unpleasant compilation errors when you don't have a `&T: 'static` around. They're only useful if you want to try downcast to a specific type, like a `Uuid` or ` Timestamp`.

I'll try flesh out our test coverage even more, since there are a lot of possible permutations of the log macros and their interactions with `format_args` to consider.

cc @Thomasdezeeuw. What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,5 +93,22 @@ The executable itself may use the `log` crate to log as well.
 If you enable the `kv_unstable` feature, you can associate structured data with your log records:
 
 ```rust
+use log::{info, trace, warn, as_serde, as_error};
 
+pub fn shave_the_yak(yak: &mut Yak) {
+    trace!(target = "yak_events"; yak = as_serde!(yak); "Commencing yak shaving");
+
+    loop {
+        match find_a_razor() {
+            Ok(razor) => {
+                info!(razor = razor; "Razor located");
+                yak.shave(razor);
+                break;
+            }
+            Err(err) => {
+                warn!(err = as_error!(err); "Unable to locate a razor, retrying");
+            }
+        }
+    }
+}
 ```
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ If you enable the `kv_unstable` feature, you can associate structured data with 
 use log::{info, trace, warn, as_serde, as_error};
 
 pub fn shave_the_yak(yak: &mut Yak) {
-    trace!(target = "yak_events"; yak = as_serde!(yak); "Commencing yak shaving");
+    trace!(target = "yak_events", yak = as_serde!(yak); "Commencing yak shaving");
 
     loop {
         match find_a_razor() {
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ This version is explicitly tested in CI and may be bumped in any release as need
 
 ## Usage
 
-## In libraries
+### In libraries
 
 Libraries should link only to the `log` crate, and use the provided macros to
 log whatever information will be useful to downstream consumers:
@@ -55,7 +55,7 @@ pub fn shave_the_yak(yak: &mut Yak) {
 }
 ```
 
-## In executables
+### In executables
 
 In order to produce log output, executables have to use a logger implementation compatible with the facade.
 There are many available implementations to choose from, here are some of the most popular ones:
@@ -87,3 +87,11 @@ function to do this. Any log messages generated before the logger is
 initialized will be ignored.
 
 The executable itself may use the `log` crate to log as well.
+
+## Structured logging
+
+If you enable the `kv_unstable` feature, you can associate structured data with your log records:
+
+```rust
+
+```
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -24,7 +24,7 @@
 //! though that default may be overridden. Logger implementations typically use
 //! the target to filter requests based on some user configuration.
 //!
-//! # Use
+//! # Usage
 //!
 //! The basic use of the log crate is through the five logging macros: [`error!`],
 //! [`warn!`], [`info!`], [`debug!`] and [`trace!`]
@@ -55,7 +55,7 @@
 //! use log::{info, warn};
 //!
 //! pub fn shave_the_yak(yak: &mut Yak) {
-//!     info!(target: "yak_events", "Commencing yak shaving for {:?}", yak);
+//!     info!(target = "yak_events"; "Commencing yak shaving for {:?}", yak);
 //!
 //!     loop {
 //!         match find_a_razor() {
@@ -85,6 +85,42 @@
 //! ### Warning
 //!
 //! The logging system may only be initialized once.
+//!
+//! ## Structured logging
+//!
+//! If you enable the `kv_unstable` feature you can associate structured values
+//! with your log records. If we take the example from before, we can include
+//! some additional context besides what's in the formatted message:
+//!
+//! ```edition2018
+//! # #[macro_use] extern crate serde;
+//! # #[derive(Debug, Serialize)] pub struct Yak(String);
+//! # impl Yak { fn shave(&mut self, _: u32) {} }
+//! # fn find_a_razor() -> Result<u32, std::io::Error> { Ok(1) }
+//! # #[cfg(feature = "kv_unstable_serde")]
+//! # fn main() {
+//! use log::{info, warn, as_serde, as_error};
+//!
+//! pub fn shave_the_yak(yak: &mut Yak) {
+//!     info!(target = "yak_events"; yak = as_serde!(yak); "Commencing yak shaving");
+//!
+//!     loop {
+//!         match find_a_razor() {
+//!             Ok(razor) => {
+//!                 info!(razor = razor; "Razor located");
+//!                 yak.shave(razor);
+//!                 break;
+//!             }
+//!             Err(err) => {
+//!                 warn!(err = as_error!(err); "Unable to locate a razor, retrying");
+//!             }
+//!         }
+//!     }
+//! }
+//! # }
+//! # #[cfg(not(feature = "kv_unstable_serde"))]
+//! # fn main() {}
+//! ```
 //!
 //! # Available logging implementations
 //!
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -55,7 +55,7 @@
 //! use log::{info, warn};
 //!
 //! pub fn shave_the_yak(yak: &mut Yak) {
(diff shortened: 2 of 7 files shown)


Thomasdezeeuw:
> Part of #436
> 
> This tweaks the new key-value macro syntax slightly so that it can be more consistent in cases where key-values are supplied and in cases where they're not. All the old syntax is still supported, but we can now write `log!` macros like:
> 
> ```rust
> info!(target = "my_target"; "format {string}");
> ```
> 
> which then supports key-value pairs as:
> 
> ```rust
> info!(target = "my_target"; key1 = value1, key2 = value2; "format {formatted1}", formatted1 = value1);
> ```
> 
> For key-values, the `target = x;` form is required. It splits our macro up into 3 logical parts, separated by `;`s:
> 
>     1. `target = "my_target"`: The portion of the log record that's considered by top-level filtering. It's equivalent to `log_enabled!`.
> 
>     2. `key1 = value1, key2 = value2`: The structured data associated with the log record.
> 
>     3. `"format {formatted1}", formatted1 = value1`: The unstructured message associated with the log record.

Did `target: x` not work? That is what is currently used.

> I think for `log` here we should also remove the `capture_*` methods on `Value` for now, because they lead to unpleasant compilation errors when you don't have a `&T: 'static` around. They're only useful if you want to try downcast to a specific type, like a `Uuid` or ` Timestamp`.

And replace them with the `from_display!` macros?

KodrAus:
> Did target: x not work? That is what is currently used.

It does work, but when I started using it I found the inconsistency with key values and named format args a bit jarring. We’ll probably want some `err = expr` syntax too at some point so thought starting to try normalize everything would be good.

> And replace them with the from_display! macros?

Those methods and direct downcasting will just go away, unless it’s something you’re currently using? Speaking of, what do you think of those little macros? The alternative is something like `tracing`’s flags, which are similar to `format_args`.

Thomasdezeeuw:
> > Did target: x not work? That is what is currently used.
> 
> It does work, but when I started using it I found the inconsistency with key values and named format args a bit jarring. We’ll probably want some `err = expr` syntax too at some point so thought starting to try normalize everything would be good.
> 

I'm not sure. One the one hand I can see your argument, on the other target it is different from the key-values as it has its own method to access it (not via `Record::key_values`, but via `Record::target`). Being inconsistent with the non-kvs macros could be annoying. For example when you quickly want to add a kv pair to a message you now also have to consider the target. But I don't know how munch of an annoyance that will be.

> > And replace them with the from_display! macros?
> 
> Those methods and direct downcasting will just go away, unless it’s something you’re currently using?

I do use the downcasting but I'll switch to the visitor API introduces in https://github.com/rust-lang/log/pull/467, so a nice deprecation period would be appreciated.

>  Speaking of, what do you think of those little macros? The alternative is something like `tracing`’s flags, which are similar to `format_args`.

I could either way. The way the tracing crate does is a bit unusual on first sight, but I image you get used to it. So it's a question of being more explicit (macros) vs quicker to write (tracing like flags). I'm with either.

KodrAus:
> I do use the downcasting but I'll switch to the visitor API 

Ah that's right, you've been forced to use a nasty chain of `if let`s haven't you because that visitor API wasn't available. I'll make sure it doesn't disappear on you in the meantime.

> when you quickly want to add a kv pair to a message you now also have to consider the target.

You can still specify key-value pairs without having to specify a target:

```rust
info!(cat_1 = "chashu", cat_2 = "nori", cat_count = 2; "hello {world}", world = "world");
```

...but that does raise an interesting forwards compatibility hazard. Say you write:

```rust
info!(target = "kv_target", cat_1 = "chashu", cat_2 = "nori"; "hello {}", "world");
```

In this scheme, that `target` will end up as a key-value pair instead of as the `target`. Since we then default the target to `module_path!()` you'll end up with `record.target()` returning something else. This is solvable with a bit more macro work.

>  The way the tracing crate does is a bit unusual on first sight, but I image you get used to it.

Yeh, I think once you learn the scheme it's not too bad. My main reason not for trying to pursue it here is that I think formatting flags are a bit of a dead end. They're crafted for the needs of `format_args`, but once we're outside their use-case we need to pick new arbitrary characters for new things, like capturing with `serde`, and those new things don't necessarily integrate with any other flags (what would it mean to capture with `serde` and also pad with spaces?). But if we want to come up with a nice shorthand scheme we could deprecate these `as_*` macros pretty easily.

I don't want to keep you blocked on this for much longer though, so maybe after we're sure we're not painted into any corners (which I don't think we are this time) we can create a bit of a checklist on #343 of incomplete parts in the macros so they don't need to hold up getting something out?

Thomasdezeeuw:
> > when you quickly want to add a kv pair to a message you now also have to consider the target.
> 
> You can still specify key-value pairs without having to specify a target:
> 
> ```rust
> info!(cat_1 = "chashu", cat_2 = "nori", cat_count = 2; "hello {world}", world = "world");
> ```

I meant going from a log with a target to a log with a target and key-values, for example:

```rust
// Original:
info!(target: "yak_events", "Commencing yak shaving for {:?}", yak);

// New
info!(target: "yak_events", key = value; "Commencing yak shaving for {:?}", yak); // This won't work.
info!(target = "yak_events", key = value; "Commencing yak shaving for {:?}", yak); // This will.
```

Or am I mistaken here?

> ...but that does raise an interesting forwards compatibility hazard. Say you write:
> 
> ```rust
> info!(target = "kv_target", cat_1 = "chashu", cat_2 = "nori"; "hello {}", "world");
> ```
> 
> In this scheme, that `target` will end up as a key-value pair instead of as the `target`. Since we then default the target to `module_path!()` you'll end up with `record.target()` returning something else. This is solvable with a bit more macro work.

I was saying it could be confusing. With your example above it's not clear that target is "special" (in the sense it can be retrieved using `Record.target` instead of being another key-value pair). Keeping the ":" syntax might make it clearer?

And it allows people to use `target` as a key in a key-value pair.

> > The way the tracing crate does is a bit unusual on first sight, but I image you get used to it.
> 
> Yeh, I think once you learn the scheme it's not too bad. My main reason not for trying to pursue it here is that I think formatting flags are a bit of a dead end. They're crafted for the needs of `format_args`, but once we're outside their use-case we need to pick new arbitrary characters for new things, like capturing with `serde`, and those new things don't necessarily integrate with any other flags (what would it mean to capture with `serde` and also pad with spaces?). But if we want to come up with a nice shorthand scheme we could deprecate these `as_*` macros pretty easily.

👍 

> I don't want to keep you blocked on this for much longer though, so maybe after we're sure we're not painted into any corners (which I don't think we are this time) we can create a bit of a checklist on #343 of incomplete parts in the macros so they don't need to hold up getting something out?

I think the only decision we need to make is for taget ":" vs "=". We can use macros for formatting and add special tracing-like flags later without breaking compatibility right?

KodrAus:
Sounds good 👍

I’m not too strongly tied to my `=` preference, possibly because I just don’t ever set `target` myself when logging. We could accept both forms still if we wanted to. I was thinking of working towards unifying key values with other pieces of metadata we associate with a log record so there isn’t a distinction in the end between target and anything else.

I would be ok with keeping target using `:` and deprecate it by making a key-value pair called `target` just work in the future.

Thomasdezeeuw:
> Sounds good 👍
> 
> I’m not too strongly tied to my `=` preference, possibly because I just don’t ever set `target` myself when logging. We could accept both forms still if we wanted to. I was thinking of working towards unifying key values with other pieces of metadata we associate with a log record so there isn’t a distinction in the end between target and anything else.
> 
> I would be ok with keeping target using `:` and deprecate it by making a key-value pair called `target` just work in the future.

Sounds good. I think for backwards compatibility we want to keep `:` for now.

Thomasdezeeuw (approved):

KodrAus:
Ok, I've reverted the need to make use the `target =` form in kv-enabled macros to avoid churning things. We can now technically write this though:

```rust
info!(target = "my_target", cat_1 = "chashu", cat_2 = "nori", cat_count = 2; "hello {}", "world");
```

In this case we'll interpret `target` as a key-value pair. This is how I think we'll be able to get away from `target:` in the future, by making the macro effectively:

```rust
$lvl:expr, $($key:ident = $value:expr),+; $($arg:tt)+
```

where we can use a tt-muncher to unpack the key-value pairs and pick off a `target`, `error` etc if we find them, but don't have any special syntax for declaring them apart from any other piece of data you might want to associate with your record.

Once this is green I'll merge it in and give it a smoke test in any candidate libraries to make sure it doesn't regress things and then get it out the door.

README.md
```diff
@@ -87,3 +87,28 @@ function to do this. Any log messages generated before the logger is
...
+
+## Structured logging
+
+If you enable the `kv_unstable` feature, you can associate structured data with your log records:
+
+```rust
+use log::{info, trace, warn, as_serde, as_error};
+
+pub fn shave_the_yak(yak: &mut Yak) {
+    trace!(target = "yak_events", yak = as_serde!(yak); "Commencing yak shaving");
```
Thomasdezeeuw, line 99: Nit: is using target here maybe confusing for new users?

src/lib.rs
```diff
@@ -55,7 +55,7 @@
 //! use log::{info, warn};
 //!
 //! pub fn shave_the_yak(yak: &mut Yak) {
-//!     info!(target: "yak_events", "Commencing yak shaving for {:?}", yak);
+//!     info!(target = "yak_events", "Commencing yak shaving for {:?}", yak);
```
Thomasdezeeuw, line 58: This needs to be reverted to `:` to fix the CI.

Thomasdezeeuw (approved):

PR #477, opened 2021-12-15
Target maybe static

jelmansouri:
Our telemetry module, enables us to pool static strings and send them once instead of copying and writing a string at each log line. The target is part of the information that gets sent to do additional filtering when logs are consumed.
The goal is to introduce this optimization without breaking the public API, so we can benefit from the optimization in all of our dependencies.
The code is less DRY than I would have hoped so I'm very open to suggestions regarding how to write this is a better fashion!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,11 @@ name = "macros"
 path = "tests/macros.rs"
 harness = true
 
+[[test]]
+name = "targets"
+path = "tests/targets.rs"
+harness = false
+
 [features]
 max_level_off   = []
 max_level_error = []
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -892,6 +892,12 @@ impl<'a> Record<'a> {
         self.metadata.target()
     }
 
+    /// The name of the target of the directive.
+    #[inline]
+    pub fn target_static(&self) -> Option<&'static str> {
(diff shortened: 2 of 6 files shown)


jelmansouri:
Regarding the MSRV, the additional code is a further optimization in cases like `info!(target: "my_target", "my log line")` I can either remove it or try some trickery with `cfg(version("1.when_it_was_introduced"))`. Maybe overkill and we can revisit when the MSRV is bumped.

KodrAus:
Thanks for your patience and for working on this one @jelmansouri! It's a bit unfortunate that

```rust
info!(target: static_str(), "Some message");
```

won't produce a static target, since it won't match the literal arm in the macro. Before accepting this I think I'd like to see some more exploration of the motivations and your specific use-case, what lead to this particular approach, and what its benefits and drawbacks are. If you're already pooling static strings then a single allocation to convert any `&str` into owned `String`s the first time you see them doesn't sound prohibitive to me, but you might have already been down that path 🙂

KodrAus:
Just coming back through some triage. It's been a while since we've had activity on this one and a lot has changed in the meantime that's created conflicts here. I'll go ahead and close this one but will be open to reviewing a fresh PR that explores static value capturing.

PR #478, opened 2021-12-27
change url in README to doc.rs

PAN-Ziyue:
changed two url in README.md to the link on doc.rs


KodrAus:
Hi @PAN-Ziyue :wave:

Thanks for your patience on this one! It looks like CI just got a bit stuck for some of the jobs, but that's unrelated to your change.

Thanks for fixing this!

Issue #484, opened 2022-02-21
Allow static string to be used as key in macros


Thomasdezeeuw:
Currently it uses an `indent`, which doesn't allow for static strings. I would like to be able to do the following:

```rust
log::warn!(trace = log::as_display!("some trace id"); "message"); // Ok.
log::warn!("logging.googleapis.com/trace" = log::as_display!("some trace id"); "message"); // Doesn't compile.
```

It could be that this is very niche and complex to implement, in which case we can continue using our work around.

Related https://github.com/rust-lang/log/issues/436, /cc @KodrAus

KodrAus:
I think this is something that would be pretty straightfoward using a tt-muncher, which we're already going to need fairly soon. We'd just need to support both `ident` and `literal` matches in our match arms. If you pass something other than a string through then we'll catch it attempting to convert that value into a `Key`.

PR #486, opened 2022-02-25
Support static string as key value

Thomasdezeeuw:
Fixes https://github.com/rust-lang/log/issues/484.


Thomasdezeeuw:
@KodrAus this was a lot easier then I expected, so I'm a little I'm forgetting about a use case I'm breaking.

src/macros.rs
```diff
@@ -30,7 +30,7 @@
 #[macro_export(local_inner_macros)]
 macro_rules! log {
     // log!(target: "my_target", Level::Info; key1 = 42, key2 = true; "a {} event", "log");
-    (target: $target:expr, $lvl:expr, $($key:ident = $value:expr),+; $($arg:tt)+) => ({
+    (target: $target:expr, $lvl:expr, $($key:tt = $value:expr),+; $($arg:tt)+) => ({
```
KodrAus, line 33: Ahh unfortunately I think this will include quotes in string keys. So for `"string/key"` the actual string we get will look like `"\"string/key\""`.

Looking at this again though, I don't think we actually need a tt muncher yet afterall. We can just use another macro around `__log_stringify!` (or just re-use `__log_stringify!` for it) that matches on literals and idents and handles them appropriately.

KodrAus:
@Thomasdezeeuw I'll still push ahead with `0.4.15` in its current state, but once we're ready with this one we can do a `0.4.16`. It would be good to get back into the habit of more regular releases here anyways.

KodrAus (approved):
This looks good to me!

PR #491, opened 2022-03-20
Implement ToValue and From<$ty> for Value for NonZero

Thomasdezeeuw:
It can already by be done by calling `.get()`, but this makes it a bit
easier.


KodrAus (approved):
Thanks @Thomasdezeeuw!

PR #502, opened 2022-04-13
Add reference to db_logger

jmmv:
db_logger is a logging implementation that writes log entries to a
database.  At the time of this writing, it supports PostgreSQL and
SQLite via the sqlx crate.


KodrAus:
Thanks @jmmv!

PR #504, opened 2022-04-20
Upgrade to value-bag 1.0.0-alpha.9 and remove by-value 128bit int conversions

KodrAus:
This PR upgrades our `value-bag` version to `1.0.0-alpha.9`, which includes a breaking change to how 128bit numbers are handled. Instead of storing them by-value we only store them by reference. Along with some other internal layout changes this reduces the size of `Value` from 48 to 24 bytes, which is a great saving for the stack-local arrays we create to hold them.

In practice this doesn't affect whether or not 128bit values can be used in the macros because they rely on the `ToValue` trait, which is always by-reference. I'd be keen to hear if anybody is relying on the `From` impls though.

I'm hoping to get this through sooner rather than later in case https://github.com/rust-lang/rust/pull/95845 is merged in its current state that breaks constant type id comparison, which will mean shipping a patch to `value-bag` for `nightly` users.

r? @Thomasdezeeuw

### DIFF
--- a/src/kv/value.rs
+++ b/src/kv/value.rs
@@ -564,9 +564,7 @@ macro_rules! impl_value_to_primitive {
     }
 }
 
-impl_to_value_primitive![
-    usize, u8, u16, u32, u64, isize, i8, i16, i32, i64, f32, f64, char, bool,
-];
+impl_to_value_primitive![usize, u8, u16, u32, u64, isize, i8, i16, i32, i64, f32, f64, char, bool,];
 
 #[rustfmt::skip]
 impl_to_value_nonzero_primitive![
--- a/src/kv/value.rs
+++ b/src/kv/value.rs
@@ -485,12 +485,14 @@ impl<'v> From<&'v i128> for Value<'v> {
 
 impl<'v> From<&'v std::num::NonZeroU128> for Value<'v> {
     fn from(v: &'v std::num::NonZeroU128) -> Value<'v> {
+        // SAFETY: `NonZeroU128` and `u128` have the same ABI
         Value::from_value_bag(unsafe { std::mem::transmute::<&std::num::NonZeroU128, &u128>(v) })
     }
 }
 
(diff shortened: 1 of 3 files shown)


KodrAus:
We can work around the breakage if it's worth it by the way. I just carried it through to `log` because it seemed appropriate.

src/kv/value.rs
```diff
@@ -514,14 +564,12 @@ macro_rules! impl_value_to_primitive {
     }
 }
 
-impl_to_value_primitive![
-    usize, u8, u16, u32, u64, u128, isize, i8, i16, i32, i64, i128, f32, f64, char, bool,
-];
+impl_to_value_primitive![usize, u8, u16, u32, u64, isize, i8, i16, i32, i64, f32, f64, char, bool,];
```
Thomasdezeeuw, line 567: Nit: final `,`.

Thomasdezeeuw (approved):
LGTM.

I also took a quick the changes to sval and I think I found a breaking one in https://github.com/sval-rs/value-bag/commit/3fb84ab0578e176b7cff1c154bda490af3795dba#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R24-R25. It uses `?` in enabling feature, which is 1.60+, while log still supports 1.31 (although we could bump that).

src/kv/value.rs
KodrAus: I think that macro implementation is a little iffy and it actually requires that trailing `,` 😄

KodrAus:
> It uses ? in enabling feature, which is 1.60+, while log still supports 1.31 (although we could bump that).

Ah well spotted! We actually [only require the latest stable for the key-value API](https://github.com/rust-lang/log/blob/master/Cargo.toml#L48), so are ok to bump this. That feature is _perfect_ for that library because it stitches together serialization APIs. Before I was wondering how I could ever introduce a new framework like `valuable` without exploding the crate graph, but now that's actually possible.

PR #511, opened 2022-05-04
clean clippy warnings

gkorland:


Thomasdezeeuw:
This breaks the MSRV, so I don't think will be accepted.

KodrAus:
Hi @gkorland :wave:

Thanks for the PR! Unfortunately @Thomasdezeeuw is right and some of these changes rely on compiler features we can't use in our minimum supported version (which is currently `1.31.0`).

PR #514, opened 2022-05-31
Move `lvl <= $crate::max_level()` inside called functions

EFanZh:
This change will reduce binary size at the cost of some performance loss.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1588,22 +1588,24 @@ pub fn __private_api_log(
     &(target, module_path, file, line): &(&str, &'static str, &'static str, u32),
     kvs: Option<&[(&str, &str)]>,
 ) {
-    if kvs.is_some() {
-        panic!(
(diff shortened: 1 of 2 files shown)


EFanZh:
I noticed a project I am working on has a decent size increase after upgrading the `log` crate dependency, it was cause by #446, but maybe there is a way to reduce the binary size impact of upgrading, so here is what I came up with. If the performance loss is unacceptable, how about using an option feature to provide this binary size over performance behavior?

EFanZh:
Close as this will expand argument too early, causing behavior change.

Thomasdezeeuw:
@EFanZh have you tried the log crate's filter features (https://docs.rs/log/latest/log/#compile-time-filters)? Those will help remove trace, debug (and more if desired) from the binary. It's also (part of) the reason why the if statement if outside of the function.

EFanZh:
@Thomasdezeeuw Most of my logs are required to be printed, so disabling logs is not an option.

PR #516, opened 2022-05-31
Apply Clippy lints

hellow554:
This is similar to #511 but adds a `clippy.toml` to tell the msrv which prevents features to be pulled in that are not released in that version.

Also add a github worflow for clippy


src/lib.rs
```diff
@@ -653,88 +585,18 @@ pub enum LevelFilter {
     Trace,
 }
 
-// Deriving generates terrible impls of these traits
```
hellow554, line 656: This statement is not right. I created a godbolt for it. There's no real difference between derived impls and custom impls (anymore): https://rust.godbolt.org/z/7K1vjG6KE

src/kv/key.rs
```diff
@@ -33,15 +31,15 @@ impl ToKey for str {
 }
 
 /// A key in a structured key-value pair.
-#[derive(Clone)]
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
```
KodrAus, line 34: I think one of the reasons we originally wrote these impls out by-hand was because there was some idea of adding an optional index to the key type that could be used as an optimization for lookup, but not considered when comparing.

That's not exactly obvious from the original code though, so maybe we can drop a comment there to suggest it.

```suggestion
// These impls must only be based on the as_str() representation of the key
// If a new field (such as an optional index) is added to the key they must not affect comparison
#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
```

KodrAus (commented):
Thanks @hellow554! I didn't know about `clippy.toml` :eyes: That's great to know. I just had one comment, which doesn't actually affect the code you've fixed, but otherwise this looks good to me.

hellow554:
Done :)

src/lib.rs
```diff
@@ -1546,7 +1420,7 @@ impl error::Error for SetLoggerError {}
 ///
 /// [`from_str`]: https://doc.rust-lang.org/std/str/trait.FromStr.html#tymethod.from_str
 #[allow(missing_copy_implementations)]
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Eq)]
```
Thomasdezeeuw, line 1423: Functional/API change, but personally I'm ok with it.

hellow554: Clippy suggested it and I thought it made sense.

Issue #509, opened 2022-04-29
Is it necessary to use Ordering::SeqCst in the log!  ??


relufi:
```rust
pub fn logger() -> &'static dyn Log {
   // Is it necessary to use Ordering::SeqCst in log!
    if STATE.load(Ordering::SeqCst) != INITIALIZED {
        static NOP: NopLogger = NopLogger;
        &NOP
    } else {
        unsafe { LOGGER }
    }
}
```

See the implementation of std::lazy::SyncOnceCell or parking_lot::Once

relufi:
I tried to use a more relaxed order and wrote a [test](https://github.com/relufi/log/blob/8b5da221f8c53211a5534fe530adb53abc31b430/src/lib.rs#L1969)
https://github.com/relufi/log/blob/use-relaxed-memory-barriers/src/lib.rs

However, when I tried to change all the memory order to Relaxed and tested on the arm platform, there was no visibility problem either.

KodrAus:
I think the ordering there could probably be downgraded so long as we always see the store of `STATE` after the store of `LOGGER`. The orderings on initialization are also pretty conservative, but it's always hard to tell whether that's because they're necessary or because they haven't been optimized.

relufi:
Reference std::sync::Once SeqCst is not necessary, I am concerned that using SeqCst in every log! will affect performance (log! is too common and SeqCst is too expensive on some platforms).
If you're worried about problems you can just introduce std::sync::Once or parking_lot::Once to avoid them.

relufi:
I am using a translation tool, so please bear with me if I make any mistakes.

Thomasdezeeuw:
Since #610 we're using `Acquire` in `logger`, so closing this.

PR #520, opened 2022-06-10
Relax atomic orderings

Cassy343:
Currently this crate uses `SeqCst` orderings for all atomic operations related to logger initialization. At most these orderings would need to be `AcqRel` since they only involve a single atomic. This PR relaxes all orderings as much as possible without introducing visibility issues.

I also changed the initialization of the filter to be `LevelFilter::Off as usize` for clarity.


Cassy343:
Closes #509 

In talking with some other developers some concern was presented over whether or not a write to `LOGGER` could be re-ordered before a relaxed atomic load despite that write being conditional. [This](https://github.com/rust-lang/log/pull/520/commits/2e69066bc6986ae179fa743daef8a380836cc1cd) commit defensively changes the ordering to `Acquire` to ensure this does not happen, but if it is guaranteed that such a reordering could not occur with `Relaxed` then that commit can be reverted.

Thomasdezeeuw:
I'm not sure that this new ordering is valid, but does it really matter? The logger is set at most once, usually during startup, personally I don't think the performance concerns at that stage outweigh getting this wrong...

Thomasdezeeuw:
@Cassy343 I'm familiar with atomic ordering, that wasn't my point. My point was that the ordering is tricky to get right (even if you following docs precisely) and that I don't think it's worth it for the setup phase of a program.

So, I propose to reduce this pr to just change the atomic ordering of the `logger` function as that's really the only one where performance matters.

Regarding the ordering `logger` I think we can lower it all the way to `Relaxed`, *iff* we properly document it. I imagine (but have no data to backup this statement) that most multi-threaded applications will setup logging before starting threads. Given that, changing `logger` to `Relaxed` shouldn't be an issue. One exception might be testing, where the creation of the test threads isn't handled by the developer (assuming the default testing harness). But again make a reasonable assuming that all tests that need logging will attempt to set it up, thus calling something that loads using `SeqCst` again, I don't think this will be an issue.

Cassy343:
That ordering should not be reduced to relaxed. Even if we properly document it, if one thread initializes the logger and another concurrently tries to access the logger, then that's undefined behavior which is not allowed to occur in safe code, so that function would need to be marked as unsafe.

I'd also wager that a non-trivial number of users are initializing the logger after spinning up threads due to libraries such as tokio and rocket providing macros that wrap your `main` and spin up a thread pool before executing your code. An acquire load is the minimally sufficient ordering to make that function safe provided that, after `LOGGER` is written to, `STATE` is set to `INITIALIZED` with release ordering or stronger.

Thomasdezeeuw:
> That ordering should not be reduced to relaxed. Even if we properly document it, if one thread initializes the logger and another concurrently tries to access the logger, then that's undefined behavior which is not allowed to occur in safe code, so that function would need to be marked as unsafe.

How do you think this is possible?

> I'd also wager that a non-trivial number of users are initializing the logger after spinning up threads due to libraries such as tokio and rocket providing macros that wrap your `main` and spin up a thread pool before executing your code. An acquire load is the minimally sufficient ordering to make that function safe provided that, after `LOGGER` is written to, `STATE` is set to `INITIALIZED` with release ordering or stronger.

I disagree. I think if you put a `atomic::fence(SeqCst)` at the bottom of `set_logger` and use `STATE.load(Relaxed) == INITIALIZED` it not possible for a thread to read `state == INITIALIZED` with the logger not set.

Cassy343:
To add on to @chorman0773's point, here is a (correctly failing) loom test demonstrating that `Relaxed` ordering causes UB:
```rs
fn main() {
    use loom::{sync::atomic::{AtomicUsize, Ordering, fence}, cell::UnsafeCell, thread};
    use std::sync::Arc;

    loom::model(|| {
        let number = Arc::new(UnsafeCell::new(0i32));
        let state = Arc::new(AtomicUsize::new(0));

        thread::spawn({
            let number = Arc::clone(&number);
            let state = Arc::clone(&state);

            move || {
                if state.load(Ordering::Relaxed) == 2 {
                    number.with(|ptr| assert_eq!(unsafe { *ptr }, 1));
                }
            }
        });

        thread::spawn(move || {
            match state.compare_exchange(0, 1, Ordering::SeqCst, Ordering::SeqCst) {
                Ok(0) => {
                    number.with_mut(|ptr| unsafe { *ptr = 1 });
                    state.store(2, Ordering::SeqCst);
                },
                _ => unreachable!()
            }

            fence(Ordering::SeqCst);
        });
    });
}
```

Thomasdezeeuw:
@Cassy343 I don't use Loom, but that's not exactly what I had in my. I was thinking the following.

 ```rust
 fn main() {
     use loom::{sync::atomic::{AtomicUsize, Ordering, fence}, cell::UnsafeCell, thread};
     use std::sync::Arc;
 
     loom::model(|| {
         let number = Arc::new(UnsafeCell::new(0i32));
         let state = Arc::new(AtomicUsize::new(0));
 
         thread::spawn({
             let number = Arc::clone(&number);
             let state = Arc::clone(&state);
 
             move || {
                 if state.load(Ordering::Relaxed) == 2 {
                     number.with(|ptr| assert_eq!(unsafe { *ptr }, 1));
                 }
             }
         });
 
         thread::spawn(move || {
             match state.compare_exchange(0, 1, Ordering::SeqCst, Ordering::SeqCst) {
                 Ok(0) => {
                     number.with_mut(|ptr| unsafe { *ptr = 1 });
                     fence(Ordering::SeqCst);  // Sync the write above.
                     state.store(2, Ordering::SeqCst);
                 },
                 _ => unreachable!()
             }
         });
     });
 }
 ```

Thomasdezeeuw:
But I think you guys are right and we still need something this proposed `fence` can sync with on the reading side (the log calls).

relufi:
> I'm not sure that this new ordering is valid, but does it really matter? The logger is set at most once, usually during startup, personally I don't think the performance concerns at that stage outweigh getting this wrong...

The impact is not limited to the startup phase, incurring unnecessary overhead every time the log!() is used, and SeqCst is a very expensive operation on some platforms.

```rust
pub fn logger() -> &'static dyn Log {
    if STATE.load(Ordering::SeqCst) != INITIALIZED {
```

relufi:
> I'm not sure that this new ordering is valid, but does it really matter? The logger is set at most once, usually during startup, personally I don't think the performance concerns at that stage outweigh getting this wrong...

The impact is not limited to the startup phase, incurring unnecessary overhead every time the log!() is used, and SeqCst is a very expensive operation on some platforms.

```rust
pub fn logger() -> &'static dyn Log {
    if STATE.load(Ordering::SeqCst) != INITIALIZED {
```

relufi:
There's a visibility issue here that requires Acquire. Please take a look at the [lib.rs](https://github.com/relufi/log/blob/8b5da221f8c53211a5534fe530adb53abc31b430/src/lib.rs#L1491)  I mentioned in #509 using fence to trigger an Acquire after set_logger_inner fails

relufi:
```suggestion
            while STATE.load(Ordering::Relaxed) == INITIALIZING {
```
There's a visibility issue here that requires Acquire. Please take a look at the [lib.rs](https://github.com/relufi/log/blob/8b5da221f8c53211a5534fe530adb53abc31b430/src/lib.rs#L1491)  I mentioned in #509 using fence to trigger an Acquire after set_logger_inner fails

KodrAus:
Thanks for the PR @Cassy343! I think one of your earlier points about not using a home-grown once cell for this is pretty valid, and once [`OnceLock` stabilizes](https://github.com/rust-lang/rust/issues/74465) we should move to it. In the meantime, synchronizing on the global logger is almost the only thing this library actually does, so it seems fair that we should try do a good job of it. Maybe we could move this synchronization code into its own sub-module and add a `loom` test for it?

NobodyXu:
`OnceLock` is stablised in 1.70 and the msrv for this crate is 1.60, so it looks like this would take another debian release to happen.

NobodyXu:
@KodrAus Maybe it's a good idea to use `once_cell` as a dep?

KodrAus:
@NobodyXu Ah we don’t actually tie our MSRV to what’s in Debian. We just lined them up this time because there happened to be a release coming up, but I don’t expect us to stay tied to 1.60 for the next few years.

We avoid adding any required dependencies to `log` so I think can’t pull in `once_cell`. Once some time has passed we’ll move to 1.70 and be able to use `OnceLock`.

NobodyXu:
> @NobodyXu Ah we don’t actually tie our MSRV to what’s in Debian. We just lined them up this time because there happened to be a release coming up, but I don’t expect us to stay tied to 1.60 for the next few years.

Wouldn't that break users who compile their programs on debian stable or make it harder for debian to package rust programs that depends on log?

> We avoid adding any required dependencies to `log` so I think can’t pull in `once_cell`. Once some time has passed we’ll move to 1.70 and be able to use `OnceLock`.

If you don't want to tie log msrv bump to debian release, then I suppose it will be safe to do it in 6-9 months of time?

Thomasdezeeuw:
> We avoid adding any required dependencies to `log` so I think can’t pull in `once_cell`. Once some time has passed we’ll move to 1.70 and be able to use `OnceLock`.

Problem is the `OnceLock` isn't available without std lib, otherwise we could have used `std::sync::Once`.

Thomasdezeeuw:
In #610 we switched to using `Acquire` for the `logger` function (i.e. one called for every log message).

We could still accept the changes here to relax the orderings in the setup functions (`set_logger` etc.). https://github.com/rust-lang/rust/pull/122729 I think set a pretty clear presented recently.

KodrAus:
Well, that's what I get for trying to use GitHub's UI to merge branches.

KodrAus:
I've re-opened this as #632 so will close this PR. The discussion is all linked so we can find our way back to it in the future if needed.

PR #524, opened 2022-08-15
Make `Level` and `LevelFilter` usable in clap

maxwase:
Allow log level enums to be used in clap.

Typical usage: 

```rust
#[derive(clap::Parser)]
struct Opt {
    #[clap(short, long, default_value = "info")]
    log_level: LevelFilter,
}
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ exclude = ["rfcs/**/*"]
 build = "build.rs"
 
 [package.metadata.docs.rs]
-features = ["std", "serde", "kv_unstable_std", "kv_unstable_sval", "kv_unstable_serde"]
+features = ["std", "serde", "clap", "kv_unstable_std", "kv_unstable_sval", "kv_unstable_serde"]
 
 [[test]]
 name = "filters"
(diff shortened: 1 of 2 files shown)


maxwase:
clap's MSRV is 1.60.0, this crate has 1.31.0, is it even possible to use this under feature? Or MSRV should be the same across all features?

Thomasdezeeuw:
I don't log should depend on clap, why can't this be added to the clap crate?

maxwase:
> I don't log should depend on clap, why can't this be added to the clap crate?

I thought about it the same way as it was with serde.

Thomasdezeeuw:
> > I don't log should depend on clap, why can't this be added to the clap crate?
> 
> I thought about it the same way as it was with serde.

We optionally depend on serde for key-value deserialising. Supporting clap is something that is completely unrelated to logging.

maxwase:
> > > I don't log should depend on clap, why can't this be added to the clap crate?
> > 
> > 
> > I thought about it the same way as it was with serde.
> 
> We optionally depend on serde for key-value deserialising. Supporting clap is something that is completely unrelated to logging.

Yeah, I think I misunderstood that logic, I guess this PR does not make sense from now

PR #543, opened 2023-01-06
Remove build.rs file

### LINKED ISSUE #489: Remove `build.rs`

GuillaumeGomez:
Fixes #489.

This is another take on #490. Since Rust 1.60, `cfg(target_has_atomic = "...")` is stable so we can now use it. However, it'll change the MSRV to 1.60 as well.

I used `target_has_atomic = "ptr"` (and not `target_has_atomic = "8"` for example) because that's what is used in the `std` for `AtomicUsize`. You can see it in the macro declaring the `AtomicUsize` [here](https://doc.rust-lang.org/1.66.0/src/core/sync/atomic.rs.html#2906).

cc @nnethercote (since you're the one who suggested this change)

EDIT: Just saw that this solution was mentioned by @KodrAus in the linked issue.


GuillaumeGomez:
Once it's figured out, we can then decide what to do with this PR. No hurry. ;)

KodrAus:
`1.60.0` is quite a jump, but is still just over 12 months old. I personally think the removal of a `build.rs` makes it worth it. I don't think it's worth doing a minor bump for this purely because we couldn't shim an `0.5.x` build in `0.4.x` without also bumping the MSRV of `0.4.x` anyways, and wouldn't want to issue a breaking release for `log` that didn't automatically forward `0.4.x` through `0.5.x`.

KodrAus (approved):
Thanks @GuillaumeGomez!

This looks good to me. If @Thomasdezeeuw is also happy I say we should get it merged and look at putting a release together 🙂

Thomasdezeeuw (approved):

PR #544, opened 2023-01-06
Add `set_max_level_racy` and gate `set_max_level`

### LINKED ISSUE #534: `AtomicUsize` has an unsound `Sync` impl
The comment on the `Sync` impl for `AtomicUsize` is:

```rust
// Any platform without atomics is unlikely to have multiple cores, so
// writing via Cell will not be a race condition.
```

While this may be true in practice, it's still technically incorrect and could open the door for bad codegen. Is there some background on why this impl is required even for platforms that don't support atomics?

djkoloski:
Calling `set_max_level` can result in a race condition on platforms that don't have an atomic compare and swap implementation. This gates `set_max_level` behind `#[cfg(atomic_cas)]` and adds a racy alternative that can be called in these situations. This mirrors the approach for `set_logger`.

Fixes #534

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1215,7 +1215,35 @@ where
 ///
 /// Note that `Trace` is the maximum level, because it provides the maximum amount of detail in the emitted logs.
 #[inline]
+#[cfg(atomic_cas)]
 pub fn set_max_level(level: LevelFilter) {
+    MAX_LOG_LEVEL_FILTER.store(level as usize, Ordering::Relaxed);
+}
+
+/// A thread-unsafe version of [`set_max_level`].
+///
+/// This function is available on all platforms, even those that do not have
+/// support for atomics that is needed by [`set_max_level`].
+///
+/// In almost all cases, [`set_max_level`] should be preferred.
+///
+/// # Safety
+///
+/// This function is only safe to call when no other level setting function is
+/// called while this function still executes.
+///
+/// This can be upheld by (for example) making sure that **there are no other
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1215,10 +1215,35 @@ where
...
+/// This can be upheld by (for example) making sure that **there are no other
+/// threads**, and (on embedded) that **interrupts are disabled**.
+///
+/// Is is safe to use all other logging functions while this function runs
+/// (including all logging macros).
+///
+/// [`set_max_level`]: fn.set_max_level.html
+#[inline]
+pub unsafe fn set_max_level_racy(level: LevelFilter) {
+    MAX_LOG_LEVEL_FILTER.store(level as usize, Ordering::Relaxed);
```
Thomasdezeeuw, line 1244: This is the same as the other function?

djkoloski: Yes, but the type of `MAX_LOG_LEVEL_FILTER` is different (`core::sync::AtomicUsize` vs `log::AtomicUsize`). When a platform doesn't support `atomic_cas`, the underlying implementation uses a `Cell<usize>` which can race with interrupts.

Thomasdezeeuw: Ah, that makes sense. Then maybe a note pointing that out would be worth it?

djkoloski: I think that's a good idea, added a comment.

Thomasdezeeuw (approved):

PR #548, opened 2023-03-28
Add the structured-logger crate to the README

zensh:
[structured-logger](https://docs.rs/structured-logger/latest/structured_logger/) is a logging implementation for the log crate that logs structured values either synchronous or asynchronous, as JSON, CBOR, or any other format, into a file, stderr, stdout, or any other destination.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,7 @@ There are many available implementations to choose from, here are some options:
     * [`flexi_logger`](https://docs.rs/flexi_logger/*/flexi_logger/)
     * [`call_logger`](https://docs.rs/call_logger/*/call_logger/)
     * [`std-logger`](https://docs.rs/std-logger/*/std_logger/)
+    * [`structured-logger`](https://docs.rs/structured-logger/latest/structured_logger/)
 * Complex configurable frameworks:
     * [`log4rs`](https://docs.rs/log4rs/*/log4rs/)
     * [`fern`](https://docs.rs/fern/*/fern/)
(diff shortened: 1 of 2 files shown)


zensh:
I have updated [structured-logger](https://docs.rs/structured-logger/latest/structured_logger/) https://github.com/iorust/structured-logger/commit/c421ecc31a4efe3bada9740fa4d6f2f448535fc3 for MSRV checking

zensh:
> I have updated [structured-logger](https://docs.rs/structured-logger/latest/structured_logger/) [iorust/structured-logger@c421ecc](https://github.com/iorust/structured-logger/commit/c421ecc31a4efe3bada9740fa4d6f2f448535fc3) for MSRV checking

Oh! it come from syn-2.0.11

JohnTitor:
I don't think it's related to CI failure here. The failure is due to syn v2 which is a transitive dependency of this crate.

JohnTitor:
On the other hand, someone may not call your crate as "most popular one" because it's pretty new.
The problem is that we don't have any solid policy around the list and it could be quite long if we accept crates unconditionally.
Maybe we could link to https://crates.io/keywords/log and/or https://crates.io/keywords/logger instead (one of this approach's cons is that this list can contain unrelated crates). I'd like to hear more opinions, cc @rust-lang/crate-maintainers

KodrAus:
I've opened #551 to discuss the future of this list. Either #550 or #543 should deal with that MSRV build failure too. It looks like we've got a merge conflict, but once that's sorted I'll go ahead and merge this one, as it's in-line with our current treatment of the integrations list.

Thanks for working on this @zensh! 🙂

PR #550, opened 2023-04-10
Use stable internals for key-value API

KodrAus:
For #149

There are two (formally) unstable libraries that we depend on in the key-value API:

- [`value-bag`](https://github.com/sval-rs/value-bag): is the backing structure for `Value` that lets us capture and erase value types. We don't want to publicly expose it.
- [`sval`](https://github.com/sval-rs/sval): is an alternative serialisation framework to `serde` with an alloc-free object-safe API. It's an optional public dependency, like `serde`.

Both of these libraries now have stable releases so this PR moves to them. This was one of the blockers to stabilising the key-value API in #436. It's also a good excuse to get another release of `log` out, since we haven't made one in a while.


Cargo.toml
```diff
@@ -48,18 +48,20 @@ std = []
...
-kv_unstable_sval = ["kv_unstable", "value-bag/sval", "sval"]
+kv_unstable_sval = ["kv_unstable", "value-bag/sval", "sval", "sval_ref"]
 kv_unstable_std = ["std", "kv_unstable", "value-bag/error"]
 kv_unstable_serde = ["kv_unstable_std", "value-bag/serde", "serde"]
 
 [dependencies]
 serde = { version = "1.0", optional = true, default-features = false }
-sval = { version = "=1.0.0-alpha.5", optional = true, default-features = false }
-value-bag = { version = "=1.0.0-alpha.9", optional = true, default-features = false }
+sval = { version = "2.1", optional = true, default-features = false }
```
KodrAus, line 57: I ended up skipping a stable `1.x` for this library in the end, because the API changed fairly significantly between beta releases.

src/kv/key.rs
```diff
@@ -48,6 +48,12 @@ impl<'k> Key<'k> {
     pub fn as_str(&self) -> &str {
         self.key
     }
+
+    /// Try get a string borrowed for the `'k` lifetime from this key.
+    pub fn to_borrowed_str(&self) -> Option<&'k str> {
```
KodrAus, line 53: This returns an `Option` even though it's currently always `Some` so that we've got room to make the internals something like a `Cow<'k, str>`.

Thomasdezeeuw (approved):

PR #553, opened 2023-04-11
Change wording of list of implementations

Thomasdezeeuw:
Updates #551.

Issue #551, opened 2023-04-10
What to do about the integration list in the README


KodrAus:
In https://github.com/rust-lang/log/pull/548#issuecomment-1486869928 @JohnTitor brought up the broader question of what to do with the list of integrations listed in the README. They're currently presented as "the most popular" implementations out there. Right now the list isn't really curated that way; if there's a new crate that integrates with `log` then we add it to the list.

The function the list serves is giving users a kick-start to get actual data out of `log`, which on its own simply pipes your events into oblivion.

Alternative strategies to achieving that function are:

- Linking a Cargo keyword for `log` or `logger` as mentioned in the linked comment.
- Updating the wording of the README to not suggest the list is curated in any way besides listing libraries that integrate with `log`. It's up to end-users to evaluate the suitability of these libraries for their own needs.

NobodyXu:
Would it be a good idea to mention `tracing_log` in the list, for it enables forwarding `tracing` to `log` or visa versa?

Thomasdezeeuw:
Maybe we should just change the wording? For example from:

> There are many available implementations to choose from, here are some of the most popular ones:

To:

> There are many available implementations to choose from, here are some options:

KodrAus:
@Thomasdezeeuw that sounds good to me.

Thomasdezeeuw:
@KodrAus any further changes we want to make?

KodrAus:
@Thomasdezeeuw I think it's all ok now. Are you happy for this issue to be closed?

PR #554, opened 2023-04-11
Add std-logger to list of implementations

Thomasdezeeuw:
While making https://github.com/rust-lang/log/pull/553 I released my own library isn't even in the list, so I though I might as well add it. Unless we decide to remove all of them in https://github.com/rust-lang/log/issues/551.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,7 @@ There are many available implementations to choose from, here are some of the mo
     * [`stderrlog`](https://docs.rs/stderrlog/*/stderrlog/)
     * [`flexi_logger`](https://docs.rs/flexi_logger/*/flexi_logger/)
     * [`call_logger`](https://docs.rs/call_logger/*/call_logger/)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -68,6 +68,7 @@ There are many available implementations to choose from, here are some of the mo
     * [`stderrlog`](https://docs.rs/stderrlog/*/stderrlog/)
     * [`flexi_logger`](https://docs.rs/flexi_logger/*/flexi_logger/)
     * [`call_logger`](https://docs.rs/call_logger/*/call_logger/)
+    * [`std-logger`](https://docs.rs/std-logger/*/std-logger/)
```
KodrAus, line 71:
```suggestion
    * [`std-logger`](https://docs.rs/std-logger/*/std_logger/)
```

KodrAus: It looks like `docs.rs` interprets the `-` in `std-logger` after the version pattern incorrectly and gives you an error page. This link seems to work correctly.

KodrAus (approved):

PR #555, opened 2023-04-13
Use target_has_atomic instead of the old atomic_cas cfg

KodrAus:
Fixes up the build after #544 since we refactored our atomic cfgs in #543.


KodrAus:
If you’ll forgive the post-merge review @Thomasdezeeuw I might push this one through to get our main build back to green 🙂

PR #557, opened 2023-04-15
Put MSRV into Cargo.toml

est31:
MSRV was increased by #543.

Also remove it from clippy.toml as clippy
reads it from Cargo.toml since 1.64.
CI only runs clippy from latest stable and not
from the MSRV so this shouldn't be a problem.


KodrAus (approved):
Thanks @est31!

PR #561, opened 2023-05-26
[doc] src/lib.rs : prefix an unused variable with an underscore

OccupyMars2025:
unused variable: `metadata`,  according to rustc, 
if this is intentional, prefix it with an underscore: `_metadata`


OccupyMars2025:
If the maintainers think it is unnecessary,  just close this pull request.

KodrAus (approved):
Thanks @OccupyMars2025! This looks good to me.

Issue #564, opened 2023-05-27
Building failed on x86_64-pc-windows-msvc nightly


DarcJC:
See https://github.com/rust-lang/log/actions/runs/5099423637/jobs/9167113151

Using  v1.71.0 nightly (1a5f8bce7 2023-05-26).
log = "0.4.17"

DarcJC:
`cargo run`'s output: 

```text
error: to use a constant of type `TypeId` in a pattern, `TypeId` must be annotated with `#[derive(PartialEq, Eq)]`
  --> C:\Users\i\.cargo\registry\src\index.crates.io-6f17d22bba15001f\value-bag-1.0.0-alpha.9\src\internal\cast\primitive.rs:43:33
   |
43 |                                   $const_ident => |v| Some(Internal::from(unsafe { &*(*v as *const Self as *const $ty) })),
   |                                   ^^^^^^^^^^^^
...
71 | /         to_internal![
72 | |             usize: (USIZE, OPTION_USIZE),
73 | |             u8: (U8, OPTION_U8),
74 | |             u16: (U16, OPTION_U16),
...  |
96 | |             String: (STRING, OPTION_STRING),
97 | |         ];
   | |_________- in this macro invocation
   |
   = note: the traits must be derived, manual `impl`s are not sufficient
   = note: see https://doc.rust-lang.org/stable/std/marker/trait.StructuralEq.html for details
   = note: this error originates in the macro `to_internal` (in Nightly builds, run with -Z macro-backtrace for more info)

error: to use a constant of type `TypeId` in a pattern, `TypeId` must be annotated with `#[derive(PartialEq, Eq)]`
  --> C:\Users\i\.cargo\registry\src\index.crates.io-6f17d22bba15001f\value-bag-1.0.0-alpha.9\src\internal\cast\primitive.rs:46:33
   |
46 |                                   $option_ident => |v| Some({
   |                                   ^^^^^^^^^^^^^
...
71 | /         to_internal![
72 | |             usize: (USIZE, OPTION_USIZE),
73 | |             u8: (U8, OPTION_U8),
74 | |             u16: (U16, OPTION_U16),
...  |
96 | |             String: (STRING, OPTION_STRING),
97 | |         ];
   | |_________- in this macro invocation
   |
   = note: the traits must be derived, manual `impl`s are not sufficient
   = note: see https://doc.rust-lang.org/stable/std/marker/trait.StructuralEq.html for details
[...]

PR #565, opened 2023-05-28
Revert MSRV bump for patch release

KodrAus:
For #564

This is already fixed on the `master` branch, but we committed not to release that until the next stable version of Debian. In the meantime, this reverts #543 so we can get a patch out that still targets `rustc` `1.31.0`. Once that is released, I'll revert this commit so #552 will still include it.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -127,11 +127,8 @@ jobs:
       - uses: actions/checkout@master
       - name: Install Rust
         run: |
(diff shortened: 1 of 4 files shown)


KodrAus:
cc @Thomasdezeeuw @JohnTitor

I'll keep this one rolling through so we can get a patch out quickly, but will make sure to CC on the revert PR as well 🙂

PR #568, opened 2023-07-03
Remove rustversion dev-dependency

Thomasdezeeuw:
Since the MSRV is now 1.60.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,3 +64,11 @@ serde_test = "1.0"
 sval = { version = "2.1" }
 sval_derive = { version = "2.1" }
 value-bag = { version = "1.4", features = ["test"] }
+
+# NOTE: log doesn't actually depent on this crate. However our dependencies,
+# serde and sval, dependent on version 1.0 of the crate, which has problem fixed
+# in 1.0.60, specifically in the following commit
+# https://github.com/dtolnay/proc-macro2/commit/e31d61910049e097afdd3d27c37786309082bdcb.
+# By defining the crate as direct dependency we can increase it's minimal
+# version making the minimal (crate) version CI happy.
+proc-macro2 = { version = "1.0.63", default-features = false }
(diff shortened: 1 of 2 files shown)


Thomasdezeeuw:
The minimal versions issue should be resolved by updating to proc-macro2  v1.0.60 (specifically in https://github.com/dtolnay/proc-macro2/commit/e31d61910049e097afdd3d27c37786309082bdcb). But I think this has to happen in serde_derive?

Thomasdezeeuw:
> The minimal versions issue should be resolved by updating to proc-macro2 v1.0.60 (specifically in [dtolnay/proc-macro2@e31d619](https://github.com/dtolnay/proc-macro2/commit/e31d61910049e097afdd3d27c37786309082bdcb)). But I think this has to happen in serde_derive?

Oh both serde_derive and sval_derive define `proc-macro2 = "1.0"` (https://github.com/serde-rs/serde/blob/015e39776fcd951d208d7fbb5f4a4de4efb4445f/serde_derive/Cargo.toml#L25 and https://github.com/sval-rs/sval/blob/3cbda0f19ac37260b44f16233944529f04666e05/derive/Cargo.toml#L25), so this is going to be a bit annoying.

KodrAus:
Ah at first I was reading _minimal versions_ as _MSRV_ and wasn’t following what you meant. Now I get it.

I’m happy to add minimal versions to CI in `sval` and bump `proc-macro2`, but it might be easier for us to simply remove those derive dependencies in `log`. It might get a little tedious, depending on where we use them though.

KodrAus:
Alternatively we could add `proc-macro2` as a dev dependency with a comment on why it’s there. That might be the least painful.

Thomasdezeeuw:
> Alternatively we could add `proc-macro2` as a dev dependency with a comment on why it’s there. That might be the least painful.

I've went with this option as it's the easiest, even though it's a little hacky.

KodrAus:
I’ll run through later and see if I can yank the derive crates out as dependencies altogether.

KodrAus (approved):
This looks good to me!

PR #569, opened 2023-07-09
Binary size optimization experiment

EFanZh:
Experiment for some ideas I have on binary size optimization.

### DIFF
--- a/src/__private_api.rs
+++ b/src/__private_api.rs
@@ -1,0 +1,321 @@
+use self::sealed::{LogArgs, LogKVs, LogLevel, LogTarget};
+use crate::{Level, LevelFilter, Metadata, Record};
+use std::cmp::Ordering;
+pub use std::convert::identity;
+use std::fmt::Arguments;
+use std::option::Option;
+pub use std::primitive::str;
+pub use std::{file, format_args, line, module_path, stringify};
+
+#[cfg(feature = "kv_unstable")]
+pub type LogKvValue<'a> = dyn crate::kv::value::ToValue + 'a;
+
+#[cfg(not(feature = "kv_unstable"))]
+pub type LogKvValue<'a> = str;
+
+mod sealed {
+    use crate::Level;
+    use std::fmt::Arguments;
+
+    pub trait LogLevel {
+        fn into_log_level(self) -> Level;
+    }
+
+    pub trait LogArgs {
+        fn with(self, f: impl FnOnce(Arguments));
+    }
+
+    pub trait LogTarget {
+        fn with(self, module_path: &'static str, f: impl FnOnce(&str));
+    }
+
+    pub trait LogKVs {
+        fn with(self, f: impl FnOnce(&[(&str, &super::LogKvValue)]));
+    }
+}
+
+// `LogLevel`.
+
+impl LogLevel for Level {
+    #[inline]
+    fn into_log_level(self) -> Level {
+        self
+    }
+}
+
+macro_rules! define_static_levels {
+    ($($ty:ident => $lvl:ident,)*) => {
+        $(
+            #[derive(Debug)]
+            pub struct $ty;
+
+            impl LogLevel for $ty {
+                #[inline]
+                fn into_log_level(self) -> Level {
+                    Level::$lvl
+                }
+            }
+
+            impl PartialEq<LevelFilter> for $ty {
+                #[inline]
+                fn eq(&self, other: &LevelFilter) -> bool {
+                    Level::$lvl.eq(other)
+                }
+            }
+
+            impl PartialOrd<LevelFilter> for $ty {
+                #[inline]
+                fn partial_cmp(&self, other: &LevelFilter) -> Option<Ordering> {
+                    Level::$lvl.partial_cmp(other)
+                }
+
+                #[inline]
+                fn lt(&self, other: &LevelFilter) -> bool {
+                    Level::$lvl.lt(other)
+                }
+
+                #[inline]
+                fn le(&self, other: &LevelFilter) -> bool {
+                    Level::$lvl.le(other)
+                }
+
+                #[inline]
+                fn gt(&self, other: &LevelFilter) -> bool {
+                    Level::$lvl.gt(other)
+                }
+
+                #[inline]
+                fn ge(&self, other: &LevelFilter) -> bool {
+                    Level::$lvl.ge(other)
+                }
+            }
+        )*
+    };
+}
+
+define_static_levels![
+    StaticLevelError => Error,
+    StaticLevelWarn => Warn,
+    StaticLevelInfo => Info,
+    StaticLevelDebug => Debug,
+    StaticLevelTrace => Trace,
+];
+
+// `LogArgs`.
+
+impl LogArgs for &str {
+    #[inline]
+    fn with(self, f: impl FnOnce(Arguments)) {
+        f(format_args!("{self}"))
+    }
+}
+
+impl LogArgs for Arguments<'_> {
+    fn with(self, f: impl FnOnce(Arguments)) {
+        f(self)
+    }
+}
+
+// `LogTarget`.
+
+impl LogTarget for &str {
+    #[inline]
+    fn with(self, _module_path: &'static str, f: impl FnOnce(&str)) {
+        f(self)
+    }
+}
+
+#[derive(Debug)]
+pub struct TargetIsModulePath;
+
+impl LogTarget for TargetIsModulePath {
+    #[inline]
+    fn with(self, module_path: &'static str, f: impl FnOnce(&str)) {
+        f(module_path)
+    }
+}
+
+// `LogKVs`.
+
+impl LogKVs for &[(&str, &LogKvValue<'_>)] {
+    #[inline]
+    fn with(self, f: impl FnOnce(&[(&str, &LogKvValue)])) {
+        f(self)
+    }
+}
+
+#[derive(Debug)]
+pub struct EmptyKVs;
+
+impl LogKVs for EmptyKVs {
+    #[inline]
+    fn with(self, f: impl FnOnce(&[(&str, &LogKvValue)])) {
+        f(&[])
+    }
+}
+
+// Log functions.
+
+fn log_0(
+    &(module_path, file): &'static (&'static str, &'static str),
+    line: u32,
+    level: Level,
+    args: Arguments,
+    target: &str,
+    kvs: &[(&str, &LogKvValue)],
+) {
+    #[cfg(not(feature = "kv_unstable"))]
+    if !kvs.is_empty() {
+        panic!(
+            "key-value support is experimental and must be enabled using the `kv_unstable` feature"
+        );
+    }
+
+    let mut builder = Record::builder();
+
+    builder
+        .args(args)
+        .level(level)
+        .target(target)
+        .module_path_static(Some(module_path))
+        .file_static(Some(file))
+        .line(Some(line));
+
+    #[cfg(feature = "kv_unstable")]
+    builder.key_values(&kvs);
+
+    crate::logger().log(&builder.build());
+}
+
+fn log_1<K>(
+    module_path_and_file: &'static (&'static str, &'static str),
+    line: u32,
+    level: Level,
+    args: Arguments,
+    target: &str,
+    kvs: K,
+) where
+    K: LogKVs,
+{
+    kvs.with(|kvs| log_0(module_path_and_file, line, level, args, target, kvs));
+}
+
+fn log_2<T, K>(
+    module_path_and_file: &'static (&'static str, &'static str),
+    line: u32,
+    level: Level,
+    args: Arguments,
+    target: T,
+    kvs: K,
+) where
+    T: LogTarget,
+    K: LogKVs,
+{
+    target.with(module_path_and_file.0, |target| {
+        log_1(module_path_and_file, line, level, args, target, kvs)
+    });
+}
+
+pub fn log_3<A, T, K>(
+    module_path_and_file: &'static (&'static str, &'static str),
+    line: u32,
+    level: Level,
+    args: A,
+    target: T,
+    kvs: K,
+) where
+    A: LogArgs,
+    T: LogTarget,
+    K: LogKVs,
+{
+    args.with(|args| log_2(module_path_and_file, line, level, args, target, kvs))
+}
+
+pub fn log<L, A, T, K>(
+    module_path_and_file: &'static (&'static str, &'static str),
+    line: u32,
+    level: L,
+    args: A,
+    target: T,
+    kvs: K,
+) where
+    L: LogLevel,
+    A: LogArgs,
+    T: LogTarget,
+    K: LogKVs,
+{
+    log_3(
+        module_path_and_file,
+        line,
+        level.into_log_level(),
+        args,
+        target,
+        kvs,
+    )
+}
+
+pub fn enabled(level: Level, target: &str) -> bool {
+    crate::logger().enabled(&Metadata::builder().level(level).target(target).build())
+}
+
+pub const fn is_literal(s: &str) -> bool {
+    let s = s.as_bytes();
+    let n = s.len();
+    let mut i = 0;
+
+    while i < n {
+        if matches!(s[i], b'{' | b'}') {
+            return false;
+        }
+
+        i += 1;
+    }
+
+    true
+}
+
+pub fn unused() -> usize {
+    fn for_all_kvs<L, A, T>() -> usize
+    where
+        L: LogLevel,
+        A: LogArgs,
+        T: LogTarget,
+    {
+        type K0 = &'static [(&'static str, &'static LogKvValue<'static>)];
+        type K1 = EmptyKVs;
+
+        log::<L, A, T, K0> as usize ^ log::<L, A, T, K1> as usize
+    }
+
+    fn for_all_targets<L, A>() -> usize
+    where
+        L: LogLevel,
+        A: LogArgs,
+    {
+        type T0 = &'static str;
+        type T1 = TargetIsModulePath;
+
+        for_all_kvs::<L, A, T0>() ^ for_all_kvs::<L, A, T1>()
+    }
+
+    fn for_all_arguments<L>() -> usize
+    where
+        L: LogLevel,
+    {
+        type A0 = &'static str;
+        type A1 = Arguments<'static>;
+
+        for_all_targets::<L, A0>() ^ for_all_targets::<L, A1>()
+    }
+
+    [
+        for_all_arguments::<Level>(),
+        for_all_arguments::<StaticLevelError>(),
+        for_all_arguments::<StaticLevelWarn>(),
+        for_all_arguments::<StaticLevelInfo>(),
+        for_all_arguments::<StaticLevelDebug>(),
+        for_all_arguments::<StaticLevelTrace>(),
+    ]
+    .iter()
+    .fold(0, |x, y| x ^ y)
+}
--- a/src/__private_api.rs
+++ b/src/__private_api.rs
@@ -26,11 +26,11 @@ mod sealed {
     }
 
     pub trait LogTarget {
-        fn with(self, module_path: &'static str, f: impl FnOnce(&str));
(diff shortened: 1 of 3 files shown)


src/macros.rs
```diff
@@ -8,6 +8,66 @@
...
 // except according to those terms.
 
+#[doc(hidden)]
+#[macro_export]
+macro_rules! __internal_log {
+    (@helper $target:expr, $lvl:expr, $kvs:expr, $($arg:tt)+) => ({
+        let lvl = $lvl;
+        if lvl <= $crate::STATIC_MAX_LEVEL && lvl <= $crate::max_level() {
+            match (::std::format_args!($($arg)+), $target, $kvs) {
+                (args, target, kvs) => if let ::std::option::Option::Some(literal) = ::std::fmt::Arguments::as_str(&args) {
```
NobodyXu, line 18: I don't think this is necessary, since it is turned into an `fmt::Argument<'_>` again when constructing `Record<'_>`.

I think this actually prevents the `Log` implementation to take advantage of `Argument::as_str`.

NobodyXu:
Is this PR trying to take advantages of https://github.com/rust-lang/rust/pull/109999/ ?

src/macros.rs
EFanZh: The idea is that passing an `Arguments` value is more costly than passing an `str` value: <https://godbolt.org/z/ehPcoqsov>. I expect an application may contain a lot of logs that only contains string literals, so If we can reduce the size of each log, the total binary size could been reduced. Essentially, this implementation tries to bring back an old optimization that is reverted by #446.

EFanZh:
> Is this PR trying to take advantages of [rust-lang/rust#109999](https://github.com/rust-lang/rust/pull/109999) ?

Not exactly. This PR intends to minimize code generated by macro expansion. This is done by moving as much code logic as possible into functions that can be shared by each macro call.

src/macros.rs
NobodyXu: You can pass a reference to Arguments, then dereference it when building Records.
It implements Copy, so you can avoid copying it in each function while preserving the ability to take advantages of Arguments::as_str

NobodyXu:
> > Is this PR trying to take advantages of [rust-lang/rust#109999](https://github.com/rust-lang/rust/pull/109999) ?
> 
> Not exactly. This PR intends to minimize code generated by macro expansion. This is done by moving as much code logic as possible into functions that can be shared by each macro call.

Thanks for explanation.

Thomasdezeeuw (commented):
Note that we use to have a special case for literals: https://github.com/rust-lang/log/pull/366, but it was removed in https://github.com/rust-lang/log/commit/2d1a477f5a711d7aa1bbf0ca0c2a01b02a236763.

src/macros.rs
EFanZh: Aside from copying, building an `Argument` object also has costs: <https://godbolt.org/z/a1eqr4bq5>.

EFanZh: It seems that passing `Arguments` and padding `&Arguments` have the same cost: <https://godbolt.org/z/Wz35xYrsr>.

EFanZh:
It seems that using "opt-level=z", the compiler won’t inline `Arguments::as_str` if there are multiple calls to `Arguments::as_str`: <https://godbolt.org/z/rzsG3fYEK>. So both `None` and `Some` branches are kept in the final binary, which will cause binary size to grow.

Maybe marking `Arguments::as_str` as `#[inline(always)]` could help, but for now, I think I’ll have to give up specializing literal arguments.

NobodyXu:
> Maybe marking Arguments::as_str as #[inline(always)] could help, but for now, I think I’ll have to give up specializing literal arguments.

You could just pass `&fmt::Argument<'_>` instead of pass it by-value, it should remove the generated code for copying it.

EFanZh:
@NobodyXu: I have tested the effect of passing `&Arguments`, and the result is not much different than passing `Arguments` directly. Maybe it’s because Rust already passes large object as references: <https://godbolt.org/z/Wz35xYrsr>. You can see in the example, passing `Arguments` and passing `&Arguments` generate the same assembly code.

NobodyXu:
> @NobodyXu: I have tested the effect of passing `&Arguments`, and the result is not much different than passing `Arguments` directly. Maybe it’s because Rust already passes large object as references: https://godbolt.org/z/Wz35xYrsr. You can see in the example, passing `Arguments` and passing `&Arguments` generate the same assembly code.

Your linked code does not actually use the Arguments, so the compiler might omits the copy requires to happen.

In your PR, you passes the Arguments to log2, log1, log0 by value, that would require copying Arguments between stacks unless inlined.

EFanZh:
> > @NobodyXu: I have tested the effect of passing `&Arguments`, and the result is not much different than passing `Arguments` directly. Maybe it’s because Rust already passes large object as references: https://godbolt.org/z/Wz35xYrsr. You can see in the example, passing `Arguments` and passing `&Arguments` generate the same assembly code.
> 
> Your linked code does not actually use the Arguments, so the compiler might omits the copy requires to happen.
> 
> In your PR, you passes the Arguments to log2, log1, log0 by value, that would require copying Arguments between stacks unless inlined.

`log0`, `log1`, `log` can only have a fixed number of instances, which means they can only contribute to at most a constant size to the final binary, while the majority binary size comes from codes generated by macro calls, since there could be a lot of log calls for a large project. So optimizing function implementations should not matter too much. But still, I will compare different ways to pass arguments, and choose the best one.

EFanZh:
OK, I think this is ready for the next step. This PR is just a proof of concept of some ideas I have, and should not be merged as is. I have tested this PR on my personal project and gained about 2.4% binary size reduction, which is significant to me. Now I want to implement these ideas in the `log` crate. Since this PR contains several optimization ideas, and I am not certain that all of them have positive effect on the binary size, so the best way maybe to discuss each idea, and decide whether it is reasonable. Finally, we may implement each acceptable idea individually.

1. Pass `&str` instead of `Arguments` if the message is a string literal.
   - I used a const function to check whether the format string can be treated as a string literal. And the condition is whether it contains `'{'` or `'}'` character, see the `is_literal` function.
2. If log level is known at compile time, pass it statically.
3. If no key-value is provided, pass the information statically.
4. If target is the same as module path, pass the information statically.
5. Pass module path and file as `&'static (&'static str, &'static str)`, so that logs from a single file can share the same argument.
6. If the kvs and format arguments do not have side effects, move the level filter check inside the function. (Needs further investigation, not implemented in this PR)
7. If the final implementation is done using generic functions like this PR, make sure all generic function instances are instantiated in this crate (the `unused` function in this PR), so that downstream crates can utilize them by enabling [`share-generics`](https://github.com/rust-lang/rust/pull/48779).

Also, is there a known open source project that uses `log` extensively, so I can test the effect of my optimization on it? I only have a private project to test. I think my optimization will need a reproducible result.

NobodyXu:
> Also, is there a known open source project that uses `log` extensively, so I can test the effect of my optimization on it? I only have a private project to test. I think my optimization will need a reproducible result.

In [`cargo-binstall`](https://github.com/cargo-bins/cargo-binstall), `log` is a dependency that is heavily used:

<details>

```
log v0.4.19
├── async_zip v0.0.15
│   └── binstalk-downloader v0.6.0 (/Users/nobodyxu/Dev/cargo-binstall/crates/binstalk-downloader)
│       └── binstalk v0.13.0 (/Users/nobodyxu/Dev/cargo-binstall/crates/binstalk)
│           └── cargo-binstall v1.0.0 (/Users/nobodyxu/Dev/cargo-binstall/crates/bin)
├── cargo-binstall v1.0.0 (/Users/nobodyxu/Dev/cargo-binstall/crates/bin)
├── gix v0.48.0
│   └── binstalk v0.13.0 (/Users/nobodyxu/Dev/cargo-binstall/crates/binstalk) (*)
├── gix-attributes v0.14.1
│   ├── gix v0.48.0 (*)
│   └── gix-worktree v0.21.1
│       └── gix v0.48.0 (*)
├── gix-config v0.25.1
│   └── gix v0.48.0 (*)
├── guess_host_triple v0.1.3
│   └── detect-targets v0.1.8 (/Users/nobodyxu/Dev/cargo-binstall/crates/detect-targets)
│       └── binstalk v0.13.0 (/Users/nobodyxu/Dev/cargo-binstall/crates/binstalk) (*)
│       [dev-dependencies]
│       └── binstalk-manifests v0.6.0 (/Users/nobodyxu/Dev/cargo-binstall/crates/binstalk-manifests)
│           └── cargo-binstall v1.0.0 (/Users/nobodyxu/Dev/cargo-binstall/crates/bin)
├── reqwest v0.11.18
│   ├── binstalk-downloader v0.6.0 (/Users/nobodyxu/Dev/cargo-binstall/crates/binstalk-downloader) (*)
│   ├── gix v0.48.0 (*)
│   └── gix-transport v0.33.1
│       ├── gix v0.48.0 (*)
│       └── gix-protocol v0.35.0
│           └── gix v0.48.0 (*)
├── rustls v0.20.8
│   ├── quinn v0.8.5
│   │   └── trust-dns-proto v0.22.0
│   │       └── trust-dns-resolver v0.22.0
│   │           ├── binstalk-downloader v0.6.0 (/Users/nobodyxu/Dev/cargo-binstall/crates/binstalk-downloader) (*)
│   │           └── reqwest v0.11.18 (*)
│   ├── quinn-proto v0.8.4
│   │   ├── quinn v0.8.5 (*)
│   │   └── quinn-udp v0.1.4
│   │       └── quinn v0.8.5 (*)
│   ├── tokio-rustls v0.23.4
│   │   ├── trust-dns-proto v0.22.0 (*)
│   │   └── trust-dns-resolver v0.22.0 (*)
│   ├── trust-dns-proto v0.22.0 (*)
│   └── trust-dns-resolver v0.22.0 (*)
├── rustls v0.21.5
│   ├── hyper-rustls v0.24.1
│   │   └── reqwest v0.11.18 (*)
│   ├── reqwest v0.11.18 (*)
│   └── tokio-rustls v0.24.1
│       ├── hyper-rustls v0.24.1 (*)
│       └── reqwest v0.11.18 (*)
├── tracing v0.1.37
│   ├── binstalk v0.13.0 (/Users/nobodyxu/Dev/cargo-binstall/crates/binstalk) (*)
│   ├── binstalk-downloader v0.6.0 (/Users/nobodyxu/Dev/cargo-binstall/crates/binstalk-downloader) (*)
│   ├── cargo-binstall v1.0.0 (/Users/nobodyxu/Dev/cargo-binstall/crates/bin)
│   ├── h2 v0.3.20
│   │   ├── hyper v0.14.27
│   │   │   ├── hyper-rustls v0.24.1 (*)
│   │   │   └── reqwest v0.11.18 (*)
│   │   ├── reqwest v0.11.18 (*)
│   │   └── trust-dns-proto v0.22.0 (*)
│   ├── hyper v0.14.27 (*)
│   ├── quinn v0.8.5 (*)
│   ├── quinn-proto v0.8.4 (*)
│   ├── quinn-udp v0.1.4 (*)
│   ├── tokio-util v0.7.8
│   │   ├── async_zip v0.0.15 (*)
│   │   ├── binstalk-downloader v0.6.0 (/Users/nobodyxu/Dev/cargo-binstall/crates/binstalk-downloader) (*)
│   │   ├── h2 v0.3.20 (*)
│   │   ├── reqwest v0.11.18 (*)
│   │   └── tower v0.4.13
│   │       └── binstalk-downloader v0.6.0 (/Users/nobodyxu/Dev/cargo-binstall/crates/binstalk-downloader) (*)
│   ├── tower v0.4.13 (*)
│   ├── trust-dns-proto v0.22.0 (*)
│   └── trust-dns-resolver v0.22.0 (*)
├── tracing-log v0.1.3
│   └── cargo-binstall v1.0.0 (/Users/nobodyxu/Dev/cargo-binstall/crates/bin)
└── wasm-bindgen-backend v0.2.87
    └── wasm-bindgen-macro-support v0.2.87
        └── wasm-bindgen-macro v0.2.87 (proc-macro)
            └── wasm-bindgen v0.2.87
                ├── js-sys v0.3.64
                │   ├── reqwest v0.11.18 (*)
                │   ├── wasm-bindgen-futures v0.4.37
                │   │   ├── reqwest v0.11.18 (*)
                │   │   └── wasm-streams v0.2.3
                │   │       └── reqwest v0.11.18 (*)
                │   ├── wasm-streams v0.2.3 (*)
                │   └── web-sys v0.3.64
                │       ├── reqwest v0.11.18 (*)
                │       ├── ring v0.16.20
                │       │   ├── quinn-proto v0.8.4 (*)
                │       │   ├── rustls v0.20.8 (*)
                │       │   ├── rustls v0.21.5 (*)
                │       │   ├── rustls-webpki v0.101.1
                │       │   │   └── rustls v0.21.5 (*)
                │       │   ├── sct v0.7.0
                │       │   │   ├── rustls v0.20.8 (*)
                │       │   │   └── rustls v0.21.5 (*)
                │       │   ├── trust-dns-proto v0.22.0 (*)
                │       │   └── webpki v0.22.0
                │       │       ├── quinn v0.8.5 (*)
                │       │       ├── quinn-proto v0.8.4 (*)
                │       │       ├── rustls v0.20.8 (*)
                │       │       ├── tokio-rustls v0.23.4 (*)
                │       │       ├── trust-dns-proto v0.22.0 (*)
                │       │       └── webpki-roots v0.22.6
                │       │           ├── reqwest v0.11.18 (*)
                │       │           ├── trust-dns-proto v0.22.0 (*)
                │       │           └── trust-dns-resolver v0.22.0 (*)
                │       ├── wasm-bindgen-futures v0.4.37 (*)
                │       └── wasm-streams v0.2.3 (*)
                ├── reqwest v0.11.18 (*)
                ├── wasm-bindgen-futures v0.4.37 (*)
                ├── wasm-streams v0.2.3 (*)
                └── web-sys v0.3.64 (*)
```

</details>

EFanZh:
> In [`cargo-binstall`](https://github.com/cargo-bins/cargo-binstall), `log` is a dependency that is heavily used

It seems that `cargo-binstall` mostly uses `tracing`, not `log`. Even though it depends on the `log` crate, not using its macros means it may not benefit from these optimizations.

NobodyXu:
> > In [`cargo-binstall`](https://github.com/cargo-bins/cargo-binstall), `log` is a dependency that is heavily used
> 
> It seems that `cargo-binstall` mostly uses logging macros from `tracing`, not `log`. Even though it depends on the `log` crate, not using its macros means it may not benefit from these optimizations.

It uses many dependencies which depends on `log`, like `gix`, `async-zip`, etc.

If you check the "details" which contained the dependency tree, you would find it actually heavily depends on `log`.

src/__private_api.rs
```diff
@@ -0,0 +1,321 @@
...
+    K: LogKVs,
+{
+    args.with(|args| log_2(module_path_and_file, line, level, args, target, kvs))
+}
+
+pub fn log<L, A, T, K>(
+    module_path_and_file: &'static (&'static str, &'static str),
+    line: u32,
+    level: L,
+    args: A,
```
NobodyXu, line 238: I think you could try passing `&Arguments<'_>` here and in all `log_*` instead of passing generics here.

I still think losing `Arguments::as_str` as an optimization is bad and passing it by reference could make up for it.

src/__private_api.rs
```diff
@@ -0,0 +1,321 @@
...
+    fn with(self, module_path: &'static str, f: impl FnOnce(&str)) {
+        f(module_path)
+    }
+}
+
+// `LogKVs`.
+
+impl LogKVs for &[(&str, &LogKvValue<'_>)] {
+    #[inline]
+    fn with(self, f: impl FnOnce(&[(&str, &LogKvValue)])) {
```
NobodyXu, line 142: Why every `Log*` trait has to accept a closure here?
Can't they just return the value directly?

Returning the closure would only add more overhead and generate more code.

NobodyXu (changes requested):

src/__private_api.rs
EFanZh: I have added a const function to check whether the format argument is a string literal as a not perfect replacement for `Arguments::as_str`, that’s why I used a generic argument here. Sometimes I need to pass `&str` as `args`.

As for `&Arguments`, I think I should do some test, then decide which one to use.

NobodyXu: > I have added a const function to check whether the format argument is a string literal as a not perfect replacement for `Arguments::as_str`

I was actually referring to using `Argument::as_str` in `Log::log`, the implementation might be able to avoid heap alloc using `Argument::as_str`, passing `&str` here breaks that optimization.

EFanZh: I am not sure I understand. `Log::log` is not implemented by this crate, how do I utilize `Argument::as_str`?

NobodyXu: > I am not sure I understand. `Log::log` is not implemented by this crate, so how do I utilize `Argument::as_str`?

Implementator of `Log::log` could use `Argument::as_str` to avoid heap allocation.

Now that you decided to pass arg as `&str` if possible, they can no longer utilize this optimization.

src/__private_api.rs
EFanZh: For `LogArgs`, I can’t return a `Arguments` value from a `&str` value:

```rust
fn f(s: &str) -> Arguments<'_> {
    format_args!("{s}") // Does not work because a local variable reference is captured.
}
```

So I have to use a callback function.

For other ones, since they need to be converted into references, the trait method may need to take a `&self` argument, and I don’t feed like taking references from `&str` or a zero sized type, since the will get a `&&str` type and `&ZST` type. But I am not very insistent on this.

NobodyXu: > So I have to use a callback function.

Yeah for `LogArgs` you have to use a callback due to limitation of `Arguments<'_>`, although I wonder
 
> For other ones, since they need to be converted into references, the trait method may need to take a `&self` argument, and I don’t feed like taking references from `&str` or a zero sized type, since the will get a `&&str` type and `&ZST` type. But I am not very insistent on this.

I don't think there's much issue with this since it will most likely be inlined.

src/__private_api.rs
EFanZh: I see your point. Maybe this can be optimized by `format_args!` as a special case?

NobodyXu: > I see your point. Maybe this can be optimized by `format_args!` as a special case?

Ideally yes, but that will require the args to be `&'static str`, which involves lifetime and is a bit hard to do.

I guess we can ask @m-ou-se for this?

EFanZh: A more simple could be `impl From<&'static str> for Arguments<'_>`, utilizing [`Arguments::new_const`](https://doc.rust-lang.org/stable/src/core/fmt/mod.rs.html#297).

src/__private_api.rs
EFanZh: Now I have removed callback functions from `LogTarget` and `LogKVs`.

EFanZh:
I have tests this PR on cargo-binstall, and here is the binary size comparison:

| 3         | off  |             1 |      14211752 |       14203560 |  -8192 |
| 3         | off  |            16 |      16739048 |       16726760 | -12288 |
| 3         | thin |             1 |      15170120 |       15170120 |      0 |
| 3         | thin |            16 |      17566376 |       17574568 |   8192 |
| 3         | fat  |             1 |      13859336 |       13859336 |      0 |
| 3         | fat  |            16 |      15231496 |       15227400 |  -4096 |
| "z"       | off  |             1 |      11451048 |       11430568 | -20480 |
| "z"       | off  |            16 |      13531848 |       13503176 | -28672 |
| "z"       | thin |             1 |      12307048 |       12286568 | -20480 |
| "z"       | thin |            16 |      13699720 |       13736584 |  36864 |
| "z"       | fat  |             1 |       9951752 |        9931272 | -20480 |
| "z"       | fat  |            16 |      10476040 |       10447368 | -28672 |

You can reproduce the result using [this program](https://github.com/EFanZh/log/tree/binary-size-test/cargo-binstall/test-binary-size).

KodrAus:
Thanks for investigating this @EFanZh! These were a useful point-in-time reference that helped make some improvements, but I think we should close now that things have moved along. Please feel free to open any fresh PRs that optimize binary size based on the current state of the code though.

PR #570, opened 2023-07-10
Remove `local_inner_macros` usage

EFanZh:
Since we have dropped support for pre-1.30 compilers.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -27,17 +27,17 @@
 ///     data.0, data.1, private_data);
 /// # }
 /// ```
-#[macro_export(local_inner_macros)]
+#[macro_export]
 macro_rules! log {
     // log!(target: "my_target", Level::Info, key1 = 42, key2 = true; "a {} event", "log");
     (target: $target:expr, $lvl:expr, $($key:tt = $value:expr),+; $($arg:tt)+) => ({
         let lvl = $lvl;
         if lvl <= $crate::STATIC_MAX_LEVEL && lvl <= $crate::max_level() {
             $crate::__private_api_log(
-                __log_format_args!($($arg)+),
+                $crate::__private_api::format_args!($($arg)+),
                 lvl,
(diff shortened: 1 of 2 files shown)


src/macros.rs
```diff
@@ -27,17 +27,17 @@
...
-#[macro_export(local_inner_macros)]
+#[macro_export]
 macro_rules! log {
     // log!(target: "my_target", Level::Info, key1 = 42, key2 = true; "a {} event", "log");
     (target: $target:expr, $lvl:expr, $($key:tt = $value:expr),+; $($arg:tt)+) => ({
         let lvl = $lvl;
         if lvl <= $crate::STATIC_MAX_LEVEL && lvl <= $crate::max_level() {
             $crate::__private_api_log(
-                __log_format_args!($($arg)+),
+                ::std::format_args!($($arg)+),
```
Thomasdezeeuw, line 37: I think this should be `core`.
```suggestion
                ::core::format_args!($($arg)+),
```

EFanZh: I don’t think this will change what `::std` means in macros.

EFanZh: I have replaced `::std` with `$crate::__private_api::` since the only thing I can trust is `$crate`.

Thomasdezeeuw (approved):
LGTM. I think we also remove `__private_api` and use `::core` instead, but this way works just as well.

EFanZh:
> LGTM. I think we also remove `__private_api` and use `::core` instead, but this way works just as well.

I think `::core` is not necessarily the `core` crate from the standard library. Like the `log` crate can rename `core` as `std`, downstream crates can also rename some other crate as `core`:

```rust
extern crate log as core;

fn main() {
    ::core::info!("xxx");
}
```

See <https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f22b3288ec76c436dc4a7a248e346072>.

PR #575, opened 2023-08-07
Group `target_module`, `path` and `file` arguments

EFanZh:
This is the first step of bring optimizations from #569 into the `log` crate.

This PR contains two commits:

- The first commit moves all private APIs into a single `__private_api` module so that it is easier to implement more elaborated optimizations.
- The second commit splits `&(target, module_path, file, line)` argument into two arguments: `&(target, module_path, file)` and `line`. The idea is that a module could have more than one log calls, so most likely, these log calls would have the same `(target, module_path, file)` triple, group them together will allow multiple log calls to reuse a same `(target, module_path, file)` instance, see <https://godbolt.org/z/z7jW67Ece> for comparison.

For one of my internal projects, this PR will reduce the binary size by about 2%.

I also tested this PR on [`cargo-binstall`](https://crates.io/crates/cargo-binstall), this is the test result:

| 3         | off  |             1 |      14211752 |       14199464 | -12288 |
| 3         | off  |            16 |      16739048 |       16739048 |      0 |
| 3         | thin |             1 |      15170120 |       15166024 |  -4096 |
| 3         | thin |            16 |      17566376 |       17566376 |      0 |
| 3         | fat  |             1 |      13859336 |       13859336 |      0 |
| 3         | fat  |            16 |      15231496 |       15231496 |      0 |
| "z"       | off  |             1 |      11451048 |       11438760 | -12288 |
| "z"       | off  |            16 |      13531848 |       13511368 | -20480 |
| "z"       | thin |             1 |      12307048 |       12298856 |  -8192 |
| "z"       | thin |            16 |      13699720 |       13695624 |  -4096 |
| "z"       | fat  |             1 |       9951752 |        9935368 | -16384 |
| "z"       | fat  |            16 |      10476040 |       10459656 | -16384 |

### DIFF
--- a/src/__private_api.rs
+++ b/src/__private_api.rs
@@ -1,0 +1,55 @@
+//! WARNING: this is not part of the crate's public API and is subject to change at any time
+
+use crate::{Level, Metadata, Record};
+use std::fmt::Arguments;
+pub use std::option::Option;
+pub use std::{file, format_args, line, module_path, stringify};
+
+#[cfg(not(feature = "kv_unstable"))]
+pub fn log(
+    args: Arguments,
+    level: Level,
+    &(target, module_path, file, line): &(&str, &'static str, &'static str, u32),
+    kvs: Option<&[(&str, &str)]>,
+) {
+    if kvs.is_some() {
+        panic!(
+            "key-value support is experimental and must be enabled using the `kv_unstable` feature"
+        )
+    }
+
+    crate::logger().log(
+        &Record::builder()
+            .args(args)
+            .level(level)
+            .target(target)
+            .module_path_static(Some(module_path))
+            .file_static(Some(file))
+            .line(Some(line))
+            .build(),
+    );
+}
+
+#[cfg(feature = "kv_unstable")]
+pub fn log(
+    args: Arguments,
+    level: Level,
+    &(target, module_path, file, line): &(&str, &'static str, &'static str, u32),
+    kvs: Option<&[(&str, &dyn crate::kv::ToValue)]>,
+) {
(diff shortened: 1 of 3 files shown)


EFanZh:
OK, I have tested the performance using this code:

```rust
use criterion::Criterion;

fn log_string_literal(c: &mut Criterion) {
    let mut benchmark_group = c.benchmark_group("log string literal");

    benchmark_group.bench_function("original", |b| b.iter(|| log::info!("abc")));
    benchmark_group.bench_function("optimized", |b| b.iter(|| log_optimized::info!("abc")));

    benchmark_group.finish();
}

criterion::criterion_group!(benches, log_string_literal);

criterion::criterion_main!(benches);
```

Dependencies in `Cargo.toml`

```toml
criterion = "*"
log = "*"
log_optimized = { git = "ssh://git@github.com/EFanZh/log.git", branch = "group-target-module-path-and-file", package = "log" }
```

On average, the original one takes 246.84 ps per iteration, and the optimized on takes 245.54 ps per iteration, I consider this difference insignificant.

NobodyXu (approved):
Amazing!

Thank you for this PR and testing this on `cargo-binstall`!

EFanZh:
Hi @KodrAus, I’d really appreciate it if you could take a look at this PR. I intend to do some binary size optimizations with ideas from #569. I plan to test and implement each idea with a separate PR, so if this PR gets accepted, I can move on to the next PR. Will these sort of optimizations be accepted?

KodrAus (approved):
Awesome, thanks for working on this @EFanZh!

PR #576, opened 2023-08-13
Specialize empty key value pairs

EFanZh:
Currently, the `__private_api::log` function takes `None::<&[(&str, _)]>` as the `kvs` argument if no key-value pair is provided, but passing a `None` value may have some extra cost. We should be able to avoid this cost by encoding this information into a zero sized type (here I am using `()`, not sure if it is better to use a dedicated new type (such as `struct EmptyKVs;`)). Constructing a zero sized value is zero cost, which is less than constructing a `None::<&[(&str, _)]>`.

### DIFF
--- a/src/__private_api.rs
+++ b/src/__private_api.rs
@@ -1,55 +1,87 @@
 //! WARNING: this is not part of the crate's public API and is subject to change at any time
 
+use self::sealed::KVs;
 use crate::{Level, Metadata, Record};
 use std::fmt::Arguments;
-pub use std::option::Option;
 pub use std::{file, format_args, line, module_path, stringify};
 
+#[cfg(feature = "kv_unstable")]
+pub type Value<'a> = dyn crate::kv::value::ToValue + 'a;
+
 #[cfg(not(feature = "kv_unstable"))]
-pub fn log(
+pub type Value<'a> = str;
+
+mod sealed {
+    /// Types for the `kv` argument.
+    pub trait KVs<'a> {
+        fn into_kvs(self) -> Option<&'a [(&'a str, &'a super::Value<'a>)]>;
+    }
+}
+
+// Types for the `kv` argument.
+
+impl<'a> KVs<'a> for &'a [(&'a str, &'a Value<'a>)] {
+    #[inline]
+    fn into_kvs(self) -> Option<&'a [(&'a str, &'a Value<'a>)]> {
+        Some(self)
+    }
+}
+
+impl<'a> KVs<'a> for () {
+    #[inline]
+    fn into_kvs(self) -> Option<&'a [(&'a str, &'a Value<'a>)]> {
+        None
+    }
+}
+
+// Log implementation.
+
+fn log_impl(
     args: Arguments,
     level: Level,
     &(target, module_path, file): &(&str, &'static str, &'static str),
     line: u32,
-    kvs: Option<&[(&str, &str)]>,
+    kvs: Option<&[(&str, &Value)]>,
 ) {
+    #[cfg(not(feature = "kv_unstable"))]
     if kvs.is_some() {
         panic!(
             "key-value support is experimental and must be enabled using the `kv_unstable` feature"
         )
     }
 
-    crate::logger().log(
-        &Record::builder()
-            .args(args)
-            .level(level)
-            .target(target)
-            .module_path_static(Some(module_path))
-            .file_static(Some(file))
-            .line(Some(line))
-            .build(),
-    );
+    let mut builder = Record::builder();
+
+    builder
+        .args(args)
+        .level(level)
+        .target(target)
+        .module_path_static(Some(module_path))
+        .file_static(Some(file))
+        .line(Some(line));
+
+    #[cfg(feature = "kv_unstable")]
+    builder.key_values(&kvs);
+
+    crate::logger().log(&builder.build());
 }
 
-#[cfg(feature = "kv_unstable")]
-pub fn log(
+pub fn log<'a, K>(
     args: Arguments,
     level: Level,
-    &(target, module_path, file): &(&str, &'static str, &'static str),
+    target_module_path_and_file: &(&str, &'static str, &'static str),
     line: u32,
-    kvs: Option<&[(&str, &dyn crate::kv::ToValue)]>,
-) {
-    crate::logger().log(
-        &Record::builder()
-            .args(args)
-            .level(level)
-            .target(target)
-            .module_path_static(Some(module_path))
-            .file_static(Some(file))
-            .line(Some(line))
-            .key_values(&kvs)
-            .build(),
-    );
+    kvs: K,
+) where
+    K: KVs<'a>,
+{
+    log_impl(
+        args,
+        level,
+        target_module_path_and_file,
+        line,
+        kvs.into_kvs(),
+    )
 }
 
 pub fn enabled(level: Level, target: &str) -> bool {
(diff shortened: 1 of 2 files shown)


src/__private_api.rs
```diff
@@ -1,57 +1,116 @@
...
+pub type Value<'a> = dyn crate::kv::value::ToValue + 'a;
+
 #[cfg(not(feature = "kv_unstable"))]
-pub fn log(
+pub type Value<'a> = str;
+
+mod sealed {
+    /// Argument that provides key-value pairs.
+    pub trait KVs<'a> {
+        fn into_kvs(self) -> Option<&'a [(&'a str, &'a super::Value<'a>)]>;
```
NobodyXu, line 17: Why not take "&self" so that it can be used as a trait object?

```suggestion
        fn into_kvs(&self) -> Option<&'a [(&'a str, &'a super::Value<'a>)]>;

```

NobodyXu (changes requested):

src/__private_api.rs
EFanZh: Normally, I prefer passing small `Copy` types by value instead of reference. Clippy has a check for this: <https://rust-lang.github.io/rust-clippy/master/index.html#/trivially_copy_pass_by_ref>. Also, this is a private trait, all usages are restricted to this crate, and it is not used as a trait object, so it is unnecessary to make it object safe.

NobodyXu: It does make sense to pass kvs in value instead of reference, resolving this.

EFanZh:
I have tested the effect on binary size of this PR on `cargo-binstall`, and this is the result:

| 3         | off  |             1 |      13754440 |       13754440 |      0 |
| 3         | off  |            16 |      16343208 |       16326824 | -16384 |
| 3         | thin |             1 |      14520328 |       14520328 |      0 |
| 3         | thin |            16 |      17051784 |       17035400 | -16384 |
| 3         | fat  |             1 |      13152232 |       13152232 |      0 |
| 3         | fat  |            16 |      14385128 |       14368744 | -16384 |
| "z"       | off  |             1 |      11247688 |       11251784 |   4096 |
| "z"       | off  |            16 |      13197448 |       13172872 | -24576 |
| "z"       | thin |             1 |      12144648 |       12136456 |  -8192 |
| "z"       | thin |            16 |      13488232 |       13488232 |      0 |
| "z"       | fat  |             1 |       9662440 |        9658344 |  -4096 |
| "z"       | fat  |            16 |      10113000 |       10100712 | -12288 |

Aside from the `(opt-level=z, lto=off, codegen-units=1)` combination, all binary size changes are non-positive, you can reproduce the result using: <https://github.com/EFanZh/log/tree/binary-size-test/cargo-binstall/test-binary-size>.

I also tested the binary sizes with the `instantiate_log_function` function removed:

| opt-level | lto  | codegen-units | original size | optimized size | diff       |
| --------- | ---- | ------------- | ------------- | -------------- | ---------- |
| 3         | off  |             1 |      13754440 |       13754440 |          0 |
| 3         | off  |            16 |      16343208 |       16326824 |     -16384 |
| 3         | thin |             1 |      14520328 |       14520328 |          0 |
| 3         | thin |            16 |      17051784 |       17035400 |     -16384 |
| 3         | fat  |             1 |      13152232 |       13152232 |          0 |
| 3         | fat  |            16 |      14385128 |       14368744 |     -16384 |
| "z"       | off  |             1 |      11247688 |       11251784 |       4096 |
| "z"       | off  |            16 |      13197448 |   **13176968** | **-20480** |
| "z"       | thin |             1 |      12144648 |       12136456 |      -8192 |
| "z"       | thin |            16 |      13488232 |   **13492328** |   **4096** |
| "z"       | fat  |             1 |       9662440 |        9658344 |      -4096 |
| "z"       | fat  |            16 |      10113000 |       10100712 |     -12288 |

The binary sizes are mostly the same as the previous one, with the exception of `(opt-level=z, lto=off, codegen-units=16)` and `(opt-level=z, lto=thin, codegen-units=16)` combination, whose effects become worse after removing the `instantiate_log_function` function.

I also tested the runtime performance with a hand written benchmark:

```text
original  => cycles: 1115, instructions: 79, wall time: 322ns.
optimized => cycles: 954, instructions: 79, wall time: 323ns.
```

You can find the benchmark code here: <https://github.com/EFanZh/log/tree/benchmark>. I give up using `criterion` because the result changes significantly between runs, and I also often run into “took zero time” error. I also tried `iai`, but I get <https://github.com/bheisler/iai/issues/34> error. Finally I decided to use [perf-event](https://crates.io/crates/perf-event), since it is what the Rust language uses for their performance benchmark: <https://github.com/rust-lang/rustc-perf/blob/master/collector/benchlib/src/measure/perf_counter/linux.rs>.

src/__private_api.rs
```diff
@@ -1,57 +1,116 @@
...
 }
+
+/// This function is not intended to be actually called by anyone. The intention of this function is for forcing the
+/// compiler to generate monomorphizations of the `log` function, so that they can be shared by different downstream
+/// crates.
+///
+/// The idea is that with [`share-generics`](https://github.com/rust-lang/rust/pull/48779), downstream crates can reuse
+/// generic monomorphizations from upstream crates, but not siblings crates. So it is best to instantiate these
+/// monomorphizations in the `log` crate, so downstream crates are guaranteed to be able to share them.
+pub fn instantiate_log_function() -> usize {
```
KodrAus, line 98: This seems a bit obtuse to include here. Based on your binary size profiling I'm just not sure this is pulling its weight. Since we don't have CI in place to monitor the impact of binary size it's possible the optimization it indirectly encourages could be undone at any point without our knowing.

So I think we should remove this function.

KodrAus (commented):
Thanks for investigating this @EFanZh! I think we should remove the `instantiate_log_function` item, but otherwise this looks good to me.

src/__private_api.rs
EFanZh: OK, I have `removed instantiate_log_function`.

EFanZh:
Hi @KodrAus, is there anything I need to do to get this PR merged?

KodrAus (approved):
Thanks for your patience on this one @EFanZh 🙇

This looks good to me.

src/__private_api.rs
```diff
@@ -1,55 +1,87 @@
...
+    kvs: K,
+) where
+    K: KVs<'a>,
+{
+    log_impl(
+        args,
+        level,
+        target_module_path_and_file,
+        line,
+        kvs.into_kvs(),
```
tisonkun, line 83: @EFanZh we still pass `None::<&[(&str, &Value)]>` here.

Could you elaborate a bit on how moving the type convert one level down can help?

I'm trying to understand how the extra abstract works (and avoid breaking it in any following patch).

src/__private_api.rs
```diff
@@ -1,55 +1,87 @@
...
+    crate::logger().log(&builder.build());
 }
 
-#[cfg(feature = "kv_unstable")]
-pub fn log(
+pub fn log<'a, K>(
     args: Arguments,
     level: Level,
-    &(target, module_path, file): &(&str, &'static str, &'static str),
+    target_module_path_and_file: &(&str, &'static str, &'static str),
```
tisonkun, line 72: Besides, does this change affect significantly? Or just some identical refactor.

src/__private_api.rs
EFanZh: Sometimes the `log` function is not inlined, which allows callers to pass ZST to the log function as `kvs` argument, which does not generate machine instructions for `kvs` argument at the caller side. There might be a lot of callers in a program, but only one callee function, so I would like to reduce the machine instructions at caller side as much as possible.

PR #577, opened 2023-08-20
Update docs to 2021 edition, test

nyurik:
* Inline format arg in documentation and a few call places
* Add a unit test for inlined format args

I was surprised to discover that log crate has no edition value... Any reason to keep it at 2015?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -48,24 +48,24 @@
 //!
 //! ### Examples
 //!
-//! ```edition2018
+//! ```edition2021
 //! # #[derive(Debug)] pub struct Yak(String);
 //! # impl Yak { fn shave(&mut self, _: u32) {} }
 //! # fn find_a_razor() -> Result<u32, u32> { Ok(1) }
 //! use log::{info, warn};
 //!
 //! pub fn shave_the_yak(yak: &mut Yak) {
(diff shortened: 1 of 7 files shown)


src/lib.rs
```diff
@@ -48,24 +48,24 @@
 //!
 //! ### Examples
 //!
-//! ```edition2018
+//! ```edition2021
```
Thomasdezeeuw, line 51: I think this can just be removed everywhere.
```suggestion
//! ```
```

Thomasdezeeuw (approved):
LTGTM. The reasons the docs to use edition 2015 is because we recently bumped our MSRV from 1.31 to 1.60 in 0.4.19 (https://github.com/rust-lang/log/pull/552).

nyurik:
@Thomasdezeeuw I think they should be removed after #580, or else they will continue using 2015 which i think is a bit confusing

KodrAus:
We can probably merge #580 first, and then circle back to this one.

nyurik:
Now that #580 is merged, I updated this PR to remove the editions and also add the 580's forgotten changelog entry.  This one should be far easier to review :)

nyurik:
@Thomasdezeeuw and @KodrAus a friendly ping. I just rebased it just to keep it current. Let me know if any changes are needed. Thx!

PR #579, opened 2023-08-20
Simplify Display impl

nyurik:
For some reason, `fmt::Display for Error` uses an unusual form `match &self.inner { &Value(ref err) => err.fmt(f), ...}`.  It seems `Value(err)` will work just as well.

If this is actually needed, we should add some comment explaining why so


Thomasdezeeuw (approved):
This code was required in older rustc versions where you had to be explicit about whether you owned the variables (`err` and `msg`) or had a reference to them (`ref err` or `ref msg`). But the compiler got better at determining this for us, so now we can use cleaner code like this.

PR #580, opened 2023-08-20
Set all crates to 2021 edition

nyurik:
For some reason all crates still use 2015 edition. I think it is a good idea to specify edition explicitly, or at least document why it should be kept back (and still add it to the cargo.toml for documentation purposes)

* Remove all `extern crate`
* changed `use self::sval::...` to `use sval::...`, and the same for `sval_ref`
* Simplified `fn get<'v>(&'v self, key: Key) -> Option<Value<'v>>` to `fn get(&self, key: Key) -> Option<Value<'_>>`
* Simplified many unit tests with references

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -61,7 +61,11 @@ jobs:
         rustup update stable --no-self-update
         rustup default stable
         rustup component add rustfmt
+      # log repo does not use Cargo workspaces, so `cargo fmt` will not check all the code
+      # perhaps this should be changed in the future
     - run: cargo fmt -- --check
+    - run: cargo fmt --manifest-path test_max_level_features/Cargo.toml -- --check
+    - run: cargo fmt --manifest-path tests/Cargo.toml -- --check
 
   clippy:
     name: Clippy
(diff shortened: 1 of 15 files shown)


Thomasdezeeuw (approved):
LGTM, but the CI is failing.

nyurik:
It took a few tries, but I think I fixed all the outstanding issues with this one.  As the result, 50 less lines of code, and a few extra CI checks.

.github/workflows/main.yml
```diff
@@ -62,6 +62,8 @@ jobs:
         rustup default stable
         rustup component add rustfmt
     - run: cargo fmt -- --check
```
Thomasdezeeuw, line 64: Maybe a note saying that `--all` doesn't work because we're not using Cargo workspaces?

Thomasdezeeuw (approved):

.github/workflows/main.yml
nyurik: will do. Perhaps in the next iteration we should add those?  I didn't want to make all changes at once :)

KodrAus:
Thanks for doing some much needed housekeeping @nyurik! It looks like we just have some conflicts to resolve now.

nyurik:
thx @KodrAus, rebased & fixed

nyurik:
@KodrAus hi, would you be the right person to review? I would like to make a few more improvements like use workspaces, but all other changes would be based on this one, so I don't want to make them until this is agreed on & merged. Thx!

KodrAus:
Merged! Thanks again @nyurik

PR #581, opened 2023-08-26
Add column numbers

printfn:
This adds column numbers to logs, using the [`column!`](https://doc.rust-lang.org/stable/std/macro.column.html) macro.


Thomasdezeeuw:
Personally I would be against this change. I don't think it adds any benefits to the log messages, usually the lines numbers themselves are enough, I rarely see two log statements on the same line. Furthermore it would increase the binary sizes, something that people have been trying to reduce lately, see https://github.com/rust-lang/log/pull/569 and https://github.com/rust-lang/log/pull/575 for example.

PR #582, opened 2023-08-29
migrate to a single workspace

nyurik:
Consolidate the build setup for all crates to use 2021-style cargo.toml with workspaces.  Note that the `tests/` dir has an unusual setup, where the tests dir is both a separate crate and a set of integration tests for the log crate.

I am not certain of all the interactions here, and making it into a regular test crate breaks the filter test, so leaving as is for now until someone can explain what's going on there.


KodrAus:
Thanks for the PR @nyurik! Sorry for the delayed response here. The integration test project is a bit of an oddity, but it seems like pulling everything into a single workspace adds to the confusion rather than reducing it. I think this could be something to revisit in the future, but for the time being the status quo is less machinery.

PR #584, opened 2023-08-29
Review of key-value feature

Thomasdezeeuw:
Discussion for https://github.com/rust-lang/log/issues/328.

### DIFF
--- a/src/kv/source.rs
+++ b/src/kv/source.rs
@@ -374,6 +374,12 @@ mod std_support {
     }
 }
 
+// REVIEW: I'm not a 100% convinced that AsmAp and AsList are pulling their
+// weight. Their Source implementation simply calls the underlying type's
+// implementation, the only difference I can see is the sval::Value and
+// serde::Serialize implementations (ignoring fmt::Debug). Does that fall within
+// scope for the log crate?
+
 /// The result of calling `Source::as_map`.
 pub struct AsMap<S>(S);
 
--- a/src/kv/value.rs
+++ b/src/kv/value.rs
@@ -83,6 +83,11 @@ macro_rules! as_sval {
 /// - Using the `ToValue` trait.
 /// - Using the standard `From` trait.
 ///
+// REVIEW: maybe add a section about which conversion method to use and when? Or
+// "the prefered ordering" of which methods to use. For example I think it's
+// quite confusing when the use capture_* and when to use from_* (the only
+// difference I can spot is the support for downcasting in capture_*).
+///
 /// ## Using the `Value::capture_*` methods
 ///
 /// `Value` offers a few constructor methods that capture values of different kinds.
@@ -376,6 +381,9 @@ impl<'v> fmt::Display for Value<'v> {
     }
 }
 
+// REVIEW: I'm wondering if people will accidentally use these implementations
+// when converting a T: fmt::Debug/Display to a Value, losing the ability to
+// downcast (without knowing it).
 impl ToValue for dyn fmt::Debug {
     fn to_value(&self) -> Value {
         Value::from_dyn_debug(self)
@@ -615,6 +623,9 @@ mod std_support {
 
     impl<'v> Value<'v> {
         /// Try convert this value into a string.
+        // REVIEW: maybe rename this to contain COW somehow? I would expect &str
+        // be returned based on the to_str method name (same as e.g.
+        // Key::to_str).
         pub fn to_str(&self) -> Option<Cow<str>> {
             self.inner.to_str()
         }
@@ -635,6 +646,9 @@ pub trait Visit<'v> {
     /// more specific methods that aren't overridden.
     /// The `Value` may be formatted using its `fmt::Debug` or `fmt::Display` implementation,
     /// or serialized using its `sval::Value` or `serde::Serialize` implementation.
+    // REVIEW: Do we want seperate methods for sval and serde values? Otherwise
+    // we still have sort of "catch all" method in visit_any. Of course they can
+    // default to calling visit_any (as all other methods do).
     fn visit_any(&mut self, value: Value) -> Result<(), Error>;
 
     /// Visit an unsigned integer.
--- a/src/kv/key.rs
+++ b/src/kv/key.rs
@@ -49,6 +49,8 @@ impl<'k> Key<'k> {
         self.key
     }
 
+    // REVIEW: do we still want this? Isn't most of log's API based on borrowed
+    // types?
     /// Try get a string borrowed for the `'k` lifetime from this key.
     pub fn to_borrowed_str(&self) -> Option<&'k str> {
         // NOTE: This API leaves room for keys to be owned


src/kv/source.rs
```diff
@@ -374,6 +374,12 @@ mod std_support {
     }
 }
 
+// REVIEW: I'm not a 100% convinced that AsmAp and AsList are pulling their
```
KodrAus, line 377: These also seem fair to remove and consider adding later, or adding in a utilities crate :+1:

src/kv/value.rs
```diff
@@ -83,6 +83,11 @@ macro_rules! as_sval {
 /// - Using the `ToValue` trait.
 /// - Using the standard `From` trait.
 ///
+// REVIEW: maybe add a section about which conversion method to use and when? Or
```
KodrAus, line 86: Good call :+1:

The preference would be for `capture_*`, then `from_*`. The reason being if you `Value::capture_display(&42)` you'll get a number, but if you use `Value::from_display(&42)` you'll get a string. I'd be open to some different naming that made this more obvious. We may consider also only adding them for `Display` and `Debug` since `serde` and `sval` are naturally structured. We could also consider offering just a single `capture` method that is what `capture_display` currently is, and using that by default in the macros. The goal is that by default if you log a boolean or number you should get a boolean or number, and you shouldn't _have_ to implement any non-standard-library traits like `ToValue` to log something.

I'm open to suggestions though. What do you think?

src/kv/value.rs
```diff
@@ -376,6 +381,9 @@ impl<'v> fmt::Display for Value<'v> {
     }
 }
 
+// REVIEW: I'm wondering if people will accidentally use these implementations
```
KodrAus, line 384: Good point :+1: I'd be up for removing them.

src/kv/value.rs
```diff
@@ -615,6 +623,9 @@ mod std_support {
 
     impl<'v> Value<'v> {
         /// Try convert this value into a string.
+        // REVIEW: maybe rename this to contain COW somehow? I would expect &str
```
KodrAus, line 626: Good point :+1: There's probably some precedence in the standard library to look at.

src/kv/value.rs
```diff
@@ -635,6 +646,9 @@ pub trait Visit<'v> {
     /// more specific methods that aren't overridden.
     /// The `Value` may be formatted using its `fmt::Debug` or `fmt::Display` implementation,
     /// or serialized using its `sval::Value` or `serde::Serialize` implementation.
+    // REVIEW: Do we want seperate methods for sval and serde values? Otherwise
```
KodrAus, line 649: That's a good question. The design here is that you shouldn't need separate methods, because you don't know what framework a user logged their structured data with, and as a consumer it doesn't matter to you. They pick the one that best suits them and you just pick the one that best suits you. `Value` will bridge them for you.

The drawback is it's really non-obvious how to work with it. You try to visit a `Value` to see what's in it and you get a method that gives you... a `Value`. This could just be an issue of documentation. The API itself doesn't give you a lot of clues about how to use it. That could be a reason to add separate methods, and examples alongside them on how to forward `visit_any` into them.

The only thing we'd need to worry about is whether `Visitor` needs to be object safe. I think it doesn't? In which case it's fine to add separate methods.

src/kv/value.rs
Thomasdezeeuw: Done in https://github.com/rust-lang/log/pull/585.

src/kv/value.rs
Thomasdezeeuw: I can't find much precedence in std lib. Using `rg --fixed-strings -e '-> Cow<'` all I could find was `IntoCow::into_cow` as a testing trait.

Maybe `to_cow_str`?

src/kv/value.rs
KodrAus: > to determine if serde or sval is used

I think this is the part you shouldn't need to do in `visit_any`. In `std_logger`'s JSON format for instance, your [`visit_any`](https://github.com/Thomasdezeeuw/std-logger/blob/eff67f4c33b89c8debd325c1ad79724e6cd6af2c/src/format/json.rs#L169-L173) implementation could be:

```rust
fn visit_any(&mut self, value: kv::Value) -> Result<(), kv::Error> {
        Ok(serde_json::to_writer(&mut self.0, &value)?)
}
```

where `self.0` is `impl std::io::Write`. You don't need to know whether the value comes from `serde` or `sval`, or even need to override any of the other `Visitor` methods. You'll get the most accurate structured representation no matter what the value was captured with.

Thomasdezeeuw: > > to determine if serde or sval is used
> 
> I think this is the part you shouldn't need to do in `visit_any`. In `std_logger`'s JSON format for instance, your [`visit_any`](https://github.com/Thomasdezeeuw/std-logger/blob/eff67f4c33b89c8debd325c1ad79724e6cd6af2c/src/format/json.rs#L169-L173) implementation could be:
> 
> ```rust
> fn visit_any(&mut self, value: kv::Value) -> Result<(), kv::Error> {
>         Ok(serde_json::to_writer(&mut self.0, &value)?)
> }
> ```
> 
> where `self.0` is `impl std::io::Write`. You don't need to know whether the value comes from `serde` or `sval`, or even need to override any of the other `Visitor` methods. You'll get the most accurate structured representation no matter what the value was captured with.

So you're saying that `kv::Value` implements `serde::Deserialize` regardless of the source of the actual value (i.e. a `sval` value would be deserialised correctly?). And vice versa?

If so, it does essentially mean that any logging implementation that want to support structured value needs to use either `serde` or `sval`.

KodrAus: > So you're saying that kv::Value implements serde::Deserialize regardless of the source of the actual value (i.e. a sval value would be deserialised correctly?). And vice versa?

That's essentially right :+1: It always implements `serde::Serialize` and the `Serialize` implementation will always use the least-lossy representation of the data you've captured.

> If so, it does essentially mean that any logging implementation that want to support structured value needs to use either serde or sval.

For simple values like numbers, strings, and booleans you don't need to. If you want to log complex values like maps or sequences then you'll need to use one of them. If you're happy to treat complex values as text then you can also always use `std::fmt`.

src/kv/value.rs
KodrAus: `to_cow_str` sounds totally fine to me here.

src/kv/value.rs
KodrAus: > Does it make sense to change the macros to always call log::ToValue::to_value($value)?

The idea was to try discourage libraries from implementing `ToValue` themselves so it doesn't become yet another widespread piece of trivia that library authors need to keep on top of. Defaulting to `Display + 'static` means a lot of library types like `Uuid` can be captured, but the drawback of throwing lifetime errors at you if you try log a non-`'static` value would be unpleasant.

For what it's worth, I would be ok with using `ToValue` by default in addition to adding some easy syntax for `Display`/`Debug`/`Serialize`/`Value`. If we did that then we could remove the `capture_*` methods altogether; they only exist to support downcasting to tell if we're logging numbers/booleans etc.

> Then, once specialisation is stablised

🥲

KodrAus:
Thanks for all the time you've spent digging through this so far @Thomasdezeeuw! I think we're getting towards something that feels appropriate for the `log` crate without losing capability.

src/kv/value.rs
Thomasdezeeuw: Created https://github.com/rust-lang/log/pull/592.

src/kv/value.rs
Thomasdezeeuw: https://github.com/rust-lang/log/pull/593

KodrAus:
Just checking back in on this, is the main blocker we've got deciding whether to default to capturing using `ToValue` and supporting some `tracing`-like syntax for changing it?

Thomasdezeeuw:
> Just checking back in on this, is the main blocker we've got deciding whether to default to capturing using `ToValue` and supporting some `tracing`-like syntax for changing it?

Yes, I think those are the two main things from this pr.

KodrAus:
I'm doing some work on this now; removing the `Value::capture_*` and downcasting methods, and improving the overall documentation for the `kv` module.

It looks like we already default to using `ToValue` for key-values, so nothing needs to change there. That just leaves some nicer macro support for tweaking how values are captured.

Are you happy with that plan @Thomasdezeeuw?

KodrAus:
I've opened #613 for this. I'll cc there once it's ready for review, but wanted to point it out here too.

Thomasdezeeuw:
> Are you happy with that plan @Thomasdezeeuw?

Yes, sounds good :+1:

PR #586, opened 2023-09-07
let feature release_max_level_* avoid useless code generation

tinglou:
if feature release_max_level_* is turned on, the lower level log statement will be ignored,


Thomasdezeeuw (changes requested):
This should already be taken care of by the if statement in https://github.com/rust-lang/log/blob/35ee9442132c4aa49d22e8f00e2ba79fe149d920/src/macros.rs#L35-L43, see https://github.com/rust-lang/log/blob/35ee9442132c4aa49d22e8f00e2ba79fe149d920/src/lib.rs#L1480-L1534.

Thomasdezeeuw:
Based on the lack of progress I'm closing this. @tinglou if you want to continue still we can reopen.

PR #587, opened 2023-09-14
Fix incorrect lifetime in Value::to_str()

peterjoel:
Without this change, the following code will not compile:

```Rust
let foo = record
    .key_values()
    .get(Key::from_str("foo"))
    .and_then(|val| val.to_str());
```
with the following error:

```lang-none
   |
32 |         .and_then(|val| val.to_str());
   |                         ^^^^^^^^^^^^ returns a reference to data owned by the current function
```

### DIFF
--- a/src/kv/value.rs
+++ b/src/kv/value.rs
@@ -635,7 +635,7 @@ mod std_support {
 
     impl<'v> Value<'v> {
         /// Try convert this value into a string.
-        pub fn to_str(&self) -> Option<Cow<str>> {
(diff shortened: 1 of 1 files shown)


Thomasdezeeuw (approved):

Thomasdezeeuw:
Not really an incorrect lifetime though, the change introduces a broader/larger lifetime.

peterjoel:
Thanks for the fast response!

And yes, _incorrect_ was perhaps a little opinionated! What I meant was: _shorter than necessary and likely unintentionally so_.

PR #592, opened 2023-10-19
Rename Value::to_str to to_cow_str

Thomasdezeeuw:


Thomasdezeeuw:
Oops, forgot to run with `--all-features`, please squash the commits.

KodrAus (approved):

darkdragon-001:
FYI: You released an breaking change in a patch version update. This broke depending projects like `tauri` (see linked issues).

I suggest reverting, releasing another patch version update for a fix, apply this change again and release again with updated major version.

Thomasdezeeuw:
> FYI: You released a breaking change (this PR) in a patch version update. This broke depending projects like `tauri` (see linked issues).

The API that was changed was only enabled by the `kv_unstable` feature, which as the name implies is unstable. This means that users opted into possible breakages, which is what happened here.

> I suggest reverting, releasing another patch version update for a fix, apply this change again and release again with updated major version.

I'm afraid that I don't think that's going to happen as the `kv` (no `_unstable`) feature was released in the same version. I suggest updating to it. The `kv` feature is covered by the same stability guarantees as the rest of the crate (i.e. a rename like this won't happen).

PR #594, opened 2023-10-20
Clarify documentation and simplify initialization of `STATIC_MAX_LEVEL`

ptosi:
Rephrase the documentation to make it clear that the `max_level_*` features are also used by release builds.

With the MSRV bump, new language features allow initializing the constant in a way that's conciser and hopefully clearer.

Add some tests, used to validate the change.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -241,9 +241,7 @@
 //!
 //! # Compile time filters
 //!
-//! Log levels can be statically disabled at compile time via Cargo features. Log invocations at
-//! disabled levels will be skipped and will not even be present in the resulting binary.
-//! This level is configured separately for release and debug builds. The features are:
+//! Log levels can be statically disabled at compile time by enabling one of these Cargo features:
 //!
 //! * `max_level_off`
 //! * `max_level_error`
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1477,65 +1480,24 @@ pub mod __private_api;
...
-        }
-        #[cfg(feature = "max_level_debug")]
-        {
-            return LevelFilter::Debug;
-        }
-
-        LevelFilter::Trace
-    }
-}
+pub const STATIC_MAX_LEVEL: LevelFilter = match cfg!(debug_assertions) {
```
Thomasdezeeuw, line 1483: I don't think this will work with our MSRV?

Thomasdezeeuw (commented):
Doc changes LGTM.

src/lib.rs
ptosi: Interesting, which bit?

I've cheked these changes with `cargo +1.60.0 test`; how do you recommend I test it?

Thomasdezeeuw: Right, log used to have a MSRV of 1.34 or something, but 1.60 should work :+1:

Thomasdezeeuw (approved):

PR #596, opened 2023-10-29
Remove remaining unsafe from `kv` module

joshlf:


Thomasdezeeuw (changes requested):
I don't think we should add dependencies to be honest, especially not for two lines of code.

joshlf:
No worries, understandable! Feel free to close.

PR #602, opened 2023-12-03
Normalize line ending

EFanZh:
Somehow I created this file with CRLF line ending, which causes some diff difficult to see, this PR tries to normalize the line ending to LF.


Thomasdezeeuw:
Thanks @EFanZh

PR #613, opened 2024-01-26
Get structured logging API ready for stabilization

### LINKED ISSUE #149: Structured Logging

### LINKED ISSUE #328: Tracking issue for kv_unstable (structured logging)

### LINKED ISSUE #357: Logging std::error::Errors

### LINKED ISSUE #436: Stabilize kv_unstable

KodrAus:
Closes #149 
Closes #328 
Closes #357 
Closes #436 
Closes #584 

Based on the review in #584

The plan is to:

- [x] Deprecate `Value::capture_*` and `Value::downcast_*` methods. All deprecated APIs are gated under the old `kv_unstable` features and will be removed once we remove those old unstable features.
- [x] Fill in the documentation for structured logging support so people can actually use it.
- [x] Add some macro syntax for capturing attributes as debug or display instead of the default `ToValue` (`key:? = value`).
- [x] Only pull in `value-bag` as a dependency if either `serde` or `sval` is also needed. In cases where you just want basic structured logging, we should do it in a dependency-free way. This will just mean adding a simple `ValueInner` enum.

I'll cc once this is ready for review, to save a bit of notification noise.

Once this PR is complete, we should be able to stabilize structured logging support in the `log` crate.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -44,10 +44,10 @@ jobs:
     - run: cargo test --verbose --all-features
     - run: cargo test --verbose --features serde
     - run: cargo test --verbose --features std
-    - run: cargo test --verbose --features kv_unstable
-    - run: cargo test --verbose --features kv_unstable_sval
-    - run: cargo test --verbose --features kv_unstable_serde
-    - run: cargo test --verbose --features "kv_unstable kv_unstable_std kv_unstable_sval kv_unstable_serde"
+    - run: cargo test --verbose --features kv
+    - run: cargo test --verbose --features kv_sval
+    - run: cargo test --verbose --features kv_serde
+    - run: cargo test --verbose --features "kv kv_std kv_sval kv_serde"
     - run: cargo run --verbose --manifest-path test_max_level_features/Cargo.toml
     - run: cargo run --verbose --manifest-path test_max_level_features/Cargo.toml --release
 
@@ -103,12 +103,12 @@ jobs:
         run: |
           rustup update nightly --no-self-update
           rustup default nightly
-      - run: cargo build --verbose -Z avoid-dev-deps --features kv_unstable
-      - run: cargo build --verbose -Z avoid-dev-deps --features "kv_unstable std"
-      - run: cargo build --verbose -Z avoid-dev-deps --features "kv_unstable kv_unstable_sval"
-      - run: cargo build --verbose -Z avoid-dev-deps --features "kv_unstable kv_unstable_serde"
-      - run: cargo build --verbose -Z avoid-dev-deps --features "kv_unstable kv_unstable_std"
-      - run: cargo build --verbose -Z avoid-dev-deps --features "kv_unstable kv_unstable_sval kv_unstable_serde"
+      - run: cargo build --verbose -Z avoid-dev-deps --features kv
+      - run: cargo build --verbose -Z avoid-dev-deps --features "kv std"
+      - run: cargo build --verbose -Z avoid-dev-deps --features "kv kv_sval"
+      - run: cargo build --verbose -Z avoid-dev-deps --features "kv kv_serde"
+      - run: cargo build --verbose -Z avoid-dev-deps --features "kv kv_std"
+      - run: cargo build --verbose -Z avoid-dev-deps --features "kv kv_sval kv_serde"
 
   minimalv:
     name: Minimal versions
@@ -119,12 +119,12 @@ jobs:
         run: |
           rustup update nightly --no-self-update
           rustup default nightly
-      - run: cargo build --verbose -Z minimal-versions --features kv_unstable
-      - run: cargo build --verbose -Z minimal-versions --features "kv_unstable std"
-      - run: cargo build --verbose -Z minimal-versions --features "kv_unstable kv_unstable_sval"
-      - run: cargo build --verbose -Z minimal-versions --features "kv_unstable kv_unstable_serde"
-      - run: cargo build --verbose -Z minimal-versions --features "kv_unstable kv_unstable_std"
-      - run: cargo build --verbose -Z minimal-versions --features "kv_unstable kv_unstable_sval kv_unstable_serde"
+      - run: cargo build --verbose -Z minimal-versions --features kv
+      - run: cargo build --verbose -Z minimal-versions --features "kv std"
+      - run: cargo build --verbose -Z minimal-versions --features "kv kv_sval"
+      - run: cargo build --verbose -Z minimal-versions --features "kv kv_serde"
+      - run: cargo build --verbose -Z minimal-versions --features "kv kv_std"
+      - run: cargo build --verbose -Z minimal-versions --features "kv kv_sval kv_serde"
 
   msrv:
     name: MSRV
@@ -135,7 +135,9 @@ jobs:
         run: |
           rustup update 1.60.0 --no-self-update
           rustup default 1.60.0
-      - run: cargo test --verbose --manifest-path tests/Cargo.toml
+      - run: |
+          cargo test --verbose --manifest-path tests/Cargo.toml
+          cargo test --verbose --manifest-path tests/Cargo.toml --features kv
 
   embedded:
     name: Embedded
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,7 +51,8 @@ kv_sval = ["kv", "value-bag/sval", "sval", "sval_ref"]
 kv_std = ["std", "kv", "value-bag/error"]
 kv_serde = ["kv_std", "value-bag/serde", "serde"]
 
-# Legacy: use `kv_*` instead
+# Deprecated: use `kv_*` instead
+# These `*_unstable` features will be removed in a future release
 kv_unstable = ["kv"]
 kv_unstable_sval = ["kv_sval"]
 kv_unstable_std = ["kv_std"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ rust-version = "1.60.0"
 edition = "2021"
 
 [package.metadata.docs.rs]
-features = ["std", "serde", "kv_unstable_std", "kv_unstable_sval", "kv_unstable_serde"]
+features = ["std", "serde", "kv_std", "kv_sval", "kv_serde"]
 
 [[test]]
 name = "filters"
@@ -46,12 +46,16 @@ release_max_level_trace = []
 
 std = []
 
-# requires the latest stable
-# this will have a tighter MSRV before stabilization
-kv_unstable = []
-kv_unstable_sval = ["kv_unstable", "value-bag/sval", "sval", "sval_ref"]
-kv_unstable_std = ["std", "kv_unstable", "value-bag/error"]
-kv_unstable_serde = ["kv_unstable_std", "value-bag/serde", "serde"]
+kv = []
+kv_sval = ["kv", "value-bag/sval", "sval", "sval_ref"]
+kv_std = ["std", "kv", "value-bag/error"]
+kv_serde = ["kv_std", "value-bag/serde", "serde"]
+
+# Legacy: use `kv_*` instead
+kv_unstable = ["kv"]
+kv_unstable_sval = ["kv_sval"]
+kv_unstable_std = ["kv_std"]
+kv_unstable_serde = ["kv_serde"]
 
 [dependencies]
 serde = { version = "1.0", optional = true, default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,6 +61,7 @@ value-bag = { version = "1.4", optional = true, default-features = false }
 
 [dev-dependencies]
 serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0"
 serde_test = "1.0"
 sval = { version = "2.1" }
 sval_derive = { version = "2.1" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,7 +48,7 @@ std = []
 
 # requires the latest stable
 # this will have a tighter MSRV before stabilization
-kv_unstable = ["value-bag"]
+kv_unstable = []
 kv_unstable_sval = ["kv_unstable", "value-bag/sval", "sval", "sval_ref"]
 kv_unstable_std = ["std", "kv_unstable", "value-bag/error"]
 kv_unstable_serde = ["kv_unstable_std", "value-bag/serde", "serde"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,10 +53,10 @@ kv_serde = ["kv_std", "value-bag/serde", "serde"]
 
 # Deprecated: use `kv_*` instead
 # These `*_unstable` features will be removed in a future release
-kv_unstable = ["kv"]
-kv_unstable_sval = ["kv_sval"]
-kv_unstable_std = ["kv_std"]
-kv_unstable_serde = ["kv_serde"]
+kv_unstable = ["kv", "value-bag"]
+kv_unstable_sval = ["kv_sval", "kv_unstable"]
+kv_unstable_std = ["kv_std", "kv_unstable"]
+kv_unstable_serde = ["kv_serde", "kv_unstable_std"]
 
 [dependencies]
 serde = { version = "1.0", optional = true, default-features = false }
--- a/README.md
+++ b/README.md
@@ -116,7 +116,7 @@ pub fn shave_the_yak(yak: &mut Yak) {
                 break;
             }
             Err(err) => {
-                warn!(err:error = err; "Unable to locate a razor, retrying");
+                warn!(err:err; "Unable to locate a razor, retrying");
             }
         }
     }
--- a/README.md
+++ b/README.md
@@ -106,17 +106,22 @@ If you enable the `kv` feature, you can associate structured data with your log 
 use log::{info, trace, warn};
 
 pub fn shave_the_yak(yak: &mut Yak) {
-    trace!(target = "yak_events", yak:serde = yak; "Commencing yak shaving");
+    // `yak:serde` will capture `yak` using its `serde::Serialize` impl
+    //
+    // You could also use `:?` for `Debug`, or `:%` for `Display`. For a
+    // full list, see the `log` crate documentation
+    trace!(target = "yak_events", yak:serde; "Commencing yak shaving");
 
     loop {
         match find_a_razor() {
             Ok(razor) => {
-                info!(razor = razor; "Razor located");
+                info!(razor; "Razor located");
                 yak.shave(razor);
                 break;
             }
-            Err(err) => {
-                warn!(err:err; "Unable to locate a razor, retrying");
+            Err(e) => {
+                // `e:err` will capture `e` using its `std::error::Error` impl
+                warn!(e:err; "Unable to locate a razor, retrying");
             }
         }
     }
--- a/README.md
+++ b/README.md
@@ -103,10 +103,10 @@ The executable itself may use the `log` crate to log as well.
 If you enable the `kv_unstable` feature, you can associate structured data with your log records:
 
 ```rust
-use log::{info, trace, warn, as_serde, as_error};
+use log::{info, trace, warn};
 
 pub fn shave_the_yak(yak: &mut Yak) {
-    trace!(target = "yak_events", yak = as_serde!(yak); "Commencing yak shaving");
+    trace!(target = "yak_events", yak:serde = yak; "Commencing yak shaving");
 
     loop {
         match find_a_razor() {
@@ -116,7 +116,7 @@ pub fn shave_the_yak(yak: &mut Yak) {
                 break;
             }
             Err(err) => {
-                warn!(err = as_error!(err); "Unable to locate a razor, retrying");
+                warn!(err:error = err; "Unable to locate a razor, retrying");
             }
         }
     }
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ The executable itself may use the `log` crate to log as well.
 
 ## Structured logging
 
-If you enable the `kv_unstable` feature, you can associate structured data with your log records:
+If you enable the `kv` feature, you can associate structured data with your log records:
 
 ```rust
 use log::{info, trace, warn};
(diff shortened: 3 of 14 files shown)


KodrAus:
We currently have `kv::source::Visitor` and `kv::value::Visit`. They're both visitor types (separate because they serve separate purposes), but should probably still be named consistently. @Thomasdezeeuw what do you think? Should we rename either `source::Visitor` or `value::Visit`? My preference would be to rename `value::Visit` to `value::Visitor`.

Thomasdezeeuw:
> We currently have `kv::source::Visitor` and `kv::value::Visit`. They're both visitor types (separate because they serve separate purposes), but should probably still be named consistently. @Thomasdezeeuw what do you think? Should we rename either `source::Visitor` or `value::Visit`? My preference would be to rename `value::Visit` to `value::Visitor`.

I think `Visitor` is more common? So I would go with that.

I would understand possible confusing between the two traits, so maybe we need to rename one of the two? What about renaming `source::Visitor` to `KeyValuePairs` or just `KeyValues`, to indicate its a collection of key-value pairs?

KodrAus:
> I would understand possible confusing between the two traits, so maybe we need to rename one of the two?

That's a fair point. Perhaps we could keep the names verb-y and use `VisitSource` and `VisitValue`?

Thomasdezeeuw:
> That's a fair point. Perhaps we could keep the names verb-y and use `VisitSource` and `VisitValue`?

Sounds good :+1:

KodrAus:
Ok, I think the last remaining thing here is macro syntax. We could adopt exactly `tracing`'s syntax of:

```
key = <modifier> arg
```

where `<modifier>` may be `?` for `fmt::Debug` or `%` for `fmt::Display`. I'm not a huge fan of leaning on sigils, or on prefixed modifiers here, because it leaves us almost no room for extensibility, but does have the benefit of being what's already there. I was thinking something like:

```
key = arg : <modifier>
```

where `<modifier`> may be blank for `fmt::Display`, `?` for `fmt::Debug`, or the literal text `serde` or `sval` for those respective libraries. In practice, it would look something like:

- `data`: captured using `ToValue`
- `data:`: captured using `fmt::Display`
- `data:?`: captured using `fmt::Debug`
- `data:serde`: captured using `serde::Serialize`
- `data:sval`: captured using `sval::Value`

We could use a simple lowering scheme so `data:?` lowered to `data:debug` and `data:` lowered to `data:display`, so it would be less easy to miss the presence of the `:`.

Another potential position could be on the key name:

```
key : <modifier> = arg
```

That would more easily let us expected `arg` to be a full expression in the future.

Any thoughts?

Thomasdezeeuw:
> ```
> key = arg : <modifier>
> ```
> 
> where `<modifier`> may be blank for `fmt::Display`, `?` for `fmt::Debug`, or the literal text `serde` or `sval` for those respective libraries. In practice, it would look something like:

I like this, feels similar to `write!` (and `print!`, etc.) macros.

>     * `data:serde`: captured using `serde::Serialize`
> 
>     * `data:sval`: captured using `sval::Value`

Small thing is versioning, in case serde/sval ever publish a v2, how do we want to handle that? We can go `serdeV1` or `serde` or something else with a version in it, but if a v2 never comes it will just be annoying. So maybe `serde`/`sval` means v1, and `serde2`/`sval2` could v2?

> We could use a simple lowering scheme so `data:?` lowered to `data:debug` and `data:` lowered to `data:display`, so it would be less easy to miss the presence of the `:`.

Because we'll be copying the standard formatting found in the std lib, I don't think will become an issue.

> Another potential position could be on the key name:
> 
> ```
> key : <modifier> = arg
> ```
> 
> That would more easily let us expected `arg` to be a full expression in the future.

How would be limiting ourselves using the above `arg : <modifier>`? I suppose we won't be able to support full expressions?

KodrAus:
> So maybe serde/sval means v1, and serde2/sval2 could v2?

That seems like a good scheme to me :+1: I spelled this out explicitly in `value-bag`, but it's not really meant to be user-facing. I think it's reasonable to make the "unversioned" ident refer to the current stable versions in `log`.

> How would be limiting ourselves using the above arg : <modifier>? I suppose we won't be able to support full expressions?

We'd just need to wrap full expressions in parenthesis or braces, because the `:` character can't appear directly after an expression in a macro rule.

KodrAus:
This is quite a big change but I think it's ready for a review now. Some things to call out:

- The macro syntax I ended up with was `key:modifier = value`, so `key:? = value` for debug for instance. This was easier to implement and didn't require breaking any existing users.
- Should we make the `kv::source` and `kv::value` modules private now there's no types in them that aren't exported in the root `kv` module? I think we probably should.

src/kv/mod.rs
```diff
@@ -11,16 +11,228 @@
...
+//! 
+//! The following capturing modifiers are supported:
+//! 
+//! - `:?` will capture the value using `Debug`.
+//! - `:debug` will capture the value using `Debug`.
+//! - `:%` will capture the value using `Display`.
+//! - `:display` will capture the value using `Display`.
+//! - `:error` will capture the value using `std::error::Error` (requires the `kv_unstable_error` feature).
+//! - `:sval` will capture the value using `sval::Value` (requires the `kv_unstable_sval` feature).
+//! - `:serde` will capture the value using `serde::Serialize` (requires the `kv_unstable_serde` feature).
```
Thomasdezeeuw, line 54: I like this :+1:

src/kv/mod.rs
```diff
@@ -11,16 +11,228 @@
...
+//! Individual values can be pulled from the source by their key:
+//!
+//! ```
+//! # fn main() -> Result<(), log::kv::Error> {
+//! use log::kv::{Source, Key, Value};
+//! # let record = log::Record::builder().key_values(&[("a", 1)]).build();
+//!
+//! // info!(a = 1; "Something of interest");
+//!
+//! let a: Value = record.key_values().get(Key::from("a")).unwrap();
```
Thomasdezeeuw, line 69: Nit: maybe add `assert_eq!(a, Value::from(1));` to make it super clear?

src/kv/mod.rs
```diff
@@ -11,16 +11,228 @@
...
+//! # Ok(())
+//! # }
+//! ```
+//!
+//! All key-values can also be enumerated using a [`VisitSource`]:
+//!
+//! ```
+//! # fn main() -> Result<(), log::kv::Error> {
+//! # let record = log::Record::builder().key_values(&[("a", 1), ("b", 2), ("c", 3)]).build();
```
Thomasdezeeuw, line 78: Nit: want to move this above the `info!` statement so they stay in sync?

src/kv/mod.rs
```diff
@@ -11,16 +11,228 @@
...
+//! ```
+//!
+//! To serialize a value to a format like JSON, you can also use either `serde` or `sval`:
+//!
+//! ```
+//! # fn main() -> Result<(), Box<dyn std::error::Error>> {
+//! # #[cfg(feature = "serde")]
+//! # {
+//! # use log::kv::Key;
+//! # #[derive(serde::Serialize)] struct Data { a: i32, b: bool, c: &'static str }
```
Thomasdezeeuw, line 192: I think we'll want to show this in the example so that people understand they need to implement/derive `serde::Serialize` for log (and `serde_json::to_string`)

src/kv/mod.rs
```diff
@@ -11,16 +11,228 @@
...
+//! # }
+//! # Ok(())
+//! # }
+//! ```
+//!
+//! The choice of serialization framework depends on the needs of the consumer.
+//! If you're in a no-std environment, you can use `sval`. In other cases, you can use `serde`.
+//! Log producers and log consumers don't need to agree on the serialization framework.
+//! A value can be captured using its `serde::Serialize` implementation and still be serialized
+//! through `sval` without losing any structure.
```
Thomasdezeeuw, line 211: Nit: maybe?
```suggestion
//! through `sval` without losing any structure or data.
```

src/kv/mod.rs
```diff
@@ -11,16 +11,228 @@
...
+//!
+//! Values can also always be formatted using the standard `Debug` and `Display`
+//! traits:
+//!
+//! ```
+//! # use log::kv::Key;
+//! # #[derive(Debug)] struct Data { a: i32, b: bool, c: &'static str }
```
Thomasdezeeuw, line 218: I would also show this in the example.

src/kv/source.rs
```diff
@@ -1,25 +1,65 @@
...
 
 use crate::kv::{Error, Key, ToKey, ToValue, Value};
 use std::fmt;
 
-/// A source of key-value pairs.
+/// A source of key-values.
 ///
 /// The source may be a single pair, a set of pairs, or a filter over a set of pairs.
-/// Use the [`Visitor`](trait.Visitor.html) trait to inspect the structured data
+/// Use the [`VisitSource`](trait.VisitSource.html) trait to inspect the structured data
```
Thomasdezeeuw, line 12: Nit: don't think it's needed any more. But if we have it in other places, then let's leave it and do it some other time for all links.
```suggestion
/// Use the [`VisitSource`] trait to inspect the structured data
```

src/kv/source.rs
```diff
@@ -34,14 +74,14 @@ pub trait Source {
...
+    /// Count the number of key-values that can be visited.
     ///
     /// # Implementation notes
     ///
-    /// A source that knows the number of key-value pairs upfront may provide a more
+    /// A source that knows the number of key-values upfront may provide a more
     /// efficient implementation.
     ///
-    /// A subsequent call to `visit` should yield the same number of key-value pairs
+    /// A subsequent call to `visit` should yield the same number of key-values
```
Thomasdezeeuw, line 84: Nit.
```suggestion
    /// A subsequent call to `visit` should yield the same number of key-values.
```

Thomasdezeeuw (approved):
I like it.

The only concern I have is that people find the modified on the key odd/unexpected, when coming from the `std::fmt` implementation. That might be worth calling out in the documentation.

Cargo.toml
```diff
@@ -48,7 +48,7 @@ std = []
 
 # requires the latest stable
 # this will have a tighter MSRV before stabilization
-kv_unstable = ["value-bag"]
+kv_unstable = []
 kv_unstable_sval = ["kv_unstable", "value-bag/sval", "sval", "sval_ref"]
 kv_unstable_std = ["std", "kv_unstable", "value-bag/error"]
 kv_unstable_serde = ["kv_unstable_std", "value-bag/serde", "serde"]
```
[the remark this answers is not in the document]
KodrAus: It will, yes :+1: We technically don't need `value-bag` for capturing `std::error::Error`s, but do for `serde` or `sval`.

src/kv/source.rs
KodrAus: Good call. I think I still need to do a pass over links and make sure they're all resolved correctly so will simplify them wherever possible.

src/__private_api.rs
```diff
@@ -87,3 +84,42 @@ pub fn log<'a, K>(
...
+
+#[cfg(feature = "kv_unstable")]
+mod kv_support {
+    use super::*;
+
+    use crate::kv;
+
+    pub type Value<'a> = kv::Value<'a>;
+
+    pub fn capture_to_value<'a, V: kv::value::ToValue + ?Sized>(v: &'a &'a V) -> Value<'a> {
```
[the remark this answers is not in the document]
KodrAus: I left a comment in the source on why we do this. It's just so we can capture `T: ?Sized` while still erasing it behind a dyn Trait. Passing in the double reference in just helps the lifetimes work out.

KodrAus:
The

```
err:error = err
```

way of capturing an error using the `std::error::Error` trait is maybe something we can improve on someday 🙂

KodrAus:
I think the last thing remaining here is what to do about the `_unstable` suffix on our features. I was thinking making them just shims for the unsuffixed versions, but continuing to accept them so that we don't break every current user of the feature. What do you think @Thomasdezeeuw?

.github/workflows/main.yml
```diff
@@ -44,10 +44,10 @@ jobs:
     - run: cargo test --verbose --all-features
     - run: cargo test --verbose --features serde
     - run: cargo test --verbose --features std
-    - run: cargo test --verbose --features kv_unstable
```
Thomasdezeeuw, line 47: We could start using [cargo-hack](https://crates.io/crates/cargo-hack) for this at some point (for another pr).

Cargo.toml
```diff
@@ -46,25 +46,30 @@ release_max_level_trace = []
...
+kv = []
+kv_sval = ["kv", "value-bag/sval", "sval", "sval_ref"]
+kv_std = ["std", "kv", "value-bag/error"]
+kv_serde = ["kv_std", "value-bag/serde", "serde"]
+
+# Legacy: use `kv_*` instead
+kv_unstable = ["kv"]
+kv_unstable_sval = ["kv_sval"]
+kv_unstable_std = ["kv_std"]
+kv_unstable_serde = ["kv_serde"]
```
Thomasdezeeuw, line 58: To bad there is only `compile_error!`, not `compile_warn!` otherwise it would be nice to add that.

Thomasdezeeuw: Actually on second though... we're going to break a lot of (unstable) code already by removing the `as_serde` macros, do we just rip the bandaid in one go and break everything? Or should we add those macros back if [`kv_unstable`] is used (and mark them as `#[deprecated]`)?

Thomasdezeeuw (approved):
> I think the last thing remaining here is what to do about the `_unstable` suffix on our features. I was thinking making them just shims for the unsuffixed versions, but continuing to accept them so that we don't break every current user of the feature. What do you think @Thomasdezeeuw?

We can, but we a also have other breakages, e.g. the removal of the `as_{display,debug, etc.}` macros. Some of the API on `Value` has changes, which is also breaking.

So I'm wondering if we should keep some of the old API, if possible, if `kv_unstable` is used? Marking it all as deprecated show it at least generates warnings (I don't know if that is possible for features). Then e.g. 6 months after the stable release we can remove the `_unstable` feature and all it's related API? This would give people some time to migrate their code.

Cargo.toml
KodrAus: I think this is a reasonable thing to do :+1:

Unfortunately you can’t mark them as deprecated but we can also re-export the renamed visitor traits under their old names when the old `_unstable` features are enabled.

KodrAus:
How does this sound as a plan for stabilization:

1. Keep all old APIs when the `kv_unstable` feature is enabled, deprecating wherever possible.
2. Set a time period where we’ll remove the `kv_unstable` feature altogether and all those deprecated items. If we keep rough tabs on public code using the old features then we could get some idea of when would be a good time to do this.

The idea being new code and any maintained in that time period should move from the unstable to the stable API without much fuss, but there will be a point where the bandaid comes off completely. The only unfortunate part is we can’t deprecate old trait names so we don’t have a good way of making users of them aware things have changed.

Thomasdezeeuw:
> How does this sound as a plan for stabilization:

Sounds good to me :+1:

KodrAus:
In this latest round of changes I've restored the old APIs under their respective `*_unstable` feature flags and added support for shorthand capturing, so instead of having to write `info!(a = a; "")` you can write `info!(a; "")`.

KodrAus:
Ok, I think this is ready for its last review now. Thank you so much for all the time you've invested in helping push this forwards so far @Thomasdezeeuw 🙏 We'd never have reached this point without you.

README.md
```diff
@@ -116,7 +116,7 @@ pub fn shave_the_yak(yak: &mut Yak) {
                 break;
             }
             Err(err) => {
-                warn!(err:error = err; "Unable to locate a razor, retrying");
+                warn!(err:err; "Unable to locate a razor, retrying");
```
Thomasdezeeuw, line 119: Nit: maybe this makes the example a little confusing? Making it look like `:` is the separator for key-value pairs?

Thomasdezeeuw (approved):
Nice! Let's merge.

I'm gonna go over the API with https://github.com/Enselic/cargo-public-api/ see what, if any, difference there are

Cargo.toml
```diff
@@ -46,25 +46,31 @@ release_max_level_trace = []
...
+kv_sval = ["kv", "value-bag/sval", "sval", "sval_ref"]
+kv_std = ["std", "kv", "value-bag/error"]
+kv_serde = ["kv_std", "value-bag/serde", "serde"]
+
+# Deprecated: use `kv_*` instead
+# These `*_unstable` features will be removed in a future release
+kv_unstable = ["kv"]
+kv_unstable_sval = ["kv_sval"]
+kv_unstable_std = ["kv_std"]
+kv_unstable_serde = ["kv_serde"]
```
Thomasdezeeuw, line 59: I think we need to enable these features.

```suggestion
kv_unstable = ["kv", "value-bag"]
kv_unstable_sval = ["kv_sval", "kv_unstable"]
kv_unstable_std = ["kv_std", "kv_unstable"]
kv_unstable_serde = ["kv_serde", "kv_unstable_std"]
```

This resolved most of the compilation errors for my crate, except for one. It can't find `log::kv::value::Visit`.

src/kv/value.rs
```diff
@@ -765,13 +596,586 @@ where
...
+        note = "downcasting has been removed; log an issue at https://github.com/rust-lang/log/issues if this is something you rely on"
+    )]
+    pub fn downcast_ref<T: 'static>(&self) -> Option<&T> {
+        None
+    }
+}
+
+// NOTE: Deprecated; but aliases can't carry this attribute
+#[cfg(feature = "kv_unstable")]
+pub use VisitValue as Visitor;
```
Thomasdezeeuw, line 1121: Should this be visit?
```suggestion
pub use VisitValue as Visit;
```

README.md
KodrAus: That's a good point. It could also look like it's trying to be a pattern binding 🤔 I'll try to clear it up.

PR #615, opened 2024-02-10
Add dependabot

oriontvv:
This PR adds dependabot config

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,15 @@
+version: 2
+updates:
+  - package-ecosystem: "cargo"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    open-pull-requests-limit: 10
+  - package-ecosystem: "docker"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"


Thomasdezeeuw:
I'm -1 on this. I don't think we need to update our dependencies unless we need a specific fix or feature. Updating dependencies with patch versions in libraries can even become a problem when those patch version have problems. I'd say let the binary (not libraries) define the dependency versions.

oriontvv:
As I understand - if patch version is not specified for dependency - it would NOT be notified about any patch update. Anyway, we could setup `update-types: minor` (or even major) [documentation](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups). Main motivation for this is that the newer library version in the binary - the less DIFFERENT transitive versions of library would in the result binary build. And it would save a lot of build time and total binary size. Also there is side benefit that dependabot can notify about major security updates.

KodrAus:
I'm personally not keen on Dependabot either, but it looks like it's used by other libraries in the `rust-lang` org: https://github.com/rust-lang/cc-rs/blob/main/.github/dependabot.yml

If there's some org-wide policy to use it then I think we should add it here too using the same configuration.

BurntSushi:
> If there's some org-wide policy to use it

I'm not aware of one. I would resist pretty strongly personally.

> Main motivation for this is that the newer library version in the binary - the less DIFFERENT transitive versions of library would in the result binary build.

This is only applicable for semver incompatible releases. Which I assume is quite rare for `log`.

KodrAus:
That being the case I’ll go ahead and close this one as something we won’t be pursuing in `log`. Thanks for submitting the PR though!

PR #618, opened 2024-02-27
CI should detect when breaking changes are introduced

nyurik:
Note that this uses a separate semver job - which may be an overkill - perhaps it should be added to the clippy job?  I wouldn't want it to compile so many times in different VMs unless needed.

See https://github.com/obi1kenobi/cargo-semver-checks?tab=readme-ov-file#quick-start

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -81,6 +81,14 @@ jobs:
     - run: cargo clippy --verbose --manifest-path test_max_level_features/Cargo.toml
     - run: cargo clippy --verbose --manifest-path tests/Cargo.toml
 
+  semver:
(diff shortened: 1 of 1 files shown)


KodrAus:
Thanks for the PR @nyurik! This seems like something worth doing going forwards. Would you be happy to port this to using `cargo-semver-checks` directly with `cargo install` rather than through the `cargo-semver-checks-action` GitHub action? That way we have a clearer idea of what each job is doing and what actions we're depending on. The tool itself looks good to me 🙂

KodrAus:
Just coming back through some triage, it's been a while here so I'll go ahead and close this one, but we can revisit in the future.

PR #621, opened 2024-03-23
Add links to `colog` crate

chrivers:
Hi 👋

I'm the maintained or the [colog](https://crates.io/crates/colog) crate.

My crate provides a quick way to set up colored logging with a (subjectively) good and clean look.

I've just released version 1.3, which in my opinion is hugely improved.

Everything is now extensively documented, and there are 7 included examples on using the crate. The crate has been around for 8 years, so it's not exactly abandonware ;-)

I would greatly appreciate it, if you would include it in the list of possible logging crates in your documentation (see this PR)

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,7 @@ There are many available implementations to choose from, here are some options:
 
 * Simple minimal loggers:
     * [`env_logger`](https://docs.rs/env_logger/*/env_logger/)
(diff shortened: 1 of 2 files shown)


KodrAus:
Thanks @chrivers! That's a nice tidy implementation too :+1:

PR #625, opened 2024-03-29
Replace deprecated std::sync::atomic::spin_loop_hint()

Catamantaloedis:
I noticed that `log`'s MSRV is now 1.60.0+, which seems to indicate that the TODO that was waiting for MSRV 1.49.0 can now be fulfilled.


KodrAus (approved):
Thanks @Catamantaloedis!

PR #627, opened 2024-03-31
Check usage of max_level features

### LINKED ISSUE #481: Consider emitting a diagnostic message when multiple incompatible log level features are used
So today I learned the hard way that crate features are additive. If a program directly or indirectly through other dependencies, depends on a crate of the same version multiples times with different features, it will use the union of features. It does not introduce multiple copies of the crate like what happens with a version mismatch. This is pretty logical behavior I suppose, but it's a bit annoying when you have features that are mutually incompatible (like static max output level in logs). 

In my specific case, I had some far flung dependency setting `release_max_level_info`, when I was trying to do trace level logging. 

The real problem here was a dependency library setting log level features. I'll sort that out with that library, but when I was trying to figure out what was going on, I hadn't considered that multiple features were active simultaneously until I did a bunch of digging (verbose builds, grepping through `.cargo` etc.). I could see legitimate use cases for multiple log levels perhaps, so I wouldn't make this an error. A compiler warning or message saying something to the effect of "hey you've got both `release_max_level_info` and `release_max_level_trace`, we're going to go with `info`" (but obviously more generic than that) could be very helpful.

Thomasdezeeuw:
With this change log will to fail to compile if two or more (release_)max_level_* features are used.

Closes #481

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -41,7 +41,7 @@ jobs:
         rustup default ${{ matrix.rust }}
     - run: cargo test --verbose
     - run: cargo test --verbose --no-default-features
-    - run: cargo test --verbose --all-features
+    - run: cargo test --verbose --features std,serde,sval,sval_ref,value-bag,kv,kv_std,kv_sval,kv_serde
     - run: cargo test --verbose --features serde
     - run: cargo test --verbose --features std
     - run: cargo test --verbose --features kv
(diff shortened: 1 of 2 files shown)


Thomasdezeeuw:
Well, at least we know it works: https://github.com/rust-lang/log/actions/runs/8500090037/job/23281579006?pr=627#step:6:51, but accepting this means `--all-features` won't work any more.

KodrAus (approved):

PR #610, opened 2024-01-02
Use `Acquire` ordering for initialization check

AngelicosPhosphoros:
Reasoning is provided in a comment.

PR #632, opened 2024-06-25
Loosen orderings for logger initialization

KodrAus:
This is a reworking of #520 using the same orderings. I think the main improvement was made in #610 so if we want to punt on this I think we reasonably can.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -341,10 +341,6 @@
 #![warn(missing_docs)]
 #![deny(missing_debug_implementations, unconditional_recursion)]
 #![cfg_attr(all(not(feature = "std"), not(test)), no_std)]
-// When compiled for the rustc compiler itself we want to make sure that this is
-// an unstable crate
-#![cfg_attr(rustbuild, feature(staged_api, rustc_private))]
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -341,10 +341,6 @@
 #![warn(missing_docs)]
 #![deny(missing_debug_implementations, unconditional_recursion)]
 #![cfg_attr(all(not(feature = "std"), not(test)), no_std)]
-// When compiled for the rustc compiler itself we want to make sure that this is
-// an unstable crate
-#![cfg_attr(rustbuild, feature(staged_api, rustc_private))]
```
KodrAus, line 346: I believe this is a left-over. As far as I know `log` isn't used in the compiler anymore, but will actually go and check.

KodrAus: Looks like it's used in some tools: https://github.com/search?q=log+repo%3Arust-lang%2Frust+path%3A*.toml&type=code&ref=advsearch

Looking at other crates.io dependencies, they don't have any special handling to appear in the compiler anymore so I think this is still safe to remove.

Thomasdezeeuw (approved):
Based on the precedent set in https://github.com/rust-lang/rust/pull/122729 this LGTM.

PR #633, opened 2024-06-25
Use Location::caller() for file and line info

KodrAus:
Rework of #599. I've followed the suggestion in https://github.com/rust-lang/log/pull/599#issuecomment-2143472357 to pass a `panic::Location` to our non-inlined private log function in the hope it will avoid bloating callsites.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -341,10 +341,6 @@
 #![warn(missing_docs)]
 #![deny(missing_debug_implementations, unconditional_recursion)]
 #![cfg_attr(all(not(feature = "std"), not(test)), no_std)]
-// When compiled for the rustc compiler itself we want to make sure that this is
-// an unstable crate
-#![cfg_attr(rustbuild, feature(staged_api, rustc_private))]
-#![cfg_attr(rustbuild, unstable(feature = "rustc_private", issue = "27812"))]
 
 #[cfg(any(
     all(feature = "max_level_off", feature = "max_level_error"),
--- a/src/__private_api.rs
+++ b/src/__private_api.rs
@@ -3,7 +3,8 @@
 use self::sealed::KVs;
 use crate::{Level, Metadata, Record};
 use std::fmt::Arguments;
-pub use std::{file, format_args, line, module_path, stringify};
+use std::panic::Location;
+pub use std::{format_args, module_path, stringify};
 
 #[cfg(not(feature = "kv"))]
 pub type Value<'a> = &'a str;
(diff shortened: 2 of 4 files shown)


src/lib.rs
```diff
@@ -341,10 +341,6 @@
 #![warn(missing_docs)]
 #![deny(missing_debug_implementations, unconditional_recursion)]
 #![cfg_attr(all(not(feature = "std"), not(test)), no_std)]
-// When compiled for the rustc compiler itself we want to make sure that this is
```
KodrAus, line 344: See #632

Thomasdezeeuw (approved):
LGTM.

I do wondering what this does to the binary size, but I'm personally not that affected by it, so I have no setup to test is properly.

EFanZh:
I haven’t tested it on a real project, but based on [this assembly comparison](https://godbolt.org/z/fsMEG3q5h), I think this PR will cause some binary size bloat.

KodrAus:
@EFanZh I think most changes we can make are likely to impact binary size in one way or another and we don't have the CI infrastructure to measure this so am hesitant to block this change purely based on the possible increase. It sets a precedent that blocks pretty much any future work.

Having said that, I do think binary size is a useful metric for us to track, so if this is really important to you then I think we'd happily accept a PR that introduced a CI job and test project to track how binary size changes. That way we can actually evaluate the impact of changes as they come through.

EFanZh:
Currently, binary size is less important to me than before, I just want to provide some possibly useful information, in case someone else is interested or affected by it.

As for the CI job, I think it will need a dedicated data server for storing the binary size history data, which I think it is a too big project for me.

Thomasdezeeuw:
> As for the CI job, I think it will need a dedicated data server for storing the binary size history data, which I think it is a too big project for me.

We could use GitHub Actions cache for that: https://github.com/actions/cache. But totally fair if you don't want to take that on.

I think we're good to merge then.

EFanZh:
For Rust version 1.79+, we can employ a similar optimization method: <https://godbolt.org/z/GaExEz9xj>.

EFanZh:
For Rust version 1.79+, with inline `const` and `Location::caller` being stabilized, we might be able to employ the same optimization: <https://godbolt.org/z/8Gn4Ycne8>.

KodrAus:
> As for the CI job, I think it will need a dedicated data server for storing the binary size history data, which I think it is a too big project for me.

Oh yes that would be a lot. I was imagining something less sophisticated. Just a test that generates a project with 1,000 log statements in it, compiles and strips it, and asserts the resulting binary is less than some constant value. If the size changed then we'd need to update that constant, giving us a history of when it changed and why. Something we can keep in mind for the future 🙂

PR #638, opened 2024-08-02
Add logforth to implementation

tisonkun:
Its abstractions are as powerful as `fern` and I integrate it with a few application to see that things work well.

PR #642, opened 2024-11-13
docs: Add logforth as an impl

tisonkun:
This is added in https://github.com/rust-lang/log/pull/638.


KodrAus (approved):
Thanks @tisonkun!

PR #643, opened 2024-11-13
Add correct lifetime to kv::Value::to_borrowed_str

stevenroose:


KodrAus (approved):
Thanks @stevenroose. There’s no reason this shouldn’t be `'v`

PR #644, opened 2024-11-14
Implement conversions between Option<Level> and LevelFilter

joshka:
This is nice to have when working with the `log` crate, as it allows
for more ergonomic conversions between the two types. Specifically, This
allows type bounds that require Into<LevelFilter> or Into<Option<Level>>
to be used. This in turn makes it possible to write more generic code
that can work with tracing and log crates interchangeably.

Specifically, this supports some ideas in
https://github.com/clap-rs/clap-verbosity-flag/issues/121

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -635,6 +635,21 @@ impl FromStr for LevelFilter {
     }
 }
 
+impl From<LevelFilter> for Option<Level> {
+    fn from(level: LevelFilter) -> Option<Level> {
+        level.to_level()
+    }
+}
+
+impl From<Option<Level>> for LevelFilter {
+    fn from(level: Option<Level>) -> LevelFilter {
+        match level {
+            Some(level) => level.to_level_filter(),
+            None => LevelFilter::Off,
+        }
+    }
+}
+
 impl fmt::Display for LevelFilter {
     fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
         fmt.pad(self.as_str())


tisonkun (commented):
While `From<LevelFilter> for Option<Level>` is trivial and straightforward, `None` to `LevelFilter::Off` may not. I support different system may have different opinion for the "default" level filter.

tisonkun:
You may write:

```
opt_level_filter.unwrap_or(LevelFilter::Off).into()
```

src/lib.rs
```diff
@@ -635,6 +635,21 @@ impl FromStr for LevelFilter {
...
+    fn from(level: LevelFilter) -> Option<Level> {
+        level.to_level()
+    }
+}
+
+impl From<Option<Level>> for LevelFilter {
+    fn from(level: Option<Level>) -> LevelFilter {
+        match level {
+            Some(level) => level.to_level_filter(),
+            None => LevelFilter::Off,
```
Thomasdezeeuw, line 648: I'm sure this is always going to be obvious.

joshka:
> While `From<LevelFilter> for Option<Level>` is trivial and straightforward, `None` to `LevelFilter::Off` may not. I suppose different system may have different opinion for the "default" level filter.

Is None indicating a default, or the absence of a level? The benefit of making None -> Off is that it round trips. Perhaps a doc comment about this would be sufficient.

> You may write:
> 
> ```
> opt_level_filter.unwrap_or(LevelFilter::Off).into()
> ```

There's no `Option<LevelFilter>` in this change. Can you clarify what you meant here?

src/lib.rs
joshka: Did you leave out "not" or are you disagreeing with the comment above here and saying that this is actually fine and a reasonable choice to convert `Option<Level>::None` to `LevelFilter::Off`?

joshka:
BTW, I solved the underlying need for this through another means (changed the trait associated type bounds), so I'm unblocked from an implementation perspective for now at least.

KodrAus:
Thanks for the PR @joshka! Since you're unblocked in another way I'll go ahead and close this one. The semantics you've chosen are reasonable, but when they're hidden in standard conversions they'll always trip _somebody_ up so we're best off not introducing them where they're not fully unambiguous.

joshka:
No problem - thanks for the feedback. I don't generally use log myself as I prefer the tracing ecosystem, so the implementation here was mainly for completeness and consistency with the same approach in the tracing crate, rather than personal usefulness.

PR #646, opened 2024-11-17
Add clang_log implementation

DDAN-17:
Just a simple logger I made


KodrAus:
Thanks @DDAN-17! Just a heads up, it looks like your `Cargo.toml` has the wrong value for your repository URL. It's pointing to https://github.com/DDAN-17/clang-log instead of https://github.com/DDAN-17/clang_log. Honestly... Rust should've just required crate names be valid identifiers, but that's a topic for another day 😄

DDAN-17:
Thanks! I just fixed that now.

PR #648, opened 2024-12-06
Bind lifetimes of &str returned from Key by the lifetime of 'k rather than the lifetime of the Key struct

gbbosak:
This allows things inside a Visitor to persist these strings for the lifetime of 'k without having to copy them.

### DIFF
--- a/src/kv/key.rs
+++ b/src/kv/key.rs
@@ -45,7 +45,7 @@ impl<'k> Key<'k> {
     }
 
     /// Get a borrowed string from this key.
-    pub fn as_str(&self) -> &str {
+    pub fn as_str(&self) -> &'k str {
         self.key
     }
 }
@@ -57,7 +57,7 @@ impl<'k> fmt::Display for Key<'k> {
 }
 
 impl<'k> AsRef<str> for Key<'k> {
-    fn as_ref(&self) -> &str {
+    fn as_ref(&self) -> &'k str {
         self.as_str()
     }
 }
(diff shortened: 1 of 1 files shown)


src/kv/key.rs
```diff
@@ -57,7 +57,7 @@ impl<'k> fmt::Display for Key<'k> {
 }
 
 impl<'k> AsRef<str> for Key<'k> {
-    fn as_ref(&self) -> &str {
+    fn as_ref(&self) -> &'k str {
```
dtolnay, line 60: This has no effect unless you add `#[refine]` (https://rust-lang.github.io/rfcs/3245-refined-impls.html).

Thomasdezeeuw (approved):

Thomasdezeeuw:
Thanks @gbbosak

src/kv/key.rs
KodrAus: Hmm, it almost seems weird this compiles at all :eyes: I hadn't seen `#[refine]` before. Thanks for pointing it out!

KodrAus:
I should've left a comment on the type suggesting why it was that way, but we intentionally hadn't used `'k` here so we were free to change the internals of `Key` to use something like `Cow<'k, str>` to support key-value sources that need to buffer their keys. I'm ok with tightening it up and not doing that if we decide it's not that important, but worth calling out that's what we lose here.

Thomasdezeeuw:
@KodrAus do we want to revert the pr to keep that option open?

KodrAus:
@Thomasdezeeuw I think we should, and provide a way to get `Option<&'k str>` from a key so you still can try if you need to.

PR #652, opened 2024-12-21
Add Ftail implementation

tjardoo:
Adds [Ftail](https://crates.io/crates/ftail) to the list of 'simple minimal loggers'.

Ftail is a simple log implementation that allows logging to the console, a single file, a daily rotating file and to add your own custom option.

Thanks!


KodrAus:
Thanks @tjardoo!

Issue #654, opened 2025-01-09
kv_std should be broken down into kv + std?


tisonkun:
That said, with `log = { version = "0.4", features = ["std", "kv"] }`, we have:

```rust
#[cfg(feature = "std")]
mod std_support {
    use super::*;

    use std::borrow::Cow;

    impl ToKey for String {
        fn to_key(&self) -> Key {
            Key::from_str(self)
        }
    }

    impl<'a> ToKey for Cow<'a, str> {
        fn to_key(&self) -> Key {
            Key::from_str(self)
        }
    }
}
```

But the same for `ToValue` is filtered out:

```rust
#[cfg(feature = "kv_std")]
mod std_support {
    use std::borrow::Cow;
    use std::rc::Rc;
    use std::sync::Arc;

    use super::*;

    impl<T> ToValue for Box<T>
    where
        T: ToValue + ?Sized,
    {
        fn to_value(&self) -> Value {
            (**self).to_value()
        }
    }

    impl<T> ToValue for Arc<T>
    where
        T: ToValue + ?Sized,
    {
        fn to_value(&self) -> Value {
            (**self).to_value()
        }
    }

    impl<T> ToValue for Rc<T>
    where
        T: ToValue + ?Sized,
    {
        fn to_value(&self) -> Value {
            (**self).to_value()
        }
    }

    impl ToValue for String {
        fn to_value(&self) -> Value {
            Value::from(&**self)
        }
    }

    impl<'v> ToValue for Cow<'v, str> {
        fn to_value(&self) -> Value {
            Value::from(&**self)
        }
    }

    impl<'v> Value<'v> {
        /// Try convert this value into a string.
        pub fn to_cow_str(&self) -> Option<Cow<'v, str>> {
            self.inner.to_str()
        }
    }

    impl<'v> From<&'v String> for Value<'v> {
        fn from(v: &'v String) -> Self {
            Value::from(&**v)
        }
    }
}
```

I'm willing to send a PR to change the feature gate of ToValue's std_support to "std", but I'd first open an issue to see if anything I missed.

KodrAus:
Hi @tisonkun :wave: We ended up with composite features like `kv_std` because it wasn't possible originally to enable a feature in a dependency based on the presence of two unrelated Cargo features in a crate. You can do that now with `dependency?/feature` syntax in `Cargo.toml`, and it looks like our MSRV is high enough to make use of it.

[...]

PR #657, opened 2025-01-10
Relax feature flag for value's std_support

tisonkun:
This closes https://github.com/rust-lang/log/issues/654#issuecomment-2581445570.

While I agree @KodrAus's comment at https://github.com/rust-lang/log/issues/654#issuecomment-2581494083, this is not a generally breaking down kv_x to kv + x.

But that value's `std_support` can work with kv + std. `kv_std` introduces an extra feature of `value-bag/error`, which is only used for:

```rust
    pub fn to_borrowed_error(&self) -> Option<&(dyn std::error::Error + 'static)> {
        self.inner.to_borrowed_error()
    }
```

above.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -47,6 +47,7 @@ jobs:
     - run: cargo test --verbose --features kv
     - run: cargo test --verbose --features kv_sval
     - run: cargo test --verbose --features kv_serde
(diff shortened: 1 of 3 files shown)


KodrAus:
Thanks @tisonkun. I think this specific change is reasonable to make :+1:

PR #659, opened 2025-01-10
Use cargo hack in CI to test all feature combinations

### LINKED ISSUE #658: 0.4.23 causes build failures
```rust
error[E0599]: no method named `to_str` found for struct `ValueBag` in the current scope
   --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/log-0.4.23/src/kv/value.rs:446:20
    |
446 |         self.inner.to_str()
    |                    ^^^^^^
    |
help: there is a method `to_string` with a similar name
    |
446 |         self.inner.to_string()
    |                    ~~~~~~~~~

```

CI failure: https://github.com/delta-io/delta-rs/actions/runs/12696166185/job/35422694933#step:6:31

KodrAus:
Closes #658 

Looks like a feature combination broke in `0.4.23`, so I've added [`cargo hack`](https://github.com/taiki-e/cargo-hack?tab=readme-ov-file#--feature-powerset) to our CI to make sure we test all possible combinations.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,11 +44,11 @@ release_max_level_info  = []
 release_max_level_debug = []
 release_max_level_trace = []
 
-std = []
+std = ["value-bag?/error"]
 
 kv = []
 kv_sval = ["kv", "value-bag/sval", "sval", "sval_ref"]
-kv_std = ["std", "kv", "value-bag/error"]
+kv_std = ["std", "kv"]
 kv_serde = ["kv_std", "value-bag/serde", "serde"]
 
 # Deprecated: use `kv_*` instead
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -44,7 +44,7 @@ release_max_level_info  = []
 release_max_level_debug = []
 release_max_level_trace = []
 
-std = []
+std = ["value-bag?/std"]
```
Thomasdezeeuw, line 47: Is this supported in Rust 1.60 (our MSRV)?

Thomasdezeeuw (approved):

Cargo.toml
tisonkun: https://doc.rust-lang.org/cargo/reference/features.html

According to the doc linked, it's available starting from 1.60 so it should be fine.

tisonkun (approved):
Thanks for fixing this. And sorry I done catch this issue in my pr.

KodrAus:
@tisonkun No problems! It's our fault for not catching it in CI. I've been meaning to add `cargo-hack` here for a while, it's a great way to test projects with lots of feature flags.

PR #664, opened 2025-02-01
Add an optional logger param

### LINKED ISSUE #645: Question: meta logging in logger?
Some `log` implementations support logger pipeline. For example, in [`logforth`](https://github.com/fast/logforth) you can write:

```rust
use log::LevelFilter;
use logforth::append;

logforth::builder()
    .dispatch(|d| {
        d.filter(LevelFilter::Error)
            .append(append::Stderr::default())
    })
    .dispatch(|d| {
        d.filter(LevelFilter::Info)
            .append(append::Stdout::default())
    })
    .apply();

log::error!("Hello error!");
log::warn!("Hello warn!");
log::info!("Hello info!");
log::debug!("Hello debug!");
log::trace!("Hello trace!");
```

When I'm trying to add some complex appenders, for example, Kafka appender, it's natural to use a Kafka client lib. However, such a client lib may log inside also, which can result in an infinite recursive logging if such an appender is installed globally.

I did some investigation and how a possible solution, but would like to open this issue for discussion:

1. Filter certain target in specific appender:

```rust
impl log::Log KafkaAppender {
  fn log(&self, record: &Record) {
    if record.target().start_with("rdkafka") { return; }
  }
}
```

This is, however, both brittle and may filter logs from user code. I don't know how targets are defined and if we can reliably depend on that matches.

tisonkun:
This closes #645 

@KodrAus I noticed that we previously introduced `target` for similar usage, but switching to a different logger is different.

This patch is for preview. I'm not quite sure how we pass the customized logger. The current implementation seems good. I think the `log` crate itself isn't suitable to hold a global `name -> logger` registry: that would be too heavy and too specific.

### DIFF
--- a/src/__private_api.rs
+++ b/src/__private_api.rs
@@ -1,7 +1,7 @@
 //! WARNING: this is not part of the crate's public API and is subject to change at any time
 
 use self::sealed::KVs;
-use crate::{Level, Log, Metadata, Record};
+use crate::{logger, Level, Log, Metadata, Record};
 use std::fmt::Arguments;
 use std::panic::Location;
 pub use std::{format_args, module_path, stringify};
@@ -34,6 +34,28 @@ impl<'a> KVs<'a> for () {
 
 // Log implementation.
 
+/// The global logger proxy.
+///
+/// This zero-sized type implements the [`Log`] trait by forwarding calls
+/// to the logger registered with the `set_boxed_logger` or `set_logger`
+/// methods if there is one, or a nop logger as default.
+#[derive(Copy, Clone, Default, Debug)]
+pub struct GlobalLogger;
+
+impl Log for GlobalLogger {
+    fn enabled(&self, metadata: &Metadata) -> bool {
+        logger().enabled(metadata)
+    }
+
+    fn log(&self, record: &Record) {
+        logger().log(record)
+    }
+
+    fn flush(&self) {
+        logger().flush()
+    }
+}
+
 fn log_impl<L: Log>(
     logger: L,
     args: Arguments,
@@ -63,7 +85,7 @@ fn log_impl<L: Log>(
 }
 
 pub fn log<'a, K, L>(
-    logger: L,
+    logger: &L,
     args: Arguments,
     level: Level,
     target_module_path_and_loc: &(&str, &'static str, &'static Location),
--- a/src/__private_api.rs
+++ b/src/__private_api.rs
@@ -1,7 +1,7 @@
 //! WARNING: this is not part of the crate's public API and is subject to change at any time
 
 use self::sealed::KVs;
-use crate::{Level, Metadata, Record};
+use crate::{Level, Log, Metadata, Record};
 use std::fmt::Arguments;
 use std::panic::Location;
 pub use std::{format_args, module_path, stringify};
@@ -34,7 +34,8 @@ impl<'a> KVs<'a> for () {
 
 // Log implementation.
 
-fn log_impl(
+fn log_impl<L: Log>(
+    logger: &L,
     args: Arguments,
     level: Level,
     &(target, module_path, loc): &(&str, &'static str, &'static Location),
@@ -58,22 +59,30 @@ fn log_impl(
     #[cfg(feature = "kv")]
     builder.key_values(&kvs);
 
-    crate::logger().log(&builder.build());
+    logger.log(&builder.build());
 }
 
-pub fn log<'a, K>(
+pub fn log<'a, K, L>(
+    logger: &L,
     args: Arguments,
     level: Level,
     target_module_path_and_loc: &(&str, &'static str, &'static Location),
     kvs: K,
 ) where
     K: KVs<'a>,
+    L: Log,
 {
-    log_impl(args, level, target_module_path_and_loc, kvs.into_kvs())
+    log_impl(
+        logger,
+        args,
+        level,
+        target_module_path_and_loc,
+        kvs.into_kvs(),
+    )
 }
 
-pub fn enabled(level: Level, target: &str) -> bool {
-    crate::logger().enabled(&Metadata::builder().level(level).target(target).build())
+pub fn enabled<L: Log>(logger: &L, level: Level, target: &str) -> bool {
+    logger.enabled(&Metadata::builder().level(level).target(target).build())
 }
 
 #[track_caller]
--- a/src/__private_api.rs
+++ b/src/__private_api.rs
@@ -35,7 +35,7 @@ impl<'a> KVs<'a> for () {
 // Log implementation.
 
 fn log_impl<L: Log>(
-    logger: &L,
+    logger: L,
     args: Arguments,
     level: Level,
     &(target, module_path, loc): &(&str, &'static str, &'static Location),
@@ -63,7 +63,7 @@ fn log_impl<L: Log>(
 }
 
 pub fn log<'a, K, L>(
-    logger: &L,
+    logger: L,
     args: Arguments,
     level: Level,
     target_module_path_and_loc: &(&str, &'static str, &'static Location),
@@ -81,7 +81,7 @@ pub fn log<'a, K, L>(
     )
 }
 
-pub fn enabled<L: Log>(logger: &L, level: Level, target: &str) -> bool {
+pub fn enabled<L: Log>(logger: L, level: Level, target: &str) -> bool {
     logger.enabled(&Metadata::builder().level(level).target(target).build())
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1216,28 +1216,6 @@ impl Log for NopLogger {
     fn flush(&self) {}
 }
 
-/// The global logger proxy.
-///
-/// This zero-sized type implements the [`Log`] trait by forwarding calls
-/// to the logger registered with the `set_boxed_logger` or `set_logger`
-/// methods if there is one, or a nop logger as default.
-#[derive(Copy, Clone, Default, Debug)]
-pub struct GlobalLogger;
-
-impl Log for GlobalLogger {
-    fn enabled(&self, metadata: &Metadata) -> bool {
-        logger().enabled(metadata)
-    }
-
-    fn log(&self, record: &Record) {
-        logger().log(record)
-    }
-
-    fn flush(&self) {
-        logger().flush()
-    }
-}
-
 impl<T> Log for &'_ T
 where
     T: ?Sized + Log,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1204,7 +1204,7 @@ pub trait Log: Sync + Send {
     fn flush(&self);
 }
 
-// Just used as a dummy initial value for LOGGER
+/// A dummy initial value for LOGGER.
 struct NopLogger;
 
 impl Log for NopLogger {
@@ -1214,6 +1214,28 @@ impl Log for NopLogger {
 
     fn log(&self, _: &Record) {}
     fn flush(&self) {}
+}
+
+/// The global logger proxy.
+///
+/// This zero-sized type implements the [`Log`] trait by forwarding calls
+/// to the logger registered with the `set_boxed_logger` or `set_logger`
+/// methods if there is one, or a nop logger as default.
+#[derive(Copy, Clone, Default, Debug)]
+pub struct GlobalLogger;
+
+impl Log for GlobalLogger {
+    fn enabled(&self, metadata: &Metadata) -> bool {
+        logger().enabled(metadata)
+    }
+
+    fn log(&self, record: &Record) {
+        logger().log(record)
+    }
+
+    fn flush(&self) {
+        logger().flush()
+    }
 }
 
 impl<T> Log for &'_ T
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -117,10 +117,10 @@ macro_rules! log {
 /// # Examples
 ///
 /// ```
-/// use log::{error, GlobalLogger};
+/// use log::error;
 ///
+/// # let my_logger = log::GlobalLogger;
 /// let (err_info, port) = ("No connection", 22);
-/// let my_logger = GlobalLogger; // can be any logger implementing the Log trait
 ///
 /// error!("Error: {err_info} on port {port}");
 /// error!(target: "app_events", "App Error: {err_info}, Port: {port}");
@@ -156,10 +156,10 @@ macro_rules! error {
 /// # Examples
 ///
 /// ```
-/// use log::{warn, GlobalLogger};
+/// use log::warn;
 ///
+/// # let my_logger = log::GlobalLogger;
 /// let warn_description = "Invalid Input";
-/// let my_logger = GlobalLogger; // can be any logger implementing the Log trait
 ///
 /// warn!("Warning! {warn_description}!");
 /// warn!(target: "input_events", "App received warning: {warn_description}");
@@ -195,11 +195,11 @@ macro_rules! warn {
 /// # Examples
 ///
 /// ```
-/// use log::{info, GlobalLogger};
+/// use log::info;
 ///
+/// # let my_logger = log::GlobalLogger;
 /// # struct Connection { port: u32, speed: f32 }
 /// let conn_info = Connection { port: 40, speed: 3.20 };
-/// let my_logger = GlobalLogger; // can be any logger implementing the Log trait
 ///
 /// info!("Connected to port {} at {} Mb/s", conn_info.port, conn_info.speed);
 /// info!(
@@ -243,11 +243,11 @@ macro_rules! info {
 /// # Examples
 ///
 /// ```
-/// use log::{debug, GlobalLogger};
+/// use log::debug;
 ///
+/// # let my_logger = log::GlobalLogger;
 /// # struct Position { x: f32, y: f32 }
 /// let pos = Position { x: 3.234, y: -1.223 };
-/// let my_logger = GlobalLogger; // can be any logger implementing the Log trait
 ///
 /// debug!("New position: x: {}, y: {}", pos.x, pos.y);
 /// debug!(target: "app_events", "New position: x: {}, y: {}", pos.x, pos.y);
@@ -283,11 +283,11 @@ macro_rules! debug {
 /// # Examples
 ///
 /// ```
-/// use log::{trace, GlobalLogger};
+/// use log::trace;
 ///
+/// # let my_logger = log::GlobalLogger;
 /// # struct Position { x: f32, y: f32 }
 /// let pos = Position { x: 3.234, y: -1.223 };
-/// let my_logger = GlobalLogger; // can be any logger implementing the Log trait
 ///
 /// trace!("Position is: x: {}, y: {}", pos.x, pos.y);
 /// trace!(target: "app_events", "x is {} and y is {}",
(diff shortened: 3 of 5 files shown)


tests/integration.rs
tisonkun, line 1: This is just a good target to test our new `logger` param.

Previously, the global logger state is shared with all the `macros` tests and causes the result unstable.

src/macros.rs
```diff
@@ -29,11 +29,13 @@
...
 macro_rules! log {
-    // log!(target: "my_target", Level::Info, key1:? = 42, key2 = true; "a {} event", "log");
-    (target: $target:expr, $lvl:expr, $($key:tt $(:$capture:tt)? $(= $value:expr)?),+; $($arg:tt)+) => ({
+    // log!(logger: my_logger, target: "my_target", Level::Info, key1:? = 42, key2 = true; "a {} event", "log");
+    (logger: $logger:expr, target: $target:expr, $lvl:expr, $($key:tt $(:$capture:tt)? $(= $value:expr)?),+; $($arg:tt)+) => ({
+        let logger = &$logger;
         let lvl = $lvl;
         if lvl <= $crate::STATIC_MAX_LEVEL && lvl <= $crate::max_level() {
-            $crate::__private_api::log::<&_>(
+            $crate::__private_api::log(
```
tisonkun, line 37: We add a new type parameter and it seems Rust 1.60 requires extra info.

I'm not familiar with this part and would investigate later. Any help is welcome.

tisonkun: Resolved.

Thomasdezeeuw:
I'm not convinced this is a common enough use case that it has to be in facade, couldn't it be implemented in the logging implementation?

tisonkun:
@Thomasdezeeuw 

[spdlog-rs](https://docs.rs/spdlog-rs/0.4.0/spdlog/macro.log.html) implements this, with their own `log!` macro. When adapting to the `log` crate's macros, the functionality can't be used.

So, the logging implementation must implement the same functionality by bypassing our macro, which loses the value of a facade. If you have an idea how to implement and use it with our abstraction and a custom impl crate, I'm glad to know.

Thomasdezeeuw:
> So, the logging implementation must implement the same functionality by bypassing our macro, which loses the value of a facade. If you have an idea how to implement and use it with our abstraction and a custom impl crate, I'm glad to know.

I my crate I used the `target` to change the target of the logs. In this it uses a constant for the `target` to switch from logging to standard error to standard out. See https://docs.rs/std-logger/latest/std_logger/macro.request.html for an example, but any macro can be used by setting the `target` to a constant (https://docs.rs/std-logger/latest/std_logger/constant.REQUEST_TARGET.html)

tisonkun:
Thanks for your information! I'm considering the difference between target and a logger field also.

In the description:

> I noticed that we previously introduced `target` for similar usage, but switching to a different logger is different.

Switching targets still use the same (global & singleton) logger instance, while setting a new logger instance can avoid meta logging cycle - https://github.com/rust-lang/log/issues/645.

Thomasdezeeuw:
> Switching targets still use the same (global & singleton) logger instance, while setting a new logger instance can avoid meta logging cycle - #645.

The global logging implementation can be set up to call different loggers depending on `target`.

KodrAus:
Personally, I think being able to target a different logger entirely using an argument is useful in a few scenarios:

1. You want to log from your logging pipeline. If your pipeline is complex enough, you sometimes end up needing to do this.
2. You don't want to use a global logger at all, but still want some of the nice ergonomics of the macros. You may do this, for example, to implement your own simplified contextual logging.

From a maintenance perspective it also keeps us a bit honest in not being able to assume that everything is shared by a single global logger.

If @Thomasdezeeuw isn't convinced it's worth adding though then we'll leave it for the time being.

Thomasdezeeuw:
I'm not convinced the additional complexity is worth the use case that is already possible via the log implementation.

@tisonkun do you have any argument why it should be done this way, and not via the log implementation?

NobodyXu:
This PR reminds me of [`with_default`](https://docs.rs/tracing/latest/tracing/subscriber/fn.with_default.html):

> Sets this Subscriber as the default for the current thread for the duration of a closure.
> 
> The default subscriber is used when creating a new Span or Event.

Sounds like you actually want something like this?

@tisonkun

tisonkun:
> Sounds like you actually want something like this?

@NobodyXu No. This is a thread-local default, which can be even worse than a global default to tune with. The `logger` field in `log!` macro is an explicit instance, `with_default` is a more subtle implict.

SpriteOvO (commented):
Hi, I'm the author of `spdlog-rs`, and I'd like to thank @tisonkun for mentioning my crate and nominating to merge this feature into `log` crate, and @KodrAus for supporting this feature in the issue. I'll give my two cents here.

The introduction of the `logger` parameter (and replaceable global logger) in `spdlog-rs` (as well as the same concept in C++ `spdlog`) was primarily to address the inconvenience of singleton - can only be set once at startup.

Imagine you have a huge project with many components. Some components just print to standard streams, some components need to report logs to a specific network endpoint (maybe even this endpoint needs to be dynamically obtained and unknown at startup). In the singleton model, the initial setup needs to prepare all components for their needs, rather than being managed by the components themselves. And since the discriminator `target` is just a string, ambiguities may be uncommon but are by no means absent especially when match rules like `starts_with` are involved.

The addition of the `logger` parameter is not killing the singleton pattern, which is still the easiest way for lightweight projects, but larger projects can benefit from this harmless addition.

src/__private_api.rs
```diff
@@ -34,7 +34,8 @@ impl<'a> KVs<'a> for () {
 
 // Log implementation.
 
-fn log_impl(
+fn log_impl<L: Log>(
+    logger: &L,
```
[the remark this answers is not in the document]
KodrAus: Having the generic here is actually a nice way to avoid dynamic dispatch by using an explicit `logger` in your `log!` calls.

tisonkun:
Merge master branch and resolve conflicts.

May we have another round to review this patch?

tests/integration.rs
```diff
@@ -53,31 +53,31 @@ fn main() {
...
+        test_line_numbers(&logger, &a);
     }
 }
 
-fn test_filter(a: &State, filter: LevelFilter) {
+fn test_filter(logger: &dyn Log, a: &State, filter: LevelFilter) {
     // tests to ensure logs with a level beneath 'max_level' are filtered out
     log::set_max_level(filter);
-    error!("");
+    error!(logger: logger, "");
```
KodrAus, line 72: I think we'd need to document this behavior. We have a hierarchy of level controls; at compile time, at runtime, and whatever the logger itself applies. We should be clear that these are features of the macros, so apply equally regardless of the logger you plug in.

KodrAus (approved):
This looks good to me @tisonkun :+1:

@Thomasdezeeuw are you happy if we move forwards with this feature? I think pushing back on additions is a good position that keeps us from accumulating bloat, but feel like this one specifically is a reasonable enhancement aligned with the existing macro capabilities.

Thomasdezeeuw:
> @Thomasdezeeuw are you happy if we move forwards with this feature? I think pushing back on additions is a good position that keeps us from accumulating bloat, but feel like this one specifically is a reasonable enhancement aligned with the existing macro capabilities.

I still think the use case can be solved without this change, hence I don't think it's worth the additional complexity to be honest.

However, if we go forward with it have we looked at the code size changes? Work has been done in the past to reduce the compiler code size, is that affected at all? /cc @EFanZh in case you're still interested in this

KodrAus:
> I still think the use case can be solved without this change

Some of them can, but some can be done more easily without needing to write dispatching logic based on targets. Being able to plug a logger in to the macro calls would let you use local loggers for contextual logging, without needing a centrally synchronized place to store them. That could also potentially be solved with the key-value support though if we let you plug in an additional `Source` alongside specific key-values.

Maybe we can think of it more like `print!` vs `write!`. The former doesn't let you specify where the output goes, whereas the latter does. If we want to decide that `log!` is just `print!` with a configurable output via `set_boxed_logger` then this change isn't something we'd pursue.  If we want the `log!` macros to be a general interface to any `Log` impl, then I think it would be very useful.

src/__private_api.rs
```diff
@@ -58,22 +59,30 @@ fn log_impl(
     #[cfg(feature = "kv")]
     builder.key_values(&kvs);
 
-    crate::logger().log(&builder.build());
+    logger.log(&builder.build());
 }
 
-pub fn log<'a, K>(
+pub fn log<'a, K, L>(
+    logger: &L,
```
EFanZh, line 66: For binary size, I think the type of logger should be `L` instead of `&L`, so that it allows us to pass zero-sized logger. A default logger could be defined as as a zero-sized type which forwards logging calls to `create::logger()`, just like what the allocator [`Global`](https://doc.rust-lang.org/stable/std/alloc/struct.Global.html) does.

EFanZh (changes requested):

src/__private_api.rs
tisonkun: Sounds reasonable. We have:

```rust
impl<T> Log for &'_ T { ... }
```

Let me send an updater later today.

Thomasdezeeuw (approved):
I won't block this (any further) if code size isn't a concern.

I do this we should add some more documentation for this change as I don't see any. I would at least expect the `logger` argument to be documented and stating that by default it will use the global logger.

tisonkun:
> I do this we should add some more documentation for this change as I don't see any. I would at least expect the logger argument to be documented and stating that by default it will use the global logger.

Thanks for your suggestions. Then please leave me a few days so that I can pick up this patch :D

src/__private_api.rs
tisonkun: @EFanZh This is resolved at https://github.com/rust-lang/log/pull/664/commits/cde258593221716c7b64c049d6619b49b30dd4de now. Please check if it's what you expect.

src/macros.rs
```diff
@@ -56,8 +60,20 @@ macro_rules! log {
...
     });
 
+    // log!(logger: my_logger, Level::Info, "a log event")
+    (logger: $logger:expr, $lvl:expr, $($arg:tt)+) => ({
+        $crate::log!(logger: $logger, target: $crate::__private_api::module_path!(), $lvl, $($arg)+)
+    });
+
+    // log!(target: "my_target", Level::Info, "a log event")
+    (target: $target:expr, $lvl:expr, $($arg:tt)+) => ({
+        $crate::log!(logger: $crate::logger(), target: $target, $lvl, $($arg)+)
```
EFanZh, line 70: The default logger should be a zero sized type, otherwise caller needs pay the cost of passing the logger value.

tests/integration.rs
```diff
@@ -53,31 +53,31 @@ fn main() {
             last_log_location: Mutex::new(None),
         });
         let a = me.clone();
-        set_boxed_logger(Box::new(Logger(me))).unwrap();
+        let logger = Logger(me);
```
SpriteOvO, line 56: What happens if we call a log macro with `logger` param here? The macro would expand a `let logger = $logger;`, which seems to move rather than borrow the logger variable?

tisonkun: IIUC it would be `debug!(logger: &logger, ...)` then. That is, use ref from the user size.

One of the rationales is @EFanZh's comment about "binary size", see:

* https://github.com/rust-lang/log/pull/664#discussion_r1972703604
* https://github.com/rust-lang/log/pull/664#discussion_r1973798743

src/macros.rs
tisonkun: Please check if this is what you mean https://github.com/rust-lang/log/pull/664/commits/2bb47d65a5c6f94bdc1d694bec4824431201e31a

src/lib.rs
```diff
@@ -1216,6 +1216,28 @@ impl Log for NopLogger {
...
+    }
+
+    fn log(&self, record: &Record) {
+        logger().log(record)
+    }
+
+    fn flush(&self) {
+        logger().flush()
+    }
+}
```
tisonkun, line 1239: Shall we `#[inline]` here? In my application development conventions, I trust the compiler to do the inlining based on its analysis.

tests/integration.rs
tisonkun: Could you please elaborate a bit what to be documented and where I should add the docs?

It seems that you'd like to emphasize that these macros can be compiled to nop if the related flags (e.g., `max_level_error`, `release_max_level_error`, etc.) enabled.

tests/macros.rs
```diff
@@ -10,6 +10,16 @@ macro_rules! all_log_macros {
...
 
+struct NopLogger;
+
+impl Log for NopLogger {
+    fn enabled(&self, _: &Metadata) -> bool {
+        false
+    }
+    fn log(&self, _: &Record) {}
+    fn flush(&self) {}
+}
```
tisonkun, line 21: This inspires me to consider whether we should export the NopLogger ..

EFanZh:
Another thought: with the current implementation, user specified logger will always be passed by reference, which might not be ideal for binary size. Do we want to support the case where user can specify a zero-sized logger to reduce binary size?

tisonkun:
> Another thought: with the current implementation, user specified logger will always be passed by reference

What does it mean? We already accept `L` that can be an owned logger. Could you give a concrete example where a performance concern can occur.

Does https://github.com/rust-lang/log/commit/2bb47d65a5c6f94bdc1d694bec4824431201e31a implements what you mean for passing a zero-sized logger? If not, why.

EFanZh:
@tisonkun: Oh, I thought the implementation still passes reference for user specified loggers. Now the implementation passes logger argument as is, then I have no further questions.

tisonkun:
@EFanZh OK. Good to know the clarification.

Then just leave me some time to figure out how to improve the docs :D

src/macros.rs
```diff
@@ -30,11 +30,13 @@
 #[macro_export]
 #[clippy::format_args]
 macro_rules! log {
-    // log!(target: "my_target", Level::Info, key1:? = 42, key2 = true; "a {} event", "log");
-    (target: $target:expr, $lvl:expr, $($key:tt $(:$capture:tt)? $(= $value:expr)?),+; $($arg:tt)+) => ({
+    // log!(logger: my_logger, target: "my_target", Level::Info, key1:? = 42, key2 = true; "a {} event", "log");
+    (logger: $logger:expr, target: $target:expr, $lvl:expr, $($key:tt $(:$capture:tt)? $(= $value:expr)?),+; $($arg:tt)+) => ({
+        let logger = $logger;
```
EFanZh, line 35: This has the semantic of executing `$logger` unconditionally, do we want to move this inside the `if` statement below to make its execution conditionally like other arguments?

tests/integration.rs
SpriteOvO: I see. If explicitly passing ref from the user side is intentional, then I'm fine with that.

tisonkun:
@Thomasdezeeuw @KodrAus Added some docs. Welcome to give a review and see how we can improve.

src/macros.rs
```diff
@@ -43,11 +81,13 @@ macro_rules! log {
         }
     });
 
-    // log!(target: "my_target", Level::Info, "a {} event", "log");
-    (target: $target:expr, $lvl:expr, $($arg:tt)+) => ({
+    // log!(logger: my_logger, target: "my_target", Level::Info, "a {} event", "log");
+    (logger: $logger:expr, target: $target:expr, $lvl:expr, $($arg:tt)+) => ({
         let lvl = $lvl;
         if lvl <= $crate::STATIC_MAX_LEVEL && lvl <= $crate::max_level() {
+            let logger = $logger;
```
EFanZh, line 88: Also, maybe move `$logger` to the argument position? So that it can support temporary expressions such as `std::convert::identity(&MyLogger::new())`;

src/lib.rs
```diff
@@ -1216,6 +1216,28 @@ impl Log for NopLogger {
     fn flush(&self) {}
 }
 
+/// The global logger proxy.
+///
+/// This zero-sized type implements the [`Log`] trait by forwarding calls
+/// to the logger registered with the `set_boxed_logger` or `set_logger`
+/// methods if there is one, or a nop logger as default.
+#[derive(Copy, Clone, Default, Debug)]
+pub struct GlobalLogger;
```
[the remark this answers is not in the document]
KodrAus: I think we should keep this private, at least for now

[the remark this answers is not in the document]
tisonkun: OK. Then we may need to `no_run` the docs? Or it may be too wordy to define a dummy logger for each doc segments.

src/macros.rs
tisonkun: Let me give it a try ..

src/lib.rs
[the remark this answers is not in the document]
tisonkun: We use this logger in the macro so we may expose it anyway. Do you mean move it to `__private_api.rs`?

KodrAus (commented):
Thanks for working through this @tisonkun! There's been a lot of back-and-forth and you've done a lot to accommodate the various requirements of other users. I've just left a few comments.

src/lib.rs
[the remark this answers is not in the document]
tisonkun: For why do we have this `GlobalLogger`, see https://github.com/rust-lang/log/pull/664#discussion_r1973798743.

src/macros.rs
```diff
@@ -13,28 +13,66 @@
...
+///     "Received errors: {}, {}",
+///     data.0, data.1
+/// );
 #[macro_export]
 #[clippy::format_args]
 macro_rules! log {
-    // log!(target: "my_target", Level::Info, key1:? = 42, key2 = true; "a {} event", "log");
-    (target: $target:expr, $lvl:expr, $($key:tt $(:$capture:tt)? $(= $value:expr)?),+; $($arg:tt)+) => ({
+    // log!(logger: my_logger, target: "my_target", Level::Info, key1:? = 42, key2 = true; "a {} event", "log");
+    (logger: $logger:expr, target: $target:expr, $lvl:expr, $($key:tt $(:$capture:tt)? $(= $value:expr)?),+; $($arg:tt)+) => ({
```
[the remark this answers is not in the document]
tisonkun: Thanks for your advice. Let me give it another consideration ..

src/lib.rs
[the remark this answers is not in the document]
KodrAus: I think we'd move it to `__private_api`. It would be an issue to try and actually use it for anything except in the `logger` argument. Attempting to install it via `set_boxed_logger` would be a bad time, for instance.

In the docs, we can at least hide the sections that define any custom loggers.

tisonkun:
@KodrAus Updated.

> In order to get the most benefit from a zero-sized logger we'd need to do some refactoring of the __private_api::log function, to invoke the log function at the callsite, or figure something else out, but we can do that in a follow-up PR.

Yeah. I think this is how `write!` avoid passing reference for zero-size instance. I tried it and find the challenge mainly comes from that `RecordBuilder<'a>` has a lifetime and perhaps borrow checker would complain. Anyway, I'll try to figure it out in a follow-up PR.

tests/integration.rs
EFanZh: @KodrAus: How about keeping `__private_api::log` taking logger argument by value, but the logger macros passes references for custom loggers, and passes global logger by value? In this way, the global logger can still utilize the ZST optimization, although not the custom ones.

KodrAus: @EFanZh That sounds good to me :+1: We're currently halfway there; the `__private_api::log` function takes the logger by value, but the way we expand `log!` macros always takes by reference. I'll follow up with that change to make an unspecified logger accept `GlobalLogger` by value. That's also a good way for me to get fully across the new code.

KodrAus (approved):
Thanks for working through this @tisonkun! This looks good to me.

PR #665, opened 2025-02-06
Enable clippy support for format-like macros

nyurik:
Add `#[clippy::format_args]` as described in [Clippy book](https://doc.rust-lang.org/nightly/clippy/attribs.html#clippyformat_args)

This is a new feature in v1.85

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -28,6 +28,7 @@
 /// # }
 /// ```
 #[macro_export]
+#[clippy::format_args]
 macro_rules! log {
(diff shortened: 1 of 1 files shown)


Thomasdezeeuw:
How does this work with older Clippy versions? Is this just ignored?

nyurik:
Clippy would work fine on anything that uses log. Older clippy would only fail if you try to run it on the log crate itself.

KodrAus (approved):
Thanks @nyurik. This looks like it should be compatible with the `log!` macro syntax.

PR #668, opened 2025-02-14
Derive `Clone` for `kv::Value`

### LINKED ISSUE #667: Derive `Clone` for `kv::Value`?
Currently the only way to clone a `kv::Value` seems to be `ToValue::to_value()`, it limits the downstream `kv::Value` wrappers to implement `Clone`.

For example:

```rust
// #[derive(Clone)]
struct Wrapper<'a>(log::kv::Value<'a>);

impl Clone for Wrapper<'_> {
    fn clone(&self) -> Self {
        Self(self.0.to_value())
    }
}
```

First, the wrapper can't directly `#[derive(Clone)]` as `kv::Value` does not implement `Clone`.

Second, the `to_value()` call raises a lifetime error:

```
lifetime may not live long enough
method was supposed to return data with lifetime `'2` but it is returning data with lifetime `'1`
```

I think adding a `#[derive(Clone)]` to `kv::Value` should be fine and it solves the problem.

SpriteOvO:
Closes #667.

### DIFF
--- a/src/kv/value.rs
+++ b/src/kv/value.rs
@@ -115,6 +115,7 @@ impl<'v> ToValue for Value<'v> {
 /// `Display` implementation, it will serialize through `serde` as a string. If it was
 /// captured as a struct using `serde`, it will also serialize as a struct
 /// through `sval`, or can be formatted using a `Debug`-compatible representation.
+#[derive(Clone)]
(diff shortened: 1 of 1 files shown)


KodrAus (approved):
Thanks @SpriteOvO! This looks good to me.

PR #673, opened 2025-03-23
Pass global logger by value, supplied logger by ref

KodrAus:
Follow-up to #664 

cc @EFanZh @tisonkun @Thomasdezeeuw 

This PR ensures we pass loggers by-ref when they're user-supplied, and by-value when it's the built-in global logger. I'll add a few tests to ensure we don't take ownership of the logger in `log!` or `log_enabled!`.

### DIFF
--- a/src/__private_api.rs
+++ b/src/__private_api.rs
@@ -85,7 +85,7 @@ fn log_impl<L: Log>(
 }
 
 pub fn log<'a, K, L>(
-    logger: &L,
+    logger: L,
     args: Arguments,
     level: Level,
     target_module_path_and_loc: &(&str, &'static str, &'static Location),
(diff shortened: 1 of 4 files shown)


tisonkun (approved):
This looks good.

While I wonder if we can somehow inline the `log_impl` in the caller so that all the zero-size logger can benefit? Just like `write!`'s

```rust
macro_rules! write {
    ($dst:expr, $($arg:tt)*) => {
        $dst.write_fmt($crate::format_args!($($arg)*)) // <- directly call on the pass-in obj
    };
}
```

KodrAus:
Yeh I think we could basically make `__private_api::log` into a trait method that we’d implement for `L: Log` and call that at the callsite.

Thomasdezeeuw (approved):

KodrAus:
I've added some more tests around the `logger` argument, just to make sure we're more likely to catch regressions in the future.

PR #675, opened 2025-03-23
Try to use a trait for calling log impl

tisonkun:
This follows up https://github.com/rust-lang/log/pull/673#pullrequestreview-2708476028.

cc @EFanZh not sure if it creates some overhead or how can I check it.

### DIFF
--- a/src/__private_api.rs
+++ b/src/__private_api.rs
@@ -52,57 +52,43 @@ impl Log for GlobalLogger {
     }
 }
 
-// Split from `log` to reduce generics and code size
-fn log_impl<L: Log>(
-    logger: L,
-    args: Arguments,
-    level: Level,
-    &(target, module_path, loc): &(&str, &'static str, &'static Location),
-    kvs: Option<&[(&str, Value)]>,
-) {
-    #[cfg(not(feature = "kv"))]
-    if kvs.is_some() {
-        panic!("key-value support is experimental and must be enabled using the `kv` feature")
+pub trait LogExt: Log {
+    fn do_enabled(&self, level: Level, target: &str) -> bool {
+        self.enabled(&Metadata::builder().level(level).target(target).build())
     }
 
-    let mut builder = Record::builder();
+    fn do_log<'a>(
+        &self,
+        args: Arguments,
+        level: Level,
+        &(target, module_path, loc): &(&str, &'static str, &'static Location),
+        kvs: impl KVs<'a>,
+    ) {
+        let kvs = kvs.into_kvs();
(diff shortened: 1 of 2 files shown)


src/__private_api.rs
```diff
@@ -52,57 +52,43 @@ impl Log for GlobalLogger {
...
-    L: Log,
-{
-    log_impl(
-        logger,
-        args,
-        level,
-        target_module_path_and_loc,
-        kvs.into_kvs(),
-    )
+    fn do_log<'a>(
```
KodrAus, line 60: If we’re going to do it this way these names should probably be much more mangled, like `__log_private_api_log`. Similarly, the name of the trait should also be ugly so it doesn’t look like something you’re meant to use directly.

tisonkun: Agree. I'd first check whether this way work (so that all the zero-size logger can benefit). And rename later.

src/macros.rs
```diff
@@ -135,8 +136,9 @@ macro_rules! __log {
     (logger: $logger:expr, target: $target:expr, $lvl:expr, $($arg:tt)+) => ({
         let lvl = $lvl;
         if lvl <= $crate::STATIC_MAX_LEVEL && lvl <= $crate::max_level() {
-            $crate::__private_api::log(
-                $logger,
+            use $crate::__private_api::LogExt;
+
+            $logger.do_log(
```
EFanZh, line 141: You may need to refer to the trait directly: `LogExt::do_log(...)`, otherwise if the logger has a `do_log` method that is not a trait method, it will be called here.

src/__private_api.rs
```diff
@@ -52,57 +52,43 @@ impl Log for GlobalLogger {
...
+        &self,
```
EFanZh, line 61: This forces caller to pass a non-ZST reference, which causes the ZST optimization to lose effect.

tisonkun: I can see that the final `fn log(&self, ..)` and `fn enabled(&self, ..)` method take a reference also.

So the optimization happens only in the caller-side, as you described in https://github.com/rust-lang/log/pull/576#discussion_r2009369123?

EFanZh: Yes. You need to make sure that the arguments passed to the outermost function are ZSTs.

tisonkun: OK. Then I'd close this PR as not work.

Although, I wonder if `write!` use `$dst.write_fmt` that accepts a reference already, is this optimization significant? Or a net win?

PR #676, opened 2025-03-24
ci: drop really old trick and use new actions 

tisonkun:
@KodrAus I found that our MSRV is not quite 1.60.0 since we rely on value-bag 1.7+ who needs 1.61 for const fn. This is also the MSRV for many other dependencies newest version (serde_derive, memchr, syn, etc.)

Also I found https://github.com/sval-rs/sval/pull/207 breaks MSRV 1.60 or 1.61.

Not sure if our MSRV for basic usage or any feature combo. I'd suggest to bump the MSRV to 1.61 and try to ensure that sval won't break the same baseline.


tests/src/build.rs
tisonkun, line 1: yak shaving from here .. even from rustc 1.60.0 we don't need this trick I think.

Cargo.toml
```diff
@@ -66,16 +66,22 @@ value-bag = { version = "1.7", optional = true, default-features = false, featur
...
 # https://github.com/dtolnay/proc-macro2/commit/e31d61910049e097afdd3d27c37786309082bdcb.
 # By defining the crate as direct dependency we can increase its minimal
 # version making the minimal (crate) version CI happy.
 proc-macro2 = { version = "1.0.63", default-features = false }
+
+[patch.crates-io]
+sval = { git = "https://github.com/tisonkun/sval.git", rev = "06aa86b" }
+sval_buffer = { git = "https://github.com/tisonkun/sval.git", rev = "06aa86b" }
+sval_nested = { git = "https://github.com/tisonkun/sval.git", rev = "06aa86b" }
+sval_ref = { git = "https://github.com/tisonkun/sval.git", rev = "06aa86b" }
```
tisonkun, line 87: for https://github.com/sval-rs/sval/pull/207

.github/workflows/main.yml
```diff
@@ -121,26 +94,24 @@ jobs:
 
   msrv:
     name: MSRV
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
```
tisonkun, line 97: Pin the version to avoid unexpected pointer changes.

.github/workflows/main.yml
```diff
@@ -1,101 +1,77 @@
 name: CI
 on: [push, pull_request]
 
-permissions:
-  contents: read #  to fetch code (actions/checkout)
```
tisonkun, line 5: Default to have

.github/workflows/main.yml
```diff
@@ -121,26 +94,24 @@ jobs:
...
+    runs-on: ubuntu-24.04
     steps:
-      - uses: actions/checkout@master
-      - name: Install Rust
-        run: |
-          rustup update 1.60.0 --no-self-update
-          rustup default 1.60.0
-      - run: |
-          cargo test --verbose --manifest-path tests/Cargo.toml
-          cargo test --verbose --manifest-path tests/Cargo.toml --features kv
```
tisonkun, line 133: Previously we check MSRV only for a subset.

.github/workflows/main.yml
KodrAus: What kinds of unexpected pointer changes are you meaning here?

.github/workflows/main.yml
KodrAus: Previously we only applied it to a subset, but now that the key-value API is stable I’m happy to add one for it. What do you think @Thomasdezeeuw?

.github/workflows/main.yml
tisonkun: I mean GitHub can change what `ubuntu-latest` means, perhaps "ubuntu 26.04" one day.

Previously, some of projects I participants find their CI failed due to this unexpected bump, while using a pinned version give us the freedom to bump on demand and more manageable (rather than affect all the new PRs).

.github/workflows/main.yml
tisonkun: Thanks for your information! I'd emphasize that ensuring MSRV for sval/serde would require bumping MSRV to 1.61, mainly for const fn. You may take it into consideration whether it's possible, since I see:

> This version is explicitly tested in CI and may be bumped in any release as needed. Maintaining compatibility with older compilers is a priority though, so the bar for bumping the minimum supported version is set very high. Any changes to the supported minimum version will be called out in the release notes.

.github/workflows/main.yml
KodrAus: This seems like something we'll simply not do if it doesn't stop working, so I'd rather leave it and see if does end up breaking.

Cargo.toml
KodrAus: I've released `1.14.1` of these crates now so we should be able to remove this now.

.github/workflows/main.yml
```diff
@@ -1,101 +1,77 @@
...
-      run: |
-        rustup update ${{ matrix.rust }} --no-self-update
-        rustup default ${{ matrix.rust }}
-        cargo install cargo-hack
+    - uses: actions/checkout@v4
+    - name: Install toolchain
+      uses: dtolnay/rust-toolchain@master
+      with:
+        toolchain: ${{ matrix.rust }}
+    - uses: taiki-e/install-action@cargo-hack
```
KodrAus, line 38: I'd rather not depend on any external actions in CI, and instead run the scripts ourselves. Doing that makes our CI less opaque, reduces our dependence on outside infrastructure, and limits our tie in to GitHub Actions itself.

tests/src/build.rs
KodrAus: Glad to see this get cleaned up! I think we were previously only testing these in our MSRV job. To get them working this way we may need to pull that `macros` module into the regular `log` crate so its tests are still picked up.

.github/workflows/main.yml
Thomasdezeeuw: I think this was about limiting the write permissions.

.github/workflows/main.yml
Thomasdezeeuw: Technically a MSRV bump is a breaking change, but since 1.61 came out in May 2022, I think it's fine we increase it.

Thomasdezeeuw (commented):
For the future it would be nice if these can be different prs. This does so many things it much harder to review.

.github/workflows/main.yml
tisonkun: OK. Let me restore this.

tests/src/build.rs
tisonkun: Do you provide any change here?

Cargo would detect `tests` folder and these two files are already tested by the after setup.

See https://doc.rust-lang.org/book/ch11-03-test-organization.html#integration-tests

Cargo.toml
tisonkun: Updated.

.github/workflows/main.yml
BurntSushi: I don't know why you say it is "technically" a breaking change. The `log` policy in the README even explicitly allows it, and that is the general practice of Rust libraries (although some do treat MSRV bumps as semver incompatible).

See also: https://doc.rust-lang.org/cargo/reference/rust-version.html#setting-and-updating-rust-version

BurntSushi: The topic has been discussed to death. Search the libs team repo or the api guidelines repo. I'm on mobile or I would link you. Increasing toolchain requirements is broadly not considered a breaking change.

KodrAus:
Thanks @tisonkun :+1: The only issue left I think is that we're no longer running our macro tests in CI. Otherwise this looks good to me.

tisonkun:
@KodrAus Aha. That's because we pass `--lib` for cargo-hack tests. Fixed in 50c2018d50517bf8ec9aafae12b3a6993c29213f

KodrAus (approved):
Thanks @tisonkun! This is a nice round of cleanups. It’s great to get those old tests into better shape

tisonkun:
Thanks for your review @KodrAus @Thomasdezeeuw!

Some out of thread comments:

@KodrAus Yes I always find that once the code structure becomes more straightforward (self-described and well structured), it can foster more contributions. Otherwise, to work around an old workaround you add some new tricks and quickly the codebase becomes misty and even unmaintainable. I like "obviously no bugs" over "no obvious bugs".

PR #687, opened 2025-06-14
Is there any reason for some of these `struct`s not to implement `Copy`?

AhoyISki:
I saw that some of the structs that are part of the `Record` didn't implement the `Copy` trait. Is there any reason for this? If there is, I guess you can ignore this pull request, but in my use case, it would be kind of useful to be able to copy a `Record`, especially considering that nothing inside of a `Record` precludes it from implementing `Copy`, since it's all just references inside.


KodrAus:
Thanks for the PR @AhoyISki. The main reason we haven't done this already would just be because it's a major API commitment that we wouldn't ever be able to back out of, and since the `Log` trait takes records by reference, there isn't a need in `log`'s own API to copy records.

AhoyISki:
That's understandable, I guess.

Issue #689, opened 2025-07-17
Confusing [cfg]s


HaoliangXu:
In lib.rs file
[https://github.com/rust-lang/log/blob/master/src/lib.rs#L421](url)
I find this snippet:
```
#[cfg(not(target_has_atomic = "ptr"))]
impl AtomicUsize {
    const fn new(v: usize) -> AtomicUsize {
        AtomicUsize { v: Cell::new(v) }
    }

    fn load(&self, _order: Ordering) -> usize {
        self.v.get()
    }

    fn store(&self, val: usize, _order: Ordering) {
        self.v.set(val)
    }

    #[cfg(target_has_atomic = "ptr")]
    fn compare_exchange(
        &self,
        current: usize,
        new: usize,
        _success: Ordering,
        _failure: Ordering,
    ) -> Result<usize, usize> {
        let prev = self.v.get();
        if current == prev {
            self.v.set(new);
        }
        Ok(prev)
    }
}
```

the thing confusing me is that the first cfg takes effect when target_has_atomic = "ptr" is false, then in the impl scope, compare_exchange function will never be compiled because it needs target_has_atomic="ptr" be true.
Am I right?

Thomasdezeeuw:
Seems like the method is not used? I think we can remove it. Good find 👍

HaoliangXu:
Thank you. I submitted a pull request and got merged, so I'm closing it.

PR #690, opened 2025-07-18
Chore: delete compare_exchange method for AtomicUsize on platforms without atomics 

HaoliangXu:
Since it's never been compiled or used.
Related to #689


KodrAus (approved):
Thanks @HaoliangXu!

Issue #460, opened 2021-08-27
Allow creation of `Level` and `LevelFilter` from `usize`


WaDelma:
Currently `Level::from_usize` and `LevelFilter::from_usize` are private and cannot be used outside the crate.

I wanted to allow providing number via command line flag and use that to construct `LevelFilter`, but to do that I would need to replicate the code of those methods.

KodrAus:
Hi @WaDelma :wave:

I don't think we want to commit to an integer representation for these and the best approach would be for you to pick one for your use case and map them to `log`'s. In my experience, integer representations, even well-understood ones like syslog, are an issue for interoperability because you've got no wiggle room for interpretation. You're just as likely to see `1` interpreted as _error_ or as _information_.

phil-opp:
> I don't think we want to commit to an integer representation for these

Given that the types are `repr(usize)`, the crate already commited to a specific integer representation in its public API. For example, the following prints `3`:

```rust
fn main() {
    dbg!(log::LevelFilter::Info as usize);
}
```

([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=6f5ff65d715f7dca66b7da01fc1cc3eb))

So I think it makes sense to expose the associated `from_usize` methods.

PR #691, opened 2025-07-30
add `TryFrom<usize>` impl for Level and LevelFilter

nebkor:
This exposes the private `Level::from_usize()` and `LevelFilter::from_usize()` functions as fallible conversions. My motivation is to avoid having to use `std::mem::transmute()`, which is unsafe, when I want to convert a value I got doing `my_log_level as usize` back to a `LevelFilter`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -460,6 +460,9 @@ static SET_LOGGER_ERROR: &str = "attempted to set a logger after the logging sys
 static LEVEL_PARSE_ERROR: &str =
     "attempted to convert a string that doesn't match an existing log level";
 
+static LEVEL_USIZE_ERROR: &str = "attempted to convert a usize that was out of valid range for \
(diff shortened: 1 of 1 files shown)


nebkor:
My specific case was to use it in this function:

```rust
fn change_log_level(delta: isize) {
    let level = log::max_level() as isize;
    let max = log::LevelFilter::max() as isize;
    let level = (level + delta).clamp(0, max) as usize;
    // levelfilter is repr(usize) so this is safe
    let level = unsafe { std::mem::transmute::<usize, LevelFilter>(level) };

    log::set_max_level(level);
}
```

and I was surprised I had to reach for `transmute` (or otherwise do something clunky like match on each level).

Thomasdezeeuw:
I don't think we want to commit to this, see https://github.com/rust-lang/log/issues/460.

PR #697, opened 2025-09-08
fixes docs.rs anchors in README.md

gabrielfalcao:
This fixes several broken anchors to the documentation of other rust crates related to the `log` one.


KodrAus:
Thanks for the PR @gabrielfalcao! These anchors as written still work, but it looks like the links for `log4rs` and `spdlog` have actually been broken by this change.

To avoid churning the readme I'll go ahead and close this one, but thanks for looking into it.

PR #698, opened 2025-09-19
Fix atomicity violation in set_logger_racy using compare_exchange

quantpoet:
This commit fixes a race condition in `set_logger_racy` where multiple threads could simultaneously initialize different loggers. 

The fix uses `atomic compare_exchange` to ensure only one thread can transition from UNINITIALIZED to INITIALIZING state, eliminating the TOCTOU bug while preserving the original API.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -431,6 +431,22 @@ impl AtomicUsize {
     fn store(&self, val: usize, _order: Ordering) {
         self.v.set(val)
     }
+
+    fn compare_exchange(
+        &self,
+        current: usize,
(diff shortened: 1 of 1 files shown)


quantpoet:
> Thanks for the PR @quantpoet.
> 
> `set_logger_racy` is racy by definition, and is intended for embedded environments that don’t have multiple threads. As a caller you’re required to ensure there is only a single call to the function. Since the compare exchange here isn’t actually atomic I don’t think it really buys us any additional safety.

Oh indeed. Thanks for your reply!

PR #703, opened 2025-09-30
docs: Add missing impls from README.md

AldaronLau:
I noticed two of the implementations called out in the README are missing from the almost identical list in the root docs.  This adds them there as well.


KodrAus (approved):
Thanks @AldaronLau!

Issue #706, opened 2025-10-18
non-primitive cast error when using multiple key-value pairs in debug! macro inside async_trait


xhermosilla:
## Description

When using the debug! (or other log-level) macro with multiple key-value pairs, compilation fails with the following error:

```rust
error[E0605]: non-primitive cast: `&[(&'static str, Value<'_>); 2]` as `&[(&str, Value<'_>)]`
```

However, when only one key-value pair is used, or when no key-values are included, it compiles and runs correctly.

### Example

This compiles fine:

```rust
debug!(host = "localhost", "Deleting documents");
```

But adding an additional key-value pair causes a compilation error:

```rust
debug!(host = "localhost", version = "1.0.0"; "Deleting documents");
```

This produces:

```rust
error[E0605]: non-primitive cast: `&[(&'static str, Value<'_>); 2]` as `&[(&str, Value<'_>)]`
```

### Context

This happens inside a function annotated with #[async_trait], but the same pattern works fine outside of async traits.
It seems related to how Rust infers lifetimes and array types in macro expansions involving slices of tuples.

The generated code from the macro expands roughly into:

```rust
&[("host", host_value), ("version", version_value)] as &[(&str, Value)]
```

which causes the compiler to reject the cast because the inferred array type has 'static keys (&'static str) but the target expects &str.

### Workaround

Replacing the slice construction inside the macro with this expression fixes the issue:

```rust
&vec![
    ($crate::__log_key!($key), $crate::__log_value!($key $(:$capture)* = $($value)*))
] as &[_],
```

Using a vec![] and taking a reference seems to allow proper type coercion to &[(&str, Value)].

### Possible Cause

This likely comes from a type coercion mismatch between:
	•	the fixed array literal ([ ... ]) which has a concrete type like [(&'static str, Value<'_>); N],
	•	and the expected trait-bound type &[(&str, Value<'_>)].

The compiler disallows this cast because arrays of tuples with different lifetimes on the first element are not directly coercible to slices of tuples with more general lifetimes.

Using vec![] allows dynamic slice coercion at runtime and avoids this limitation.

## Environment

	•	Rust version: 1.81.0
	•	Crate: log (latest release at time of writing)
	•	Feature: kv enabled
	•	Context: using within #[async_trait] functions

KodrAus:
[...]

PR #707, opened 2025-10-19
Fix up some possible lifetime issues in newer editions

KodrAus:
For #706

### DIFF
--- a/src/__private_api.rs
+++ b/src/__private_api.rs
@@ -18,7 +18,7 @@ mod sealed {
 
 // Types for the `kv` argument.
 
-impl<'a> KVs<'a> for &'a [(&'a str, Value<'a>)] {
+impl<'a, const N: usize> KVs<'a> for &'a [(&'a str, Value<'a>); N] {
     #[inline]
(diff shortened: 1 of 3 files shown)


src/macros.rs
```diff
@@ -431,9 +431,7 @@ macro_rules! __log_logger {
         $crate::__private_api::GlobalLogger
     }};
 
-    ($logger:expr) => {{
-        &($logger)
-    }};
+    ($logger:expr) => {&($logger)};
```
Thomasdezeeuw: The code block shouldn't matter right? (fine with the code change though)

src/__private_api.rs
```diff
@@ -18,7 +18,7 @@ mod sealed {
 
 // Types for the `kv` argument.
 
-impl<'a> KVs<'a> for &'a [(&'a str, Value<'a>)] {
+impl<'a, const N: usize> KVs<'a> for &'a [(&'a str, Value<'a>); N] {
```
Thomasdezeeuw: I'm fine with the change. I do wonder if this changes the binary size as now create code per `N`/size, but I don't know if this will impact anything.

PR #713, opened 2025-11-28
Fix clippy lints

Thomasdezeeuw:


Thomasdezeeuw:
> This new lint has been the utter bane of my existence 🙃

I think that's true for all maintainers of older projects 😅

PR #721, opened 2025-12-18
feat: impl ToValue for core::net types (Fixes #711)

### LINKED ISSUE #711: Implement `ToValue` for `IpAddr`, and `SocketAddr`
The types `IpAddr` and `SocketAddr` and the like are all defined in `core::net`, so these can be implemented even without the standard library, as far as I know.

JonathanWilbur:
Fixes #711

### DIFF
--- a/src/kv/value.rs
+++ b/src/kv/value.rs
@@ -4,6 +4,7 @@
 //! capturing and serializing them.
 
 use std::fmt;
+use core::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr, SocketAddrV4, SocketAddrV6};
 
 pub use crate::kv::Error;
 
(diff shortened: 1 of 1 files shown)


JonathanWilbur:
This currently isn't working. It seems that I have to make the `ValueBag` type support these new variants as well.

Example:

```
error[E0277]: the trait bound `ValueBag<'_>: From<std::net::SocketAddrV6>` is not satisfied
   --> src/kv/value.rs:388:39
    |
388 |                       Value::from_inner(value)
    |                       ----------------- ^^^^^ the trait `From<std::net::SocketAddrV6>` is not implemented for `ValueBag<'_>`
    |                       |
    |                       required by a bound introduced by this call
...
```

@KodrAus since you are the maintainer of the `value-bag` crate, would you take a PR to add variants for the `core::net` types, or is there a better way?

Thomasdezeeuw (commented):
I'm not sure about this. Where should we draw the line on what to support "natively" vs. using the `fmt::Display` implementation?

src/kv/value.rs
```diff
@@ -519,6 +548,16 @@ pub trait VisitValue<'v> {
         self.visit_str(&*value.encode_utf8(&mut b))
     }
 
+    /// Visit an IP Address
+    fn visit_ip_addr(&mut self, value: IpAddr) -> Result<(), Error> {
```
KodrAus, line 552: I don't think we should expose visitors for these types, and instead let them fall through to `visit_any`.

KodrAus:
Since we capture types using `ToValue` by default, adding `ToValue` impls to types from the standard library seems ok to me. We can draw the line at anything that's in the standard library and has a well understood text-based format.

@JonathanWilbur for those missing `From` impls, I think we can just not add a by-value conversion from `IpAddr` to `Value`, and instead rely on the by-ref `ToValue` and `From<&'_ T>` conversions. Only primitive copy types like integers and booleans have by-value conversions into `Value`.

Thomasdezeeuw:
> Since we capture types using `ToValue` by default, adding `ToValue` impls to types from the standard library seems ok to me. We can draw the line at anything that's in the standard library and has a well understood text-based format.

Makes sense to me 👍

KodrAus:
This was rebased in #721

PR #723, opened 2026-04-11
Hash-pin all action references, address other zizmor findings

woodruffw:
Hello! Apologies for the cold PR.

I'm opening this in my capacity as one of [uv](https://github.com/astral-sh/uv)'s maintainers; we have a set of downstreams (including `log`!) that we depend on, and we'd like to ensure their CI/CD processes are as hermetic and secure as possible (within the limits of GitHub's platform).

To that effect, this PR contains a few different commits that aim to make `log`'s CI more secure. None of these changes fix _vulnerabilities_; they're purely defense-in-depth changes that will make a future [Trivy-style compromise](https://www.wiz.io/blog/trivy-compromised-teampcp-supply-chain-attack) less fruitful for an attacker. 

To summarize:

- I've hash-pinned all of your action dependencies with `pinact run -v`. You can use Dependabot to keep these up to date with minimal maintenance burden, although it isn't enabled at the moment. I'm happy to also send a follow-up PR enabling Dependabot.
- I've disabled `actions/checkout`'s default credential-persistence behavior with `persist-credentials: false`.

_Most_ of the above was detected automatically with [zizmor](https://docs.zizmor.sh), which you can [integrate](https://docs.zizmor.sh/integrations/#github-actions) into GitHub Actions if you'd like. I've left that out of this PR however, since not every project wants another thing running in CI. But let me know if you'd like it and I'd be happy to send a follow-up PR!

Last but not least, please let me know if there's any other information I can provide. All of the above was 100% human written and reviewed 🙂

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -34,7 +34,7 @@ jobs:
             os: windows-latest
             rust: stable-x86_64-pc-windows-gnu
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
     - uses: dtolnay/rust-toolchain@master
       with:
         toolchain: ${{ matrix.rust }}
@@ -47,7 +47,7 @@ jobs:
     name: Check Format and Clippy
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
     - uses: dtolnay/rust-toolchain@stable
       with:
         components: clippy
@@ -60,7 +60,7 @@ jobs:
     name: Check Documentation
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
     - uses: dtolnay/rust-toolchain@stable
     - name: Run rustdoc
       env:
@@ -71,7 +71,7 @@ jobs:
     name: Feature check
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
       - uses: dtolnay/rust-toolchain@nightly
       - run: cargo build --verbose -Z avoid-dev-deps --features kv
       - run: cargo build --verbose -Z avoid-dev-deps --features "kv std"
@@ -84,7 +84,7 @@ jobs:
     name: Minimal versions
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
       - uses: dtolnay/rust-toolchain@nightly
       - run: cargo build --verbose -Z minimal-versions --features kv
       - run: cargo build --verbose -Z minimal-versions --features "kv std"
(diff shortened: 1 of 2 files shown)


KodrAus:
Thanks for the PR @woodruffw! Pinning our dependencies is a good idea, and should be enough of an improvement here without introducing any new tooling :+1:

The build failure is unrelated. It looks like we need to bump our MSRV.

.github/workflows/main.yml
```diff
@@ -34,11 +34,15 @@ jobs:
             os: windows-latest
             rust: stable-x86_64-pc-windows-gnu
     steps:
-    - uses: actions/checkout@v4
-    - uses: dtolnay/rust-toolchain@master
+    - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
```
Thomasdezeeuw: We should probably update our version while we're at it.
```suggestion
    - uses: actions/checkout@0c366fd6a839edf440554fa01a7085ccba70ac98 # v6.0.2
```

.github/workflows/main.yml
```diff
@@ -97,11 +116,15 @@ jobs:
     name: MSRV
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@1.68.0
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+        with:
+          persist-credentials: false
+      - uses: dtolnay/rust-toolchain@ae07f54001936ffa837af4d2b910727ec57b2084 # 1.68.0
```
Thomasdezeeuw: I think we'll want to use https://github.com/dtolnay/rust-toolchain/commit/757872fd8b3cd6b57649eec0d077552291e396b9, which is 16.8.2, which is what we really should be using.

```suggestion
      - uses: dtolnay/rust-toolchain@757872fd8b3cd6b57649eec0d077552291e396b9 # 1.68.2
```

Thomasdezeeuw (approved):
I'm 👍 on the idea. Here are the (unique) commits we're using:
* https://github.com/actions/checkout/commit/34e114876b0b11c390a56381ad16ebd13914f8d5 (v4.3.1)
* https://github.com/dtolnay/rust-toolchain/commit/ae07f54001936ffa837af4d2b910727ec57b2084 (1.68.0)
* https://github.com/dtolnay/rust-toolchain/commit/e97e2d8cc328f1b50210efc529dca0028893a2d9 (current master)
* https://github.com/taiki-e/install-action/commit/0abfcd587b70a713fdaa7fb502c885e2112acb15 (current main)

I've checked them all they seem good 👍 (GitHub actually allows references to commits in forks via the main repo path, so even using a commit hash isn't a full security solution)

I've made two suggestions to update checkout to v6 and the rust toolchain to 1.68.2:
* https://github.com/actions/checkout/commit/0c366fd6a839edf440554fa01a7085ccba70ac98 (current main)
* https://github.com/dtolnay/rust-toolchain/commit/757872fd8b3cd6b57649eec0d077552291e396b9 (1.68.2, my suggestion)

.github/workflows/main.yml
woodruffw: Oh, this was actually my mistake -- I meant to pin this at `v1` and then do `toolchain: 1.68.0`. I'll switch to that and also make it `1.68.2`!

woodruffw: (Pinning at `v1` is important because otherwise tools like Dependabot will misbehave and try to "upgrade" the action here, which would then do the wrong thing and bump your Rust toolchain version.)

KodrAus:
Perhaps while we're here we should bump that version to `1.71` so we get a green build. We've tried previously to limit our MSRV as it's pulled forwards by dependencies but it's a bit of a losing battle.

woodruffw:
Done, bumped to 1.71 🙂

.github/workflows/main.yml
```diff
@@ -97,11 +116,16 @@ jobs:
...
-      - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@1.68.0
+      - uses: actions/checkout@0c366fd6a839edf440554fa01a7085ccba70ac98 # v6.0.2
+        with:
+          persist-credentials: false
+      - uses: dtolnay/rust-toolchain@e97e2d8cc328f1b50210efc529dca0028893a2d9 # v1
         with:
           components: clippy
-      - uses: taiki-e/install-action@cargo-hack
+          toolchain: "1.71"
```
KodrAus, line 125: We'll just need to update the value in our `Cargo.toml` too to `1.71`

woodruffw: Thanks, done!

KodrAus (approved):
Thanks for working through this @woodruffw!

PR #724, opened 2026-05-17
Support capturing of `std::net` types

### LINKED ISSUE #711: Implement `ToValue` for `IpAddr`, and `SocketAddr`

KodrAus:
Rebase of #721 

Closes #711 

I've taken the original implementation from #721 and tweaked it to maintain consistent behaviour between our internal capturing, and `value-bag`'s. Thanks for originally working on this @JonathanWilbur!

### DIFF
--- a/src/kv/value.rs
+++ b/src/kv/value.rs
@@ -4,6 +4,7 @@
 //! capturing and serializing them.
 
 use std::fmt;
+use core::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr, SocketAddrV4, SocketAddrV6};
 
 pub use crate::kv::Error;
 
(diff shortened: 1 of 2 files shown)


KodrAus:
I'll need to feature gate these on `std`, since they're not available in `core` until `1.77`.

Thomasdezeeuw (approved):
Code LGTM.

A lot of this types have been stabilised as `core::net::` as well in 1.77 (released 21 March, 2024). Should we start keeping a lit of these kind of improvements we can do once we bump the MSRV?

KodrAus:
That might be a good idea :+1: The other big one is `Error`, which is available from `1.81.0`

PR #727, opened 2026-06-01
Leverage static str key when possible

tisonkun:
Replace `&'k str` in `Key<'k>` with `MaybeStaticStr<'k>` to leverage `&'static str` when in most cases it is.

Open for review, especially on:

1. Whether it is backward compatible.
2. Whether `MaybeStaticStr` causes significant indirection.

Otherwise, for downstream append, even if the key is a static str, one may have to `key.to_string` to resolve some lifetime issue. See:

* https://github.com/open-telemetry/opentelemetry-rust/blob/f7b0dd999d3d6434a6e52291efccc98e9dfcea81/opentelemetry-appender-log/src/lib.rs#L216 (they have `Key::from_static_str`)
* https://github.com/slog-rs/stdlog/blob/252b6bb6435a8a3ba44748b2e1626aa676508097/kv.rs#L46 (they have `slog::Key::from(static_str)`)

### DIFF
--- a/src/kv/key.rs
+++ b/src/kv/key.rs
@@ -1,5 +1,6 @@
 //! Structured keys.
 
+use crate::MaybeStaticStr;
 use std::borrow::Borrow;
 use std::fmt;
 
@@ -35,15 +36,23 @@ impl ToKey for str {
 // If a new field (such as an optional index) is added to the key they must not affect comparison
 #[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct Key<'k> {
-    // NOTE: This may become `Cow<'k, str>`
-    key: &'k str,
+    key: MaybeStaticStr<'k>,
 }
 
 impl<'k> Key<'k> {
     /// Get a key from a borrowed string.
     #[allow(clippy::should_implement_trait)] // Part of the public API now.
     pub fn from_str(key: &'k str) -> Self {
-        Key { key }
+        Key {
+            key: MaybeStaticStr::Borrowed(key),
+        }
+    }
+
+    /// Get a key from a static str.
+    pub fn from_static_str(key: &'static str) -> Self {
+        Key {
+            key: MaybeStaticStr::Static(key),
+        }
     }
 
     /// Get a borrowed string from this key.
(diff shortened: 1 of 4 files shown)


KodrAus (approved):
This looks like a good change to me, thanks @tisonkun!

src/kv/key.rs
```diff
@@ -35,40 +36,56 @@ impl ToKey for str {
...
     #[allow(clippy::should_implement_trait)] // Part of the public API now.
     pub fn from_str(key: &'k str) -> Self {
-        Key { key }
+        Key {
+            key: MaybeStaticStr::Borrowed(key),
+        }
+    }
+
+    /// Get a key from a static str.
+    pub fn from_static_str(key: &'static str) -> Self {
```
tisonkun: Maybe `from_static`. Not sure what a good name is, but both seem to be intuitive :P

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/log/727/761461a5d0c8ea3d483d79b1de0205c2897318d2..db145979e229549215300f2696fa89b215cb1cab)*

KodrAus: I think `from_static_str` is a good name, because it adds to the existing `from_str` name :+1:

KodrAus: Actually, looking at our other methods that use this type, they all carry a `_static` suffix, like `module_path_static`. So I think we should call this `from_str_static`.

KodrAus (changes requested):
As noted, we should call the method `from_str_static`, to match the convention of other methods in `log`

src/macros.rs
```diff
@@ -445,11 +445,11 @@ macro_rules! __log_logger {
 macro_rules! __log_key {
     // key1 = 42
     ($($args:ident)*) => {
-        $crate::__private_api::stringify!($($args)*)
+        $crate::__private_api::Key::from_static_str($crate::__private_api::stringify!($($args)*))
     };
     // "key1" = 42
     ($($args:expr)*) => {
-        $($args)*
+        $crate::__private_api::Key::from_str($($args)*)
```
Thomasdezeeuw: Would it be possible to use the static version here as well? I think it would be confusing that `key = 42` is static and `"key" = 42` is not.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/log/727/761461a5d0c8ea3d483d79b1de0205c2897318d2..db145979e229549215300f2696fa89b215cb1cab)*

tisonkun: `:expr` is not necessary to be `&'static`. I'd like to know if we can distinguish the case here as well.

I'll try to investigate it later.

src/kv/key.rs
tisonkun: This is updated.

src/macros.rs
tisonkun: 756c279649f79ce0ef8dccf952c5df4017791d1c supports this by determining `literal` as a special case. Also tests added.

Thomasdezeeuw (approved):

KodrAus (approved):
Thanks @tisonkun

Issue #661, opened 2025-01-11
log 0.4.24 with std feature always adds value-bag to lock file


zh-jq:
An empty crate with the following dependencies will always pull in a log of dependencies
```toml
[dependencies]
log = { version = "0.4", features = ["std"] }
```

Here is the output when update log from 0.4.22 to 0.4.24:
```shell
# cargo update -p log --precise 0.4.24
    Updating crates.io index
      Adding erased-serde v0.4.5
      Adding itoa v1.0.14
    Updating log v0.4.23 -> v0.4.24
      Adding proc-macro2 v1.0.93
      Adding quote v1.0.38
      Adding ryu v1.0.18
      Adding serde v1.0.217
      Adding serde_derive v1.0.217
      Adding serde_fmt v1.0.3
      Adding sval v2.13.2
      Adding sval_buffer v2.13.2
      Adding sval_dynamic v2.13.2
      Adding sval_fmt v2.13.2
      Adding sval_json v2.13.2
      Adding sval_nested v2.13.2
      Adding sval_ref v2.13.2
      Adding sval_serde v2.13.2
      Adding syn v2.0.96
      Adding typeid v1.0.2
      Adding unicode-ident v1.0.14
      Adding value-bag v1.10.0
      Adding value-bag-serde1 v1.10.0
      Adding value-bag-sval2 v1.10.0
```

hanna-kruppe:
From what I can tell, it's added to the lockfile due to a Cargo bug (https://github.com/rust-lang/cargo/issues/10801) but not actually built. Still annoying because spurious lockfile entries mislead a lots of helpful tools that rely on parsing the lockfile or the output of `cargo metadata`.

KodrAus:
That’s a pretty alarming number of dependencies we don’t actually use. The bloat may be enough to warrant trying to find a workaround here, but I suspect we may be a bit stuck. At least based on that linked issue it looks like we’ve got plenty of company.

hanna-kruppe:
What would be the downside of reverting back to "compound" features like `kv_std`? Is it just less convenient/idiomatic or is there useful functionality that can only be provided with the new style? Removing the feature implication again is technically a breaking change for anyone who started relying on it, but if it's done quickly enough (and 0.4.24 is yanked so nothing new starts depending on it) it might be fine for practical purposes.

KodrAus:
Until we figure this out I've yanked `0.4.24`.

[...]

PR #729, opened 2026-06-03
kv::std_support may not need value-bag

tisonkun:
Exclude `value-bag` from the dependency tree when we don't need it.

The code would still leverage `value-bag` if it is enabled, thanks to the Value inner design.

cc @KodrAus

### DIFF
--- a/src/kv/value.rs
+++ b/src/kv/value.rs
@@ -415,7 +415,7 @@ impl<'v> Value<'v> {
     }
 }
 
-#[cfg(feature = "kv_std")]
+#[cfg(all(feature = "kv", feature = "std"))]
 mod std_support {
     use std::borrow::Cow;
     use std::rc::Rc;
@@ -462,6 +462,7 @@ mod std_support {
         }
     }
 
+    #[cfg(feature = "kv_std")]
     impl<'v> Value<'v> {
         /// Try convert this value into a string.
         pub fn to_cow_str(&self) -> Option<Cow<'v, str>> {
--- a/src/kv/value.rs
+++ b/src/kv/value.rs
@@ -415,7 +415,7 @@ impl<'v> Value<'v> {
     }
 }
 
-#[cfg(all(feature = "kv", feature = "std"))]
+#[cfg(feature = "std")]
 mod std_support {
     use std::borrow::Cow;
     use std::rc::Rc;


src/kv/value.rs
```diff
@@ -462,6 +462,7 @@ mod std_support {
         }
     }
 
+    #[cfg(feature = "kv_std")]
```
tisonkun, line 465: We may implement this for the non-value-bag Value. But that can be a follow-up.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/log/729/580839288e5f2babc17e6c36f7d56e60082a47ef..7bc120062895aadd440ab015e62275841465a1a6)*

KodrAus:
We basically shouldn’t depend on `kv_std` anywhere except where it’s actually needed. It’s just an unfortunate workaround for not being able to activate Cargo features in a dependency based on the presence of multiple other features.

tisonkun:
@KodrAus I remembered https://github.com/rust-lang/log/pull/657 .. Let me see if this makes some different.

tisonkun:
@KodrAus I believe the issue is `std = ["value-bag?/std"]` but now we don't have that anymore.

tisonkun:
That is, I test this PR locally to exclude `value-bag` from my dependency tree and it works well.

KodrAus:
If I remember correctly, part of the issue too was we couldn’t use a combination of `dep:` and `?` in our MSRV. Now that it’s raised we might be able to do that now.

We can continue to follow up with those kinds of improvements though. This specific change is still a step forwards.

src/kv/value.rs
```diff
@@ -415,7 +415,7 @@ impl<'v> Value<'v> {
     }
 }
 
-#[cfg(feature = "kv_std")]
+#[cfg(all(feature = "kv", feature = "std"))]
```
KodrAus: This should just need to gate on `std`. We’re already under the `kv` feature.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/log/729/580839288e5f2babc17e6c36f7d56e60082a47ef..7bc120062895aadd440ab015e62275841465a1a6)*

tisonkun: Updated.

KodrAus (approved):
Thanks @tisonkun

tisonkun:
@KodrAus Thanks for your review and merging!

I'd appreciate it if you could test it (locally?) to prevent a regression to https://github.com/rust-lang/log/issues/661 and cut a release if it works well :D

I can move to that version downstream then. So far I'm patching in `Cargo.toml` to use a git dep.

KodrAus:
I think we're ok there because we aren't adding any dependencies based on the presence of `std` here.

PR #732, opened 2026-06-19
Fixed key comparison

matteo-zeggiotti-ok:
Key comparison was borken between 'static and generic 'a str

### DIFF
--- a/src/kv/key.rs
+++ b/src/kv/key.rs
@@ -34,7 +34,7 @@ impl ToKey for str {
 /// A key in a key-value.
 // These impls must only be based on the as_str() representation of the key
 // If a new field (such as an optional index) is added to the key they must not affect comparison
-#[derive(Clone, Debug, Eq, Ord, Hash)]
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct Key<'k> {
     key: MaybeStaticStr<'k>,
 }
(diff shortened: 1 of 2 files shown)


src/kv/key.rs
```diff
@@ -107,6 +107,18 @@ impl<'k> From<&'k str> for Key<'k> {
     }
 }
 
+impl PartialEq for Key<'_> {
```
Thomasdezeeuw: This should probably be done for `MaybeStaticStr`, so it's fixed for all usages of it

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/log/732/a5b5b2113e2767801250af184d6c3971e689ae3b..0d71d3c685f2e23b1ad209b48408efe1205b18b0)*

src/lib.rs
```diff
@@ -777,7 +777,7 @@ impl LevelFilter {
     }
 }
 
-#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+#[derive(Copy, Clone, Eq, Ord, Hash, Debug)]
```
KodrAus: We’ll need to override the other comparison traits too to be based off the underlying `&str`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/log/732/a5b5b2113e2767801250af184d6c3971e689ae3b..920e7dc2811c18a228bf78e818196de950659d85)*

src/lib.rs
```diff
@@ -777,7 +777,7 @@ impl LevelFilter {
     }
 }
 
-#[derive(Copy, Clone, Eq, Ord, Hash, Debug)]
+#[derive(Copy, Clone, Hash, Debug)]
```
KodrAus: We also need to do `Hash`. Sorry I was unclear @matteo-zeggiotti-ok 🙂 Thanks for fixing this up!

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/log/732/a5b5b2113e2767801250af184d6c3971e689ae3b..cc89cc6e41190de36892e33fff48e5f48cf57fa9)*

matteo-zeggiotti-ok: Thanks for your patience!

KodrAus (approved):
Thanks for fixing this up @matteo-zeggiotti-ok!

PR #734, opened 2026-06-20
Prepare for 0.4.33 release

KodrAus:
https://github.com/rust-lang/log/compare/0.4.32...main

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -348,7 +348,7 @@
 #![doc(
     html_logo_url = "https://prev.rust-lang.org/logos/rust-logo-128x128-blk-v2.png",
     html_favicon_url = "https://prev.rust-lang.org/favicon.ico",
-    html_root_url = "https://docs.rs/log/0.4.32"
+    html_root_url = "https://docs.rs/log/0.4.33"
 )]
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -348,7 +348,7 @@
 #![doc(
     html_logo_url = "https://prev.rust-lang.org/logos/rust-logo-128x128-blk-v2.png",
     html_favicon_url = "https://prev.rust-lang.org/favicon.ico",
-    html_root_url = "https://docs.rs/log/0.4.32"
+    html_root_url = "https://docs.rs/log/0.4.33"
```
Thomasdezeeuw, line 351: Offtopic for this pr, but do we still need this?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/log/734/87e062162e051d54bb553aacae3f0c6c4c213e59..6a24abf0835cef62e3d882287c97307dd3ecb403)*

Thomasdezeeuw (approved):

src/lib.rs
KodrAus: I don't believe so. We can probably remove it next release :+1:

PR #736, opened 2026-07-29
Derive Default for Level and LevelFilter enums

ssrlive:
When using log crate, we might sometimes define a struct with the `Default` macro,
but if `Level` or `LevelFilter` doesn't implement the `Default` trait, this struct will fail to compile.

Therefore, I created this PR. I hope you will approve it and merge it.

```rust
#[derive(Debug, Clone, Default)]
pub struct Args {
    pub item1: Option<String>,

    /// Log level (off, error, warn, info, debug, trace)
    pub log: log::LevelFilter,
}
```


Thomasdezeeuw:
I'm not sure about implementing default for the level, I imagine we might not be able to agree on what the default should be.

Btw you can also implement `Default` manually, this way you can continue with your project without having to wait on this.

KodrAus:
Thanks for the PR @ssrlive! I share the same reservations as @Thomasdezeeuw, and don’t believe we should implement `Default` for level. There are reasonable arguments for picking the lowest/highest/info and none is a universal default.

Thomasdezeeuw:
Since we got no response after raising our concern I'm going to close this.

PR #739, opened 2026-09-19
initial draft for downcastable Log

### LINKED ISSUE #666: Feature Request: downcasting `Log`, but for real
Seems #399 was already closed (4.5 years ago, for bad usecase & maybe backwards compatibility reasons?), but I have an actual use!

My logger writes a message on panic, but does so with a custom format different to anything you can write with the `Log` trait. So I need to be able to downcast the `&dyn Log` into my own logger from the panic handler. I could probably use a `OnceLock` to store if my logger's been initialized but `log` already does that with the (`set_`)`logger` methods, so it feels redundant.

For now my current way of doing this is this very cursed set of functions:
```rs
fn as_dyn_ref(logger: *const Logger) -> *const dyn Log {
  // split into one function to always attach the same metadata
  logger as *const dyn Log
}
fn upcast_log(logger: &'static Logger) -> &'static dyn Log {
  // SAFETY: as_dyn_ref returns a reference to the same object as passed in
  unsafe { &*as_dyn_ref(logger) }
}
fn downcast_log(log: &'static dyn Log) -> Option<&'static Logger> {
  // horribly cursed implementation to fetch a reference to the installed logger
  let (logger_ptr, logger_meta) = (&raw const *log).to_raw_parts();
  let (_, fake_logger_meta) = as_dyn_ref(ptr::null::<Logger>()).to_raw_parts();
  (logger_meta == fake_logger_meta).then(|| {
    // SAFETY: v-tables match so it's probably ours!
    unsafe { &*logger_ptr.cast::<Logger>() }
  })
}
```
but that's a lot of unsafe, and nightly-only.

To prevent needing to add a `+ 'static` bound on `Log` (which probably breaks compatibility somehow), yandros suggested ([in rplcs #dark-arts](https://discord.com/channels/273534239310479360/592856094527848449/1336662380997902396)) adding something like:
```rs
fn type_id(&self) -> TypeId
  where Self: 'static
{
  TypeId::of<Self>()
}
```
as an automatic implementation on the `Log` trait.

djugei:
fixes #666.
i do not expect this to be merged as is, this is more of a discussion helper. Names are up for bikeshedding.
technically this is not a breaking change i think, as anything that implements log automatically implements LogAny though i agree that it looks weird.

can't just have a method on Log sadly as that would introduce a `Sized` requirement (for some reason). Can't add the Any bound to Log itself as it would need to be 'static

i have tested this with indicatif-log-bridge and env_logger and it works


KodrAus:
Thanks for the PR @djugei! I think we can get away without the `as_any` method, since recent Rust releases support dyn trait upcasting. Maybe something like:

```rust
pub trait GlobalLog: Log + Any {}

impl<T: Log + Any> GlobalLog for T {}
```

I did look at a few alternative designs that try to fully hide the downcasting support, but they all had unsatisfying trade-offs since we accept a `Box<dyn Trait>` in our public API.

djugei:
just makes user code a bit uglier
`log::logger().as_any().downcast_ref()`
`(log::logger() as &'static dyn std::any::Any).downcast_ref()`
but yes, no additional capabilities so can be removed.

is this feature in general something the project would like to add?

KodrAus:
I'd be in favour of this approach with the `as_any` method removed. I think we should call the trait `GlobalLog`, and use it as our marker for a `Log` that can be plugged into the shared slot.

Thomasdezeeuw:
I think this a breaking change. We didn't the `std::any::Any` trait bound on these functions before, now we effectively do. So we might as well do `pub trait Log: std::any::Any { ... }`.

KodrAus:
@Thomasdezeeuw Yeh I checked this as well, and couldn’t actually trigger a compile error by flipping from `Box<dyn Log>` to `Box<dyn GlobalLog>`, I believe because of the default `'static` object bound. If there’s any chance this is a breaking change though we should avoid it.

Adding an `Any` bound to `Log` itself is almost certainly a breaking change.

Thomasdezeeuw:
Not 100% accurate, but I think https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=7e1712bc6dbdeb86780c85de4b0a7e6c shows it's a problem.

djugei:
i tried doing `Log: Any` but that lead to lifetime issues with `impl Log for &'_ T: Log` so i did a separate trait, though i guess those could be tightened to &'static and it would be fine.

i think any concrete type that implements Log will also implement Any, but if you have a function that bounds only on Log then the compiler is ignorant of that fact, so yeah i guess this is a breaking change after all, though one that can be fixed with minimal changes/no structural changes.

KodrAus:
Ah yes I see. If you explicitly already have a `Box<dyn Log>` you can’t pass it to the `set_boxed_logger` function when we introduce `GlobalLog`.

That would mean we’d have to introduce a new set of APIs for this if we wanted to do it and deprecate the old ones. I’m not sure that level of churn is worth it.

djugei:
> That would mean we’d have to introduce a new set of APIs for this if we wanted to do it and deprecate the old ones. I’m not sure that level of churn is worth it.

it is even worse than that: since we need to store the logger we have to decide on a concrete type, so those would be 2 completely separate apis, which does not make a lot of sense to me, as it would essentially be keeping two versions. in that case i would just keep one version in the library and let cargo do its job for versioning.

alternatively we can try to think about &'static Log always being Any a bit harder and maybe do an unsafe cast if we can convince ourselves that it is always safe? i sadly do not know a ton about trait objects.

then again this functionality is achievable without using Any at all (see issue) but uses unsafe to do so in the first place. i was hoping for an elegant solution with Any.

